PR #1, opened 2018-06-28
Feature/annot page new scroll

Ren22:


metaspace/webapp/src/components/ImageLoader.vue
```diff
@@ -202,6 +211,34 @@
...
+             y = (event.clientY - rect.top) / this.scaleFactor / this.zoom,
+             xOffset = -(this.zoom / newZoom - 1) * x + this.zoom / newZoom * this.xOffset,
+             yOffset = -(this.zoom / newZoom - 1) * y + this.zoom / newZoom * this.yOffset;
+
+         this.$emit('zoom', {zoom: newZoom});
+         this.$emit('move', {xOffset, yOffset});
+       }
+       else if (event.deltaY) {
+         this.overlayFadingIn = true;
+         clearTimeout(tmId);
```
LachlanStuart: `tmId` needs to be an instance property or a Vue `data`, not a plain variable. Currently `clearTimeout` is always called with `0` because it's a new variable every time this function is called.

metaspace/webapp/src/components/ImageLoader.vue
```diff
@@ -411,9 +449,37 @@
...
    cursor: grab;
    width: 100%;
+   height: 100%;
    line-height: 0px;
    display: flex;
    flex-direction: column;
    align-self: center;
  }
+
+ .image-loader__overlay--text {
```
LachlanStuart: Per BEM, there should only be a single `-` between `overlay` and `text`. `image-loader__overlay--text` means "In the `image-loader` Block this is an Element called `overlay` that has the `text` Modifier", whereas `image-loader__overlay-text` means "In the `image-loader` Block this is an Element called `overlay-text`."

metaspace/webapp/src/components/annotation-widgets/default/MainImage.vue
```diff
@@ -41,19 +43,6 @@
             </div>
         </div>
     </div>
-    <div class="main-ion-image-container-row2">
```
LachlanStuart: There is now an unused `.main-ion-image-container-row2` in the `<style>` section that should be cleaned up.

PR #2, opened 2018-07-02
graphql mutations refactoring

intsco:
- Adapted the graphql and web app code to the new mutations schema
- Bug fix in sm/engine/dataset_manager.py


intsco:
@LachlanStuart please check my changes in the webapp code.
@Ren22 would it be much work for you to update 'python-client' to work with the new mutations?

Ren22:
@intsco I think it would be possible for me to update python-client in the next couple of days

PR #3, opened 2018-07-03
Python client/renamed package

Ren22:


Ren22:
@intsco @LachlanStuart if the name of package is fine, I can merge these changed to the master

PR #6, opened 2018-07-04
Add padlock icon to annotation view to indicate private datasets

LachlanStuart:
I ran out of inspiration on this one. I'm not completely happy with it, but I'm not sure what else to do. Comments & suggestions welcome.

Normal:
image

Private:
image

Hover text:
image

Share link hover text (this was changed from a native browser tooltip to an Element UI popover to be consistent with the padlock):
image

### DIFF
--- a/metaspace/webapp/src/api/annotation.ts
+++ b/metaspace/webapp/src/api/annotation.ts
@@ -22,6 +22,7 @@ export const annotationListQuery =
           name
           polarity
           metadataJson
+          isPublic
         }
         isotopeImages {
           mz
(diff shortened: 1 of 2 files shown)


theodev:
Looks good to me!

I can imaging it might be changing the width of the horizontal space where
the mol formula and the link are in, so when browsing, they will be
"jumping" right-left, but this is a minor issue and should not be too
annoying

Ren22:
from UX perspective I think it's good and it should be clear why there's a padlock next to the annotation name, the only thing I'd avoid is a small pixel like detail on padlock(that is, a hole) that might be invisible depending on the screen size and don't bring any functional connotation.

LachlanStuart:
Good point about the small detail @Ren22 although I don't think the keyhole is too much of an issue - [most "locked" icons have keyholes](https://www.google.com/search?q=locked+icon&tbm=isch), so I think people will be used to it. Keeping the keyhole means that we can use a larger version of the same icon on the Datasets page, which [looks a bit like a handbag](http://iconshow.me/locked-icon) if it doesn't have a keyhole at the 48x48 size.

PR #8, opened 2018-07-13
Proposed changes to graphql for Data Management

LachlanStuart:
This PR is not for merging - GitHub just seems like a better place for sharing this than Google Docs. Feel free to comment, add commits, or even directly edit the files through GitHub.

These are the API changes I think are necessary for the first stage of the Data Management project. Please review them for sanity, check that they meet the requirements and your expectations of the data model, etc.

This is split into multiple files because Apollo has a feature to [stitch schemas together into one API](https://www.apollographql.com/docs/graphql-tools/schema-stitching.html), which we should use.

Mockups: https://app.moqups.com/LachlanStuart/liDSQ2TVih/view
Requirements: https://docs.google.com/document/d/1L5f4Oml2d3YgBnWG8hHyEuzh2XAjcTx2L4dnKKElsPg/edit
Old user stories: https://docs.google.com/presentation/d/1DQSDHpBOkRJMqwzNIbcmAVXpHlIEHUfIC9U57HxPnKc/edit#slide=id.g2717c0572c_0_15

This intentionally does not include authentication (create account, login, logout, reset password, etc.), as I'm planning to implement it outside of GraphQL. However after the user is logged in, it should still be possible to query everything about the currently logged-in user through GraphQL.


metaspace/graphql/schema.graphql
```diff
@@ -310,6 +318,8 @@ type Query {
   thumbnailImage(datasetId: String!): String,
 
   reprocessingNeeded(input: DatasetUpdateInput!): Boolean!
+
+  currentUserLastSubmittedDataset(): Dataset
```
LachlanStuart: I've added this because I'd like to stop using browser storage to keep a copy of the user's last submission. It adds the potential for a lot of bugs as we can't reliably test migrations for the data stored in peoples' browsers.

metaspace/graphql/schemas/group.graphql
```diff
@@ -0,0 +1,65 @@
...
+  currentUserRoleInGroup(groupId: ID!): String
+}
+
+type Mutation {
+  ## Managing groups
+  createGroup(groupDetails: CreateGroupInput!): Group!
+  updateGroup(groupDetails: UpdateGroupInput!): Group!
+  deleteGroup(groupId: ID!): Boolean!
+
+  ## Managing group users
```
LachlanStuart: There are a lot of ways that these functions could be designed. I've just gone with what seemed most descriptive through the API, but other ideas are welcome. e.g. we could have CRUD functions on `UserGroup`s with lots of security checks - to request access a user would `createUserGroup(groupId: groupId, userId: currentUserId, role: 'PENDING')`, and to accept them the group admin would `updateUserGroup(groupId: groupId, userId: currentUserId, role: 'MEMBER')`

metaspace/graphql/schemas/group.graphql
```diff
@@ -0,0 +1,65 @@
...
+}
+
+type CreateGroupInput {
+  name: String!
+  shortName: String!
+  principalInvestigatorEmail: String!
+}
+
+type UpdateGroupInput {
+  id: ID!
```
intsco: @LachlanStuart why don't make **id** a separate argument of the mutation? This way we can separate what is being updated from what is not.

LachlanStuart: That makes sense. I've changed it.

Ren22:
@LachlanStuart I'd like to introduce several functions to the schema for the user profile page. Can we discuss it first or should I create a commit already for the feedback?

LachlanStuart:
@intsco @Ren22 I've pushed a commit that stitches the schemas together and adds mocking for the resolvers. GraphQL should now return random data for all the unimplemented resolvers, e.g.

image

I had to use a separate package `merge-graphql-schemas` instead of Apollo's schema stitching, because the version of Apollo that we're on doesn't support it. I tried doing a partial upgrade of `graphql-tools` to get a version that supported schema stitching, but the newest version that I could upgrade to was buggy.

metaspace/graphql/schemas/group.graphql
```diff
@@ -0,0 +1,66 @@
...
+  updateGroup(id: ID!, groupDetails: UpdateGroupInput!): Group!
+  deleteGroup(groupId: ID!): Boolean!
+
+  ## Managing group users
+  leaveGroup(groupId: ID!): Boolean!
+  removeUserFromGroup(groupId: ID!, userId: ID!): Boolean!
+
+  ## User requests access
+  # `bringDatasets` contains a list of IDs of datasets that the user wants to bring into the group with them
+  requestAccessToGroup(groupId: ID!, bringDatasets: [ID!]!): UserGroup!
```
Ren22: I'm not very sure if whether `bringDatasets: [ID!]!` should be `[ID!]!`

intsco: I'd say [ID!] is enough

LachlanStuart: My thinking was just that it's not good having unnecessary ambiguity. `null` and `[]` would be exactly the same value. Allowing both values would mean that the server code would have to have code for handling both, even though they both do the same thing.

Normally in this case it's better to use default values rather than to introduce multiple potential values, i.e. `bringDatasets: [ID!]! = []`. Unfortunately our version of GraphQL has a bug and requires that non-nullable parameters always be provided, even if they have a default, so `bringDatasets: [ID!] = []` is the best we can do here until we upgrade.

metaspace/graphql/schemas/group.graphql
```diff
@@ -0,0 +1,66 @@
...
+
+type Mutation {
+  ## Managing groups
+  createGroup(groupDetails: CreateGroupInput!): Group!
+  updateGroup(id: ID!, groupDetails: UpdateGroupInput!): Group!
+  deleteGroup(groupId: ID!): Boolean!
+
+  ## Managing group users
+  leaveGroup(groupId: ID!): Boolean!
+  removeUserFromGroup(groupId: ID!, userId: ID!): Boolean!
```
LachlanStuart: @Ren22 @intsco I've just noticed how inconsistent I've been here with the argument names (`groupId` in some places, `id` in others). Would you prefer `groupId`/`userId` to be used everywhere, or only when necessary?

e.g. 
```graphql
  updateGroup(groupId: ID!, groupDetails: UpdateGroupInput!): Group!
  leaveGroup(groupId: ID!): Boolean!
  removeUserFromGroup(groupId: ID!, userId: ID!): Boolean!
```

or

```graphql
  updateGroup(id: ID!, details: UpdateGroupInput!): Group!
  leaveGroup(id: ID!): Boolean!
  removeUserFromGroup(id: ID!, userId: ID!): Boolean!
```

or some other rule?

Ren22: @LachlanStuart  I'd say that 1st option makes it more understandable and distinct, even though it's more wordy

PR #10, opened 2018-07-16
Fixes to scrolling on annotations page

Ren22:
- fixed text alignment problem (Use ctrl to zoom in..)
- filling of the empty space on Y axis (before only free space on X axis was used when zooming)
- domtoimage browser dependency warning(works on Firefox/Chrome, warns on other browsers)

### DIFF
--- a/metaspace/webapp/src/components/annotation-widgets/default/MainImage.vue
+++ b/metaspace/webapp/src/components/annotation-widgets/default/MainImage.vue
@@ -105,7 +105,7 @@ export default class MainImage extends Vue {
         })
     }
 
-    showBrowserWarning(): any {
+    showBrowserWarning() {
       this.$alert('Due to technical limitations we are only able to support downloading layered and/or zoomed images' +
       ' on Chrome and Firefox. As a workaround, it is possible to get a copy of the raw ion image by right-clicking ' +
       'it and clicking "Save picture as", however this will not take into account your current zoom ' +
--- a/metaspace/webapp/src/components/annotation-widgets/default/MainImage.vue
+++ b/metaspace/webapp/src/components/annotation-widgets/default/MainImage.vue
@@ -44,7 +44,7 @@
                      width="32px"
                      style="opacity: 0.3"
                      title="Your browser is not supported"
-                     @click="browserWarning"
+                     @click="showBrowserWarning"
                      v-else/>
             </div>
         </div>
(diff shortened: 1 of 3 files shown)


metaspace/webapp/src/components/annotation-widgets/default/MainImage.vue
```diff
@@ -99,8 +105,13 @@ export default class MainImage extends Vue {
         })
     }
 
+    browserWarning(): string {
+      return alert('Your browser is not supported, please change to Chrome/Firefox instead')
```
LachlanStuart: * Please change this to "Due to technical limitations we are only able to support downloading layered and/or zoomed images on Chrome and Firefox. As a workaround, it is possible to get a copy of the raw ion image by right-clicking it and clicking "Save image...", however this will not take into account your current zoom settings or show the optical image."
* This function's return value isn't used and probably won't ever be used. It shouldn't return anything, and it should be named with an imperative phrase, e.g. `showBrowserWarning`
* Please use an ElementUI alert, because browser alerts aren't very nice.

e.g.
```
showBrowserWarning(): {
    this.$alert('Due to technical limitations we are only able to support downloading layered and/or zoomed images on Chrome and Firefox. As a workaround, it is possible to get a copy of the raw ion image by right-clicking it and clicking "Save picture as", however this will not take into account your current zoom settings or show the optical image.');
}
```

Ren22:
@LachlanStuart thank you for the review, I actually just saw that you left it, as there was surprisingly no message to my email. Will change the alert message to native element-ui style.

metaspace/webapp/src/components/annotation-widgets/default/MainImage.vue
```diff
@@ -105,8 +105,11 @@ export default class MainImage extends Vue {
         })
     }
 
-    browserWarning(): string {
-      return alert('Your browser is not supported, please change to Chrome/Firefox instead')
+    showBrowserWarning(): any {
```
Ren22: should it be changed to void instead?

intsco:
@Ren22 We are introducing a new team policy. The creator of a PR is responsible for merging it. If you need a confirmation or help, you ask for a review.

metaspace/webapp/src/components/annotation-widgets/default/MainImage.vue
intsco: If it doesn't return anything, I would say it should have 'void' return type

LachlanStuart: I forgot to delete the `:` in my suggestion. `void` is definitely better than `any` here, but the best option is to not specify a type for the return value, and not use the `return` keyword anywhere in the function, so that TypeScript infers the type as `void`.

Ren22:
@intsco got it, thanks for letting me know!

intsco:
@Ren22 Oh, and we also agreed to do "Squash and Merge" instead of just "Merge"

intsco:
@Ren22 this too

PR #11, opened 2018-07-18
User Authentication (UI & API shims)

LachlanStuart:
Major changes:
* Adds UI for account creation, sign in and password reset to `webapp` (Hidden behind a feature flag - set `config.features.newAuth` to `false` to keep the old UI/backend or `true` to use the new code)
* Adds Passport to `graphql` to handle username/password and Google logins, including Redis sessions and JWT generation
* Adds TypeScript support to `graphql`
* Adds Jest tests to `webapp` (run on every commit by CircleCI)
* **Lots of new config settings** - make sure to update your local config files
* Starts using a new module-centric directory structure in both `webapp` and `graphql`, which I'd like to transition to. 
    * The goal is to keep related code together as much as possible so that if we have lots of small components, utility functions that are only used in a few places, tests, etc. it doesn't become overwhelming. What constitutes "related code" is completely subjective, but note that the goal is keeping code easy to think about, not to make an accurate taxonomic hierarchy of our code.
    * I've added `index.ts` files as facades to the module, which basically list the "public" parts of the modules which other parts of the application may use. I'm not 100% sure we'll keep doing this - I've had trouble with WebStorm in the past when using facades, but hopefully they've fixed those issues by now.
    * In `webapp` I've intentionally left `api` separate from the modules - the files in this directory should each represent modules in `graphql`, not in `webapp`. There usually won't be a 1:1 correlation.
    
Other changes:
* Splits `webapp`'s `tsconfig.json` into `tsconfig.json` (for tests) and `tsconfig.client.json` (for browser). I was tempted to also add a configuration for the server, but when we remove the authentication from `webapp` there won't be enough server-side code to justify using TypeScript.
* Changed `yarn test` in `webapp` to run the Jest unit tests instead of Testcafe e2e tests. 
* Removed `bluebird`, `babel`, `pg-promise`, `promise` dependencies, and `webpack`-related `package.json` scripts from `graphql` as they seemed to be unnecessary.


metaspace/webapp/package.json
```diff
@@ -8,11 +8,13 @@
     "dev": "cross-env NODE_ENV=development node --max_old_space_size=4096 server.js",
     "build": "cross-env NODE_ENV=production webpack --progress --hide-modules",
     "build-ci": "cross-env NODE_ENV=production webpack --hide-modules",
-    "test": "testcafe firefox:headless tests/e2e/annotations.js --app 'yarn run dev' --no-color --app-init-delay=20000 --quarantine-mode --assertion-timeout=20000",
+    "test": "jest",
```
LachlanStuart: BTW, it's also possible to have this running in "watch mode" with `yarn run test --watch`. Yarn does an excellent job of only re-running tests that cover the places where you've made code changes.

metaspace/graphql/dsMutation.js
```diff
@@ -9,10 +9,8 @@ const {db, logger, fetchDS, assertUserCanEditDataset,
     addProcessingConfig, fetchMolecularDatabases} = require('./utils.js'),
   metadataSchema = require('./metadata_schema.json');
 
-let {molecularDatabases} = 1;
-
-ajv = new Ajv({allErrors: true});
-validator = ajv.compile(metadataSchema);
```
LachlanStuart: No functionality change here - these variables just needed to be declared properly because now TypeScript is parsing this file, and even when it's running in JS mode it has slightly stricter syntax requirements.

metaspace/webapp/src/modules/Account/components/CreateAccountDialog.vue
```diff
@@ -0,0 +1,215 @@
...
+      }
+    }
+
+    onClose() {
+      this.$store.commit('account/hideDialog', 'createAccount');
+    }
+  }
+</script>
+
+<style scoped lang="scss">
```
LachlanStuart: @Ren22 Note the `scoped` attribute on this `<style>` block. It's the Scoped CSS feature that `vue-loader` has that I mentioned when we were discussing BEM. If you look at the generated HTML, Vue adds an extra attribute to every element within this component and to all of the CSS selectors here. This means that rules defined here only affect elements in this component (unless `/deep/` is used to indicate that parts of a selector should be un-scoped).

The output looks like this:
HTML:
```html
<div data-v-17e4e483 class="container">
  <div data-v-17e4e483 class="left-side">
    <form data-v-17e4e483 class="el-form">
```
CSS:
```css
.container[data-v-17e4e483] {
    display: flex;
    ...
}
.left-side[data-v-17e4e483] {
    ...
}
/*  Note that `.el-dialog` isn't filtered by the `[data-v-17e4e483]` attribute - that's because 
    of the `/deep/` operator. It's needed because the `.el-dialog` element isn't rendered by 
    this component - it's rendered in ElementUI's Dialog component instead */
.el-dialog__wrapper[data-v-17e4e483] .el-dialog {
    width: 400px;
}
```

PR #12, opened 2018-07-21
User profile page

Ren22:
Added UI component for a user profile page. Currently the implementation part is just a mock-up, I will populate it with code as it will become clear with a scheme on graphql for user information.


metaspace/webapp/src/components/UserEditPage.vue
```diff
@@ -0,0 +1,408 @@
...
+          }
+        });
+        this.$message({
+          type: "success",
+          message: "You have successfully deleted your account!"
+        })
+      } catch (err) {
+        reportError(err, 'There was a problem with deleting the user account.');
+      } finally {
+        this.cancelDeleteAccount();
```
LachlanStuart: If this function is going to be used in a place where account deletion isn't being cancelled, it needs to be named appropriately. e.g. `closeDeleteAccountDialog`.

metaspace/webapp/src/components/UserEditPage.vue
```diff
@@ -0,0 +1,408 @@
...
+            deleteDatasets: this.delDatasets
+          }
+        });
+        this.$message({
+          type: "success",
+          message: "You have successfully deleted your account!"
+        })
+      } catch (err) {
+        reportError(err, 'There was a problem with deleting the user account.');
+      } finally {
```
LachlanStuart: After deleting the user account (preferably in the `finally` method, so that it gets run even if there's a server error), please refresh the user's authentication status. 

If this PR gets merged before #19, then use this code:
```
import tokenAutorefresh from '../tokenAutorefresh';
await tokenAutorefresh.refreshJwt();
```

If it's merged after, use this:
```
import {refreshLoginStatus} from '../graphqlClient';
refreshLoginStatus();
```

metaspace/webapp/src/components/UserEditPage.vue
```diff
@@ -0,0 +1,408 @@
...
+          message: "New message to verify your account was sent to your account"
+        });
+      } catch(err) {
+        reportError(err, 'There was a problem with updating email.');
+      } finally {
+        this.isUserDetailsLoading = false;
+      }
+    }
+
+    async leaveGroup(ind: number, rows: GroupsData[]) {
```
LachlanStuart: 1. It would be better to pass the group or group ID instead of `ind` here. `rows` could potentially change during some of the async operations, which would make `rows[ind].id` point to a different group than the user clicked on.
2. This function shouldn't need the `rows` argument - it can access `this.groupsData` directly.

metaspace/webapp/src/components/UserEditPage.vue
```diff
@@ -0,0 +1,408 @@
...
+      try {
+        try {
+          await this.$msgbox({
+            message: 'Are you sure you want to leave this group?',
```
LachlanStuart: This message should include the group name, so that there's no uncertainty about the user clicked.

metaspace/webapp/src/components/UserEditPage.vue
```diff
@@ -0,0 +1,408 @@
...
+    <div class="user-edit-page">
+      <el-dialog
+        title="Delete account"
+        :visible.sync="showDeleteAccountDialog"
+        width="30%"
+        :lock-scroll="false">
+        <p>If you delete your account, you will lose access to any datasets, groups and
+        projects that have been explicitly shared with you</p>
+        <el-checkbox v-model="delDatasets" style="margin-left: 20px">
+          Delete datasets the that I have submitted
```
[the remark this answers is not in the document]
Ren22: Sorry, that was a remnant of the previous sentence here

metaspace/webapp/src/components/UserEditPage.vue
```diff
@@ -0,0 +1,408 @@
...
+              </el-col>
+            </div>
+          </el-form>
+        </el-row>
+
+        <div class="groups">
+          <h2>Groups</h2>
+          <el-table
+            :data="groupsData"
+            border
```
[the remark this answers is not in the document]
Ren22: yep , indeed looks cleaner

Ren22:
@LachlanStuart  this PR was updated after the great code review you left earlier. 
The test failed as there were troubles we discovered earlier (1 snapshot failed, please see CI log). 
Tested it manually so far.

metaspace/webapp/src/components/UserEditPage.vue
```diff
@@ -0,0 +1,470 @@
...
+        this.isUserDetailsLoading = false;
+      }
+    }
+
+    async leaveGroup(action: string, groupRow: GroupsData) {
+      try {
+        try {
+          await this.$msgbox({
+            message: (action === 'leave') ? `Are you sure you want to leave ${groupRow.name} group?` :
+              `Are you sure you want to decline the invitation to ${groupRow.name} group?`,
```
LachlanStuart: Please remove "group" on the above two lines as it doesn't make sense when a group name is inserted. i.e. 
```
`Are you sure you want to leave ${groupRow.name}?` :
    `Are you sure you want to decline the invitation to ${groupRow.name}?`,
```

metaspace/webapp/src/components/__snapshots__/UserEditPage.spec.ts.snap
```diff
@@ -0,0 +1,168 @@
...
+              </div>
+              <div class="el-scrollbar__bar is-horizontal">
+                <div class="el-scrollbar__thumb" style="transform: translateX(0%); -webkit-transform: translateX(0%);"></div>
+              </div>
+              <div class="el-scrollbar__bar is-vertical">
+                <div class="el-scrollbar__thumb" style="transform: translateY(0%); -webkit-transform: translateY(0%);"></div>
+              </div>
+            </div>
+            <p class="el-select-dropdown__empty">
+              无数据
```
LachlanStuart: image

That explains it! I guess we're not setting the locale when running in Jest, so ElementUI is defaulting to Chinese.

PR #15, opened 2018-07-26
DB backed user storage, hashed passwords, tokens with expiry

intsco:
- Postgres database schema postgres created on startup. All user data stored in this `auth.user`.
- Passwords hashed using bcrypt
- User emails are sent using AWS SES
- Tests for user operations, including email sending, with jest
- **graphql config updated (test one as well)**

### DIFF
--- a/metaspace/graphql/src/modules/auth/db-user.spec.ts
+++ b/metaspace/graphql/src/modules/auth/db-user.spec.ts
@@ -1,4 +1,5 @@
 import config from '../../utils/config';
+import {createExpiry} from "./utils";
 import {knex,
   initSchema,
   takeFirst} from './db';
@@ -47,7 +48,7 @@ describe('Database operations with user', () => {
   });
 
   afterEach(async () => {
-    await knex('user').delete();
+    await knex.from('user').truncate();
   });
 
   test('create absolutely new user', async () => {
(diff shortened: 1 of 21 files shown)


metaspace/graphql/src/modules/auth/db-user.spec.ts
```diff
@@ -0,0 +1,234 @@
...
+    // await knexAdmin.destroy();
+    // await require('./db').knex.destroy();
+  });
+
+  beforeEach(async () => {
+  });
+
+  afterEach(async () => {
+    await knex.from('user').truncate();
+  });
```
LachlanStuart: I think updating `knex` fixed part of the issue with database connections staying open. 

Running the tests in this branch I get "Jest did not exit one second after the test run has completed.", but if I add the following code, Jest closes normally:
```
  afterAll(async () => {
    await knex.destroy();
  });
```

metaspace/graphql/src/modules/auth/db-user.ts
```diff
@@ -0,0 +1,188 @@
...
+      resetPasswordTokenExpires: createExpiry()
+    };
+    await updateTable('user', updUser);
+  }
+  else {
+    resetPasswordToken = user.resetPasswordToken;
+  }
+  const link = `${config.web_public_url}/#/account/reset-password?email=${encodeURIComponent(email)}&token=${encodeURIComponent(resetPasswordToken)}`;
+  emailService.sendResetPasswordEmail(email, link);
+  console.log(`Sent password reset email to ${email}: ${link}`);
```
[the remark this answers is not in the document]
intsco: Yep, I was thinking about a logging library too

metaspace/graphql/src/modules/auth/db-user.spec.ts
intsco: I already updated `afterAll` method

PR #19, opened 2018-08-02
Group profile page and menu changes

LachlanStuart:
Asana task: https://app.asana.com/0/474912821417535/739618576152066

This adds UI for users to be able to see a group and its datasets, request to join the group and import previously submitted datasets as they join a group.

Changes:
* Added UI for seeing a group, requesting access to a group, importing datasets into a group
* Refactored the menu bar a bit to add the user's current group, and a sub-menu to hold the Log Out link
* Added a `hasGroup` filter, a `Dataset.uploadDT` field, and stub code for the `currentUser` query to graphql
* Made logging in/out refresh all graphql data on the page (via `refreshLoginStatus`)
* Fixed a bug that caused the summary plots to not render. I think this was introduced with the `vue-apollo` upgrade

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@ jobs:
         environment:
           POSTGRES_USER: sm
           POSTGRES_PASSWORD: password
-      
+
       - image: node:8
 
     working_directory: ~/metaspace/metaspace/graphql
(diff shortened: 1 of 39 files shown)


Ren22:
@LachlanStuart wanted to merge it into master , but there is a hanging conflict with `SignInDialog.vue`. 
Should I try to fix it or you will do it yourself?

LachlanStuart:
@Ren22 I'll merge it

PR #23, opened 2018-08-06
Better reporting of TypeScript errors

LachlanStuart:
* Adds a build step for webapp to CircleCI so that it 
* Adds codecov upload to CircleCI for webapp. Coverage is run as a separate pass from the normal tests because the `--coverage` instrumentation causes there to be inaccurate line numbers in stack traces, which would make debugging failed tests much.
* Outputs TypeScript errors to browser console, e.g.
image

* Turns off `vue-analytics` debug mode so that there's less spam in the browser console.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,7 +73,7 @@ jobs:
           name: Upload coverage
           command: |
             yarn run test --coverage
-            codecov -F webapp -p ../..
+            codecov --root ../.. -F webapp
 
   test-webapp-e2e:
     docker:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -60,10 +61,19 @@ jobs:
...
+            yarn run build-ci
       - run:
           name: Run tests
           command: |
             yarn run test
+      - run:
+          name: Upload coverage
+          command: |
+            yarn run test --coverage
+            codecov -F webapp
```
LachlanStuart: [Documentation on the "flags" / the `-F` parameter](https://docs.codecov.io/docs/flags)

PR #25, opened 2018-08-12
Migration: merges all additional info fields in metadata into one 

Ren22:
called 'Supplementary'. Tracks problematic datasets.


metaspace/engine/scripts/merge_additional_info.py
```diff
@@ -0,0 +1,58 @@
...
+            for k in entriesToRemove:
+                if (metadata['Additional_Information'][k] != ''):
+                    allVals = allVals + metadata['Additional_Information'][k] + ' '
+                metadata['Additional_Information'].pop(k, None)
+            metadata['Additional_Information']['Supplementary'] = allVals
+            db.alter(UPDATE_METADATA, params=(json.dumps(metadata), datasetID,))
+        else:
+            for k in entriesToRemove:
+                if k in metadata['Additional_Information']:
+                    logs = logs + ('True | PROPERTY {} DOES EXISTS ON Additional_Information | ID={}\n'.format(k, datasetID))
```
Ren22: a small grammar mistake which shouldn't affect the meaning

LachlanStuart:
@intsco Please re-review this. I've updated it to use `sm.engine.dataset` so that it also updates the ES index, and changed the way the fields are combined so that it's possible to tell the original fields apart in the combined field.

PR #28, opened 2018-08-22
GraphQL cleanups & schema for Projects

LachlanStuart:
* Adds Project schema
* Removes `bringDatasets` parameters in favor of a dedicated `importDatasetsIntoGroup` mutation
* Adds `urlSlug` field & note that it can be used instead of `id` for the `group` and `project` queries


metaspace/graphql/schemas/group.graphql
```diff
@@ -33,7 +36,7 @@ enum UserGroupRole {
 }
 
 type Query {
-  group(groupId: ID!): Group
+  group(groupId: ID!): Group           # If a non-UUID string is used for groupId, the group will be looked up by urlSlug instead
```
intsco: Shall we check the input string format or introduce two optional arguments (groupId, groupSlug)?

metaspace/graphql/schemas/project.graphql
```diff
@@ -0,0 +1,63 @@
...
+  project: Project!
+  role: UserProjectRole!
+  numDatasets: Int!
+}
+
+enum UserProjectRole {
+  INVITED    # User was invited into the project but hasn't confirmed it yet
+  PENDING    # User requested to join the project but the PI hasn't confirmed it yet
+  MEMBER     # Normal member of the project
+  ADMIN      # Project administrator
```
intsco: `admin` or `manager`? It would be convenient to reserve the term `admin` for metaspace administrators

metaspace/graphql/schemas/project.graphql
```diff
@@ -0,0 +1,63 @@
...
+}
+
+type Query {
+  project(projectId: ID!): Project           # If a non-UUID string is used for projectId, the project will be looked up by urlSlug instead
+  allProjects(query: String): [Project!]!
```
intsco: What kind of queries are we expecting to be passed to `allProjects`?

metaspace/graphql/schemas/group.graphql
LachlanStuart: I wouldn't handle it through optional arguments, it would be much cleaner to have a separate `groupBySlug(urlSlug: String!)` query to prevent the ambiguous case where both or neither argument was supplied.

The main difference is whether the ID/urlSlug distinction is determined on the client side or server side. "Explicit is better than implicit", so I'll add a 2nd query.

metaspace/graphql/schemas/project.graphql
LachlanStuart: [I only have plans for text-based search.](https://app.moqups.com/LachlanStuart/liDSQ2TVih/view/page/a36ada7c0) Do you think I should turn it into an extensible structure like we have with `DatasetFilter`?

intsco: I don't know, maybe not. It just doesn't say explicitly that the search will be by name.

metaspace/graphql/schemas/group.graphql
intsco: Then we'll need the same for projects

metaspace/graphql/schemas/project.graphql
LachlanStuart: Ideally it would also search the description field(s) when they come

metaspace/graphql/schemas/project.graphql
intsco: what about this?

LachlanStuart: Makes sense, change pushed.

metaspace/graphql/schemas/group.graphql
LachlanStuart: Done

metaspace/webapp/src/api/project.ts
```diff
@@ -0,0 +1,130 @@
+import gql from 'graphql-tag';
+
+export type ProjectRole = 'INVITED' | 'PENDING' | 'MEMBER' | 'ADMIN';
```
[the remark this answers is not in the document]
LachlanStuart: > Actually tt would also be convenient to take these values from one place (for graphql and web app).

I've been thinking the same thing. `webapp` already partially depends on the merged schema from `graphql`; the output from `gen-binding` would also be useful.

PR #31, opened 2018-08-24
User resolvers

intsco:
- Upgraded Apollo Server to ^2.*, graphql-tools to ^3.*
- Pulled out schema generation into separate modules
- Fixed mol_db role provisioning in Ansible
- New provisioning steps for the graphql project (database schema and extension)
- User and Group models and resolvers


metaspace/graphql/.gitignore
```diff
@@ -1,3 +1,9 @@
 config/*.js
 !config/test.js
 metadata_schema.json
+.idea/
+.vscode/
+node_modules/
```
LachlanStuart: `.idea`, `.vscode` and `node_modules` should have been inherited from the `.gitignore` in the root of the repository

metaspace/graphql/schemas/user.graphql
```diff
@@ -1,8 +1,8 @@
 type User {
   id: ID!
   name: String!
-  role: String            # null if current user is not allowed to see
-  email: String           # null if current user is not allowed to see
+  role: String!            # null if current user is not allowed to see
+  email: String!           # null if current user is not allowed to see
```
LachlanStuart: As discussed, `Dataset.submitter` should be changed to use something other than `User` as it only needs `id` and `name`. 

Also, I think `email` should stay nullable because `Group.members` and `Project.members` are `User`s, but users generally shouldn't be able to see each others' roles and email addresses unless they have some sort of Admin/PI/Manager status.

intsco:
@LachlanStuart this PR is getting huge so we should finally merge it.

metaspace/graphql/src/modules/auth/controller.ts
```diff
@@ -66,18 +65,15 @@ const configureJwt = (app: Express) => {
...
       payload = {
-        'iss': 'METASPACE2020',
-        'sub': user.id,
-        'name': user.name,
-        'email': user.email,
-        'iat': nowSeconds,
-        'exp': expSeconds == null ? undefined : nowSeconds + expSeconds,
-        'role': user.role,
+        iss: 'METASPACE2020',
+        user,
```
LachlanStuart: It looks like this `user` field would include the credentials. IMO you should pick out the specific fields you need out of `user` here, to avoid anything accidentally leaking in, i.e.
```typescript
const {id, name, email, role} = user;
payload = {
  iss: 'METASPACE2020',
  user: {id, name, email, role},
```

intsco: Credentials won't be included here but specifying fields explicitly sound reasonable.

PR #36, opened 2018-08-31
Fixes

LachlanStuart:
A bunch of fixes in preparation to deploy to staging. See individual commits for details.

### DIFF
--- a/metaspace/graphql/src/modules/auth/controller.ts
+++ b/metaspace/graphql/src/modules/auth/controller.ts
@@ -184,10 +184,9 @@ const configureCreateAccount = (app: Express) => {
 
   app.get('/api_auth/verifyemail', preventCache, async (req, res, next) => {
     const {email, token} = req.query;
-    // TODO: Better handling for when the user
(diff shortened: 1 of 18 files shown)


metaspace/graphql/src/modules/auth/controller.ts
```diff
@@ -184,10 +184,9 @@ const configureCreateAccount = (app: Express) => {
 
   app.get('/api_auth/verifyemail', preventCache, async (req, res, next) => {
     const {email, token} = req.query;
-    // TODO: Better handling for when the user
-    const user = await verifyEmail(email, token);
-    if (user) {
-      req.login(user, (err) => {
+    const userId = await verifyEmail(email, token);
```
intsco: I am not sure if it's a good idea to return `user id` anymore since now `/api_auth/verifyemail` is inconsistent with `/api_auth/signin`

LachlanStuart: Good point. I'll fix it in another PR

PR #37, opened 2018-09-03
Add codecov config

LachlanStuart:


.codecov.yml
```diff
@@ -0,0 +1,22 @@
+codecov:
+  notify:
+    require_ci_to_pass: yes
+
+comment: false
```
LachlanStuart: @intsco Do you use the comments that Codecov adds to every PR? If not, I'd like to remove them as I find the notifications annoying. This flag disables them.

.codecov.yml
```diff
@@ -0,0 +1,22 @@
...
+    project:
+      engine:
+        range: 70...100
+        threshold: 0.2
+        flags: engine
+      webapp:
+        range: 40...100
+        threshold: 0.2
+        flags: webapp
+    patch: no
```
LachlanStuart: The "patch" metric only considers the coverage change since the last commit. It's pretty useless for PRs, since PRs should be considered as a whole.

.codecov.yml
```diff
@@ -0,0 +1,22 @@
...
+  precision: 2
+  round: down
+
+  status:
+    project:
+      engine:
+        range: 70...100
+        threshold: 0.2
+        flags: engine
+      webapp:
```
LachlanStuart: Splitting this into `engine` and `webapp` means the two can pass/fail separately

.codecov.yml
```diff
@@ -0,0 +1,22 @@
...
+
+coverage:
+  precision: 2
+  round: down
+
+  status:
+    project:
+      engine:
+        range: 70...100
+        threshold: 0.2
```
LachlanStuart: This is probably the biggest change in this PR: allowing coverage to regress by up to 0.2% without causing a failure. This means it won't complain if you make a small change to uncovered code, or if you delete some covered code.

LachlanStuart:
I think `codecov/project/engine` is only failing because that project isn't in master

PR #38, opened 2018-09-03
Make verifyEmail return a User object

LachlanStuart:


metaspace/graphql/package.json
```diff
@@ -90,7 +90,7 @@
       "<rootDir>/tests"
     ],
     "transform": {
-      "^.+\\.(ts|tsx)?$": "ts-jest"
+      "^.+\\.(js|ts|tsx)?$": "ts-jest"
```
LachlanStuart: My change to make `utils.js` use ES6 modules broke the tests. Making `ts-jest` handle JS files fixes it, as TS will transpile ES6 modules in JS code

intsco: Using TS to transpile ES6 to JS is funny )

PR #43, opened 2018-09-05
Feature/help button

Ren22:


intsco:
@Ren22 having a screenshot would be awesome :)

Ren22:
@intsco makes sense :) 
screenshot from 2018-09-06 11-03-03
screenshot from 2018-09-06 11-03-30

intsco:
Actually, I don't know what is better to have a pop-up window or open the document in another browser tab.
We should also ask @LachlanStuart or @theodev

Ren22:
@intsco we discussed it with Lachlan before and agreed upon that pop window will be a better option. Imo it's good because it doesn't interfere the process of filling metadata by sending user to another page. But the minor thing that I can add is a button to open it a new window from pop-up, if user is struggling with the resolution

metaspace/webapp/src/modules/MetadataEditor/UploadPage.vue
```diff
@@ -277,16 +287,26 @@
...
+    justify-content: space-between;
+    padding: 5px 0;
   }
 
-  .el-button__metadata {
-    margin: auto 0;
-    padding: 20px 20px;
+  .md-editor-submit > button {
+    flex: 1;
+    margin: auto 5px;
```
LachlanStuart: The two buttons are different heights, and neither matches the height of the upload area. If you remove this `margin: auto`, you can line them up with the upload area by adding `align-items: stretch` to `.md-editor-submit`, and changing `.md-editor-submit`'s margins or padding to match the upload area's `.qq-uploader` element.

metaspace/webapp/src/modules/MetadataEditor/UploadPage.vue
```diff
@@ -1,5 +1,12 @@
 <template>
   <div class="md-editor">
+    <el-dialog
+      :visible.sync="helpDialog"
+      :lock-scroll="false"
+      append-to-body>
+      <embed :src="helpLink"
+             style="width:800px; height: 800px;">
```
LachlanStuart: These instructions shouldn't be embedded in the dialog. Because it's a long document, it feels cramped putting it in such a small window. They should open in a new window instead. Also, this fixed width & height doesn't look good on other screen resolutions:
image

Ren22: @LachlanStuart I have a feeling that opening this doc in a new window will result in not-reading this documents at all. Personally I wouldn't do it. Here is a good read:
https://www.smashingmagazine.com/2008/07/should-links-open-in-new-windows/
IMO, the optimal solution would be to revise this help doc and provide user with an opportunity to open it in a new window from this dialog.

intsco: We might have a special case here as the document indeed seems too small for this window.

Ren22:
@LachlanStuart do you think this PR can be merged?

PR #44, opened 2018-09-05
* Updated queries for the current version of Metaspace on staging

Ren22:


intsco:
@Ren22 too bad your merged this RP so fast )
I wanted to know why you replaced the default config to point at staging

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -68,14 +68,36 @@ def listQuery(self, field_name, query, variables={}, batch_size=50000):
...
+        group {
+          id
+          name
+          shortName
+          urlSlug
+          members {
+            user {
+              name
+            }
+            group {
```
intsco: @Ren22 what's the use-case of returning a group with members and info about all groups of those members?

Ren22:
@intsco ah, but that's not a problem to change it back to the normal metaspace config when the final release is ready

metaspace/python-client/metaspace/sm_annotation_utils.py
Ren22: @intsco we might want to have this information easily available I think. 
For that I'm still going to add few properties in SMDataset class.

intsco:
@Ren22 please delete branches after merging

PR #51, opened 2018-09-13
SM API - GraphQL - Web App interaction bug fixes

intsco:
* sm api: return ds_id on create
* graphql: typos and small fixes
* groupIds stored in jwt
* use approved flag when saving a dataset
* moved dataset model into a separate module


intsco:
@LachlanStuart I actually don't know much about `generate-graphql-schema`

PR #53, opened 2018-09-13
Added a Jupyter notebook to upload a dataset example.

Ren22:


metaspace/python-client/example/upload_dataset_api.ipynb
```diff
@@ -0,0 +1,228 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Your API token:\n",
+    "\n",
+    "sm._config['jwt'] = '' # Please enter inside the quotes\n",
```
intsco: Try to avoid the need for changing private fields of an object. JWT token could be provided as an argument to SMIntance class constructor.

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -18,13 +14,16 @@ def _extract_data(res):
...
+            pprint.pprint(res_json)
+            raise Exception("Validaton of metadata failed, please check that all fields are correctly entered")
         raise Exception(res.json()['errors'][0]['message'])
 
 
 DEFAULT_CONFIG = {
     'graphql_url': 'http://metaspace2020.eu/graphql',
     'moldb_url': 'http://metaspace2020.eu/mol_db/v1',
-    'jwt': 'eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJNRVRBU1BBQ0UyMDIwIiwicm9sZSI6ImFub255bW91cyJ9.Hl0h6crcHLb-SPm7nomXkQco5l2iAO6D1bwdjmOaFXM'
+    'jwt': ''
```
intsco: Do we use a public anonymous jwt token or not?

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -206,40 +205,33 @@ def getRegisteredImage(self, dsid, zoom_level = 8):
         variables = {"datasetId": dsid}
         return self.query(query, variables)
 
-    def submitDataset(self, data_path, metadata, priority=0, dsid=None):
+    def createDataset(self, data_path, metadata, priority=0, dsName=None,
+                      isPublic=None, molDBs=None, adducts=None, dsid=None):
```
intsco: According to PEP8 function arguments as well as variables should be named using lowercase symbols with underscores.
https://www.python.org/dev/peps/pep-0008/#function-and-method-arguments
But that's the problem on all the code in `python-client` project.

intsco:
It's not mentioned anywhere that a user will need to have a AWS IAM account with write access to `sm-external-export` bucket.

metaspace/python-client/example/upload_dataset_api.ipynb
Ren22: hmm, that was a way to change jwt in config of an existing SMInstance object, initially it comes as an empty field now.. What can be the improvement for that?

metaspace/python-client/metaspace/sm_annotation_utils.py
Ren22: we don't need to specify it explicitly anymore, as it set up by default when user hits `/graphql`

metaspace/python-client/example/upload_dataset_api.ipynb
intsco: If we're talking about good coding style, I'd create a method set_jwt then

metaspace/python-client/metaspace/sm_annotation_utils.py
intsco: By looking at GraphQLClient code, I would tell that it will send `Authorization: Bearer ` if jwt not provided. And I think it's not valid header value.

Ren22: @intsco true, I'll fix it

metaspace/python-client/README.md
```diff
@@ -18,9 +18,21 @@ You need Python 3 to use this library.
 If you are trying to access annotations from the main public metaspace instance
 then make sure you include the branch name of the latest release branch, e.g.
 
-`pip install git+https://github.com/metaspace2020/metaspace@rel-v0.10#subdirectory=metaspace/python-client`
-
-The list of branches can be found [here](https://github.com/metaspace2020/metaspace/branches/all?&query=rel-v).
+`pip install git+https://github.com/metaspace2020/metaspace@release/v0.11-rc#subdirectory=metaspace/python-client`
+
+## Jupyter Playbooks
```
intsco: It's close to nit-picking but in Jupyter world they don't use the term Playbooks, they use Notebooks instead

Ren22:
@intsco do you think this branch is good to be merged?

PR #54, opened 2018-09-14
Python client/update dataset dbs

Ren22:


metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -467,6 +495,10 @@ def status(self):
     def _baseurl(self):
         return self._gqclient.url.rsplit("/", 1)[0]
 
+    @property
+    def adducts(self):
```
intsco: Duplicated property?

intsco: Line 475

Ren22: right, didn't see it, thanks!

PR #55, opened 2018-09-17
These changes add 2 extra fields to metadata sections(Pixel size)

Ren22:
Also removed +/- button signes for the input fields of the reolsving power


metaspace/metadata/min metadata schema.json
```diff
@@ -79,7 +79,7 @@
     "MS_Analysis": {
       "type": "object",
       "description": "Mass spectrometry settings",
-      "required": ["Polarity", "Ionisation_Source", "Analyzer"],
+      "required": ["Polarity", "Ionisation_Source", "Analyzer", "Pixel_Size"],
```
intsco: We should double check with @theodev whether we want to make `Pixel size` field required

metaspace/metadata/min metadata schema.json
```diff
@@ -116,11 +116,30 @@
...
               "default": 140000
             }
           }
+        },
+        "Pixel_Size": {
+          "type": "object",
+          "title": "The size of a pixel in um",
+          "required": ["PS_X", "PS_Y"],
+          "properties": {
+            "PS_X": {
```
intsco: The full path to this field will look like `...->Pixel_Size->PS_X`. Not sure if duplication is justified here.

metaspace/metadata/min metadata schema.json
Ren22: afair , Theo wanted these fields to be initially required , and it might entail changes to the existing datasets metadata

Ren22: @intsco

metaspace/metadata/min metadata schema.json
```diff
@@ -116,11 +116,30 @@
...
+              "maximum": 1000,
+              "default": 500
+            },
+            "PS_Y": {
+              "type": "number",
+              "minimum": 0,
+              "maximum": 1000,
+              "default": 400
+            }
+          }
```
LachlanStuart: This only applies to Imaging MS. It should be in `ims.json`.

metaspace/metadata/min metadata schema.json
Ren22: there are 2 fields actually , on for X, another for Y axis

metaspace/metadata/min metadata schema.json
```diff
@@ -116,11 +116,30 @@
...
-              "maximum": 1000000,
+              "minimum": 50000,
+              "maximum": 400000,
               "default": 140000
             }
           }
+        },
+        "Pixel_Size": {
+          "type": "object",
+          "title": "The size of a pixel in um",
```
LachlanStuart: `um` should be `μm`

metaspace/metadata/min metadata schema.json
```diff
@@ -116,11 +116,30 @@
             "Resolving_Power": {
               "description": "Peak FWHM / m/z",
               "type": "number",
-              "minimum": 10000,
-              "maximum": 1000000,
+              "minimum": 50000,
+              "maximum": 400000,
```
LachlanStuart: Why have these changed?

metaspace/webapp/src/modules/MetadataEditor/formStructure.ts
```diff
@@ -101,6 +103,8 @@ function getFieldType(prop: JsonSchemaProperty, propName: string): FormFieldEdit
       return 'person';
     } else if (prop.properties && prop.properties.Resolving_Power != null) {
       return 'detectorResolvingPower';
+    } else if (prop.properties && prop.properties.PS_X != null) {
+      return 'pixelSize';
```
LachlanStuart: It is better to add a `smEditorType` field to the metadata JSON than to try to infer the type by its properties.

metaspace/webapp/src/modules/MetadataEditor/inputs/DetectorResolvingPowerInput.vue
```diff
@@ -2,20 +2,22 @@
   <el-row>
     <el-col class="subfield" :span="12">
       <el-form-item :class="{'is-error': error && error.mz}" required>
-        <el-input-number
+        <el-input
+          type="number"
```
LachlanStuart: `<input type="number">` is not a good solution:
* It still shows the arrows. Every browser has different native controls, but pretty much all of them clash with Element UI's design.
* Its `value` is a string, not a number, which leads to validation errors:
image

A couple possible solutions:
* Make a new component that uses `<el-input type="text">` but converts the `value` to a number. Ideally also filtering out non-numeric key presses for good UX.
* Use `<el-input-number :controls="false">` to just hide the plus/minus buttons from the old component. This looks weird with all the extra space, but left-aligning the text and reducing the width of the inputs should improve it.

metaspace/webapp/src/modules/MetadataEditor/inputs/PixelSizeInput.vue
```diff
@@ -0,0 +1,63 @@
...
+    </el-col>
+  </el-row>
+</template>
+
+<script lang="ts">
+  import Vue from 'vue';
+  import { Component, Prop } from 'vue-property-decorator';
+  import {PixelSize} from '../formStructure';
+
+  @Component({name: 'pixel-size-input'})
```
LachlanStuart: There's no need to specify `name` on components. Vue figures out the name automatically based on the class name, and the name used when the component is registered.

metaspace/metadata/min metadata schema.json
LachlanStuart: Yep, we discussed it and agreed to make it required, with the reasoning that not many people edit old datasets, and people will always know the sizes for new datasets.

metaspace/metadata/min metadata schema.json
Ren22: There was a suggestion to change this range as these are mostly used numbers @LachlanStuart

metaspace/metadata/min metadata schema.json
intsco: okay

metaspace/metadata/min metadata schema.json
intsco: yep, then just having X and Y should be sufficient, right? :)

metaspace/metadata/min metadata schema.json
LachlanStuart: I remember Theo explaining that they were typical values, but I don't remember him asking for it to be changed.

On production we have 109 datasets with a resolving power less than 50,000, and 40 with a resolving power greater than 400,000. I don't think we should reduce the allowed values for this field.

Ren22:
@LachlanStuart if these changes are good, I'll also substitute two inputs in `Detector Resolving Power` with the new ones by the means of CustomNumberInput component

metaspace/webapp/src/modules/MetadataEditor/inputs/CustomNumberInput.vue
```diff
@@ -0,0 +1,44 @@
...
+	export default class CustomNumberInput extends Vue{
+
+  	@Prop({default: 0})
+    value!: number;
+
+    inputVal: any = this.value;
+    arrowsVals: string[] = ['ArrowLeft', 'ArrowRight'];
+
+    checkNumber(val: any){
+      let charCode = (val.which) ? val.which : val.keyCode;
```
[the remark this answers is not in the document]
Ren22: Thanks for suggestions @LachlanStuart 
I was looking for different ways of how to implement it and I agree that it's not the best solution for that as indeed `Ctrl+A, Ctrl+C` and other combinations are ignored.
Is there any other way better of how to check for numeric value while preserving other keyboard combinations? I'd be happy to implement it

Ren22:
@intsco there's no usage for this script as ims/lcms.json will be imported from `sm-webapp/src/assets` directory and they will be generated(and stay up-to-date) every time when web-app is restarted with webpack.

metaspace/webapp/src/modules/MetadataEditor/inputs/CustomNumberInput.vue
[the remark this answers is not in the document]
LachlanStuart: @Ren22 I do not recommend it, but using the `preventDefault` approach, the best I can come up with is this:
```typescript
checkNumber(event: KeyboardEvent) {
  if ((event.key || '').length === 1 && !event.ctrlKey && !event.altKey && !event.metaKey && !'0123456789-'.includes(event.key)) {
    event.preventDefault()
  }
}
```

This makes a few assumptions:
* All non-character keys have an `event.key` that is longer than 1 character. AFAICT this is a safe assumption as all special inputs have proper names like `'Backspace'` and `'ArrowLeft'`.
* All keyboard inputs with Ctrl, Alt or Meta will be somehow handled by the browser/OS instead of affecting the input.
* `'0123456789-'` includes all valid characters. If this input were used for floats, we'd have to care about the whether `,` or `.` is used by the user's current locale.
* `String.prototype.includes` is polyfilled, because it's not supported on IE. We polyfill it using polyfill.io.

However, my recommendation is still to just circumvent `<el-input>`'s bugginess by changing its `$data.currentValue` directly. There are fewer edge cases that way, as it handles cases where the input changes through non-keyboard interactions, such as clipboard pasted values and browser autofilled values.

metaspace/metadata/specialised metadata/ims.json
```diff
@@ -32,6 +32,33 @@
...
+              "type": "number",
+              "minimum": 0,
+              "maximum": 1000,
+              "default": 500
+            },
+            "Yaxis": {
+              "type": "number",
+              "minimum": 0,
+              "maximum": 1000,
+              "default": 400
```
LachlanStuart: Where did you get these default values? Wouldn't it be better to leave it blank so that people don't just leave it as the default value?

metaspace/graphql/dsMutation.js
```diff
@@ -45,6 +43,17 @@ function trimEmptyFields(schema, value) {
 }
 
 function validateMetadata(metadata) {
+  const ajv = new Ajv({allErrors: true});
+  const metadataType = metadata.Data_Type;
+  let validator = {};
+  let metadataSchema = {};
+  if (metadataType === 'Imaging MS') {
```
LachlanStuart: Use `metadataMapping.js` to find which schema to use instead of hard-coding `metadataType` values.

Ren22: should I copypaste this `metadataMapping.js` from webapp to use it in graphql?

Ren22: ah right, I just overlooked it, thanks! will do

Ren22:
@LachlanStuart updated this PR

LachlanStuart:
Please add `metadataSchemas/*.json` and `metadataSchemas/metadataMapping.js` to graphql's `.gitignore`

LachlanStuart:
Also please update the snapshots before merging

metaspace/webapp/src/modules/MetadataEditor/__snapshots__/MetadataEditor.spec.ts.snap
```diff
@@ -13,10 +13,6 @@ Object {
       "mz": 123,
     },
     "Ionisation_Source": "none",
-    "Pixel_Size": Object {
-      "Xaxis": 500,
-      "Yaxis": 400,
-    },
```
LachlanStuart: It looks like this snapshot file was updated on the wrong branch, because now it doesn't include the `Pixel_Size` section at all

intsco:
@Ren22 please del the branch

PR #58, opened 2018-09-18
Docker tweaks

LachlanStuart:
Changes:
* Expose elasticsearch, redis and postgres ports. I had previously hidden those ports while trying to get Docker to work as an isolated, single-port system. That never worked out, but I often need to directly connect to ElasticSearch and Postgres.
* Add an additional docker-compose configuration file to run docker images in a user account. This prevents them from creating files with root permissions on shared volumes.

After this is merged, to get the opened ports you'll just need to `dc kill` and `dc up -d` the affected containers. To run containers as your user account you'll need to update your `.env` file per the instructions in `.env.example` and drop and recreate the affected containers with:

```bash
dc kill sm-webapp sm-graphql sm-api sm-update-daemon sm-annotate-daemon sm-mol-db postgres
dc rm -f sm-webapp sm-graphql sm-api sm-update-daemon sm-annotate-daemon sm-mol-db postgres
sudo chown -R $USER:$USER .
dc up -d sm-webapp sm-graphql sm-api sm-update-daemon sm-annotate-daemon sm-mol-db postgres
```

### DIFF
--- a/docker/.env.example
+++ b/docker/.env.example
@@ -16,3 +16,11 @@ METASPACE_BRANCH=master
 
 # Ports
 WWW_PORT=8999
+
+# To run containers as a non-root user so that they don't change file permissions,
+# append `:docker-compose.uid.yml` to COMPOSE_FILE, and put your user ID
+# and group ID here, separated by a colon i.e. `SM_USER=<user id>:<group id>`.
+# You can find your user ID and group ID with the `id` command in Linux.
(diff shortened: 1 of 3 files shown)


LachlanStuart:
@Ren22 Yes, unfortunately. I wish the `docker-compose.yml` files were more configurable. We could make Docker configuration a lot tidier by using Ansible instead of Docker Compose, but there are better things to spend time on.

PR #63, opened 2018-09-21
* Changed optical image size from 25MB to 50MB

Ren22:


metaspace/webapp/src/modules/ImageAlignment/ImageAlignmentPage.vue
```diff
@@ -48,7 +48,7 @@
           </div>
 
           <div class="el-upload__tip" slot="tip">
-            JPEG file less than {{ limitMB }}MB in size
+            JPEG file less than {{ 50 }}MB in size
```
LachlanStuart: You need to change the value of `limitMB`, not just replace it in the text. The function `onFileChange` displays an error if the file bigger than `limitMB`.

Ren22: right, mistakenly hard-coded this one

Ren22:
@intsco can be merged to the next release branch(that is , master)

PR #64, opened 2018-09-21
Migration script for Data Management

LachlanStuart:
This script is an interactive process, and should be run step-by-step:
1. It grabs all datasets and makes a CSV file with all combinations of (Institution, Submitter Email, Submitter Name, PI Email, PI Name) for manual clean up. This is needed because it later makes users and groups based on the Submitters, PIs and Institutions, and the details need to be consistent.
2. After importing the cleaned up data it validates it and updates the datasets in the DB, then triggers an ES update for them.
3. It dumps a list of all combinations of Institution and Submitter/PI for manual clean up.
4. It uses the cleaned up Institutions / Users to add users to groups and assign them a role

This script will need some more work before it's ready for use:
* I'll need to double-check with someone else that the data clean-ups are valid
* After #51 is merged I need to check compatibility with the `ESExporter.update_ds` function
* It will need to be updated for any upcoming ElasticSearch and DB schema changes


intsco:
All looks good for one time run script. I'd propose not to merge this PR until we're done with other features for the release.

LachlanStuart:
I've pushed some changes:
* A bug fix to look up users by not_approved_email as well as email
* It now just completely removes the old Submitted_By fields from metadata, as we're no longer doing a live upgrade
* Made all datasets group_approved

LachlanStuart:
I'm merging this as the other features are mostly complete, so this is unlikely to have any further changes.

PR #65, opened 2018-09-24
User authorisation using scopes

intsco:
* User authorisation using scopes
* Refactored the way dataset filters applied
* CurrentUserLastSubmittedDataset resolver

### DIFF
--- a/metaspace/graphql/dsMutation.js
+++ b/metaspace/graphql/dsMutation.js
@@ -6,8 +6,8 @@ const jsondiffpatch = require('jsondiffpatch'),
   _ = require('lodash');
 
 const {logger, fetchEngineDS, fetchMolecularDatabases} = require('./utils.js'),
-  {Dataset: DatasetModel} = require('./src/modules/user/model'),
-  {UserGroup: UserGrouModel, UserGroupRoleOptions} = require('./src/modules/group/model'),
+  {Dataset: DatasetModel} = require('./src/modules/dataset/model'),
+  {UserGroup: UserGroupModel, UserGroupRoleOptions} = require('./src/modules/group/model'),
   metadataMapping = require('./metadataSchemas/metadataMapping').default;
 
 function isEmpty(obj) {
@@ -154,8 +154,12 @@ const isMemberOf = async (connection, user, groupId) => {
     userId: user.id,
     groupId
   });
-  return (!userGroup || ![UserGroupRoleOptions.MEMBER,
-    UserGroupRoleOptions.PRINCIPAL_INVESTIGATOR].includes(userGroup.role));
+  let isMember = false;
+  if (userGroup) {
+    isMember = [UserGroupRoleOptions.MEMBER,
+      UserGroupRoleOptions.PRINCIPAL_INVESTIGATOR].includes(userGroup.role);
+  }
+  return isMember;
 };
 
 const saveDS = async (connection, dsId, submitterId, groupId, approved) => {
--- a/metaspace/graphql/dsMutation.js
+++ b/metaspace/graphql/dsMutation.js
@@ -127,28 +127,6 @@ async function smAPIRequest(uri, body) {
   }
 }
 
-const hasEditAccess = async (connection, user, dsId) => {
-  if (!user)
-    throw new UserError('Access denied');
-
-  if (user.role === 'admin')
-    return;
-
-  if (dsId) {
-    const ds = await connection.getRepository(DatasetModel).findOne({
-      id: dsId
-    });
-    if (!ds)
-      throw new UserError(`DS ${dsId} does not exist`);
-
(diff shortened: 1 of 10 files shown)


metaspace/graphql/esConnector.js
```diff
@@ -271,21 +273,18 @@ module.exports.esCountGroupedResults = function(args, docType, user) {
...
-async function getById(docType, id, user, ignorePermissions=false) {
-  const resp = await es.get({ index: esIndex, type: docType, id, ignore: [404] });
-  if (!resp.found) {
-    return null;
-  } else if (ignorePermissions || canUserViewEsDataset(resp, user)) {
-    return resp;
-  } else {
-    throw new Error(`Unauthorized: user ${user.email} tried to access ${docType} ${id}`)
-  }
+async function getById(args, docType, user) {
```
LachlanStuart: Now that this no longer handles the "byId" part of the query, maybe "getFirst" or "getOneDocument" would be a better name.

intsco:
@LachlanStuart please approve if everything is fine

metaspace/graphql/src/modules/group/controller.ts
```diff
@@ -33,12 +15,36 @@ const findUserByEmail = async (connection: Connection, email: string) => {
...
+
 export const Resolvers = {
   UserGroup: {
     async numDatasets(userGroup: UserGroupModel, _: any, {connection}: any) {
-      return await connection.getRepository(DatasetModel).count({
-        where: {userId: userGroup.userId}
+      const num = await connection.getRepository(DatasetModel).count({
+        where: {userId: userGroup.user.id}
       });
+      return num || 0;
```
LachlanStuart: Is there a reason for this? `Repository.count` looks like it always resolves to a number.

metaspace/graphql/src/modules/group/controller.ts
```diff
@@ -53,38 +59,56 @@ export const Resolvers = {
...
+      const userGroups = userGroupModels.map(ug => ({
+        user: ug.user,
+        group: ug.group,
+        role: ug.role
+      }));
+      return userGroups.map(ug => ({
+        user: {...ug.user, scopeRole},
+        group: ug.group,
+        role: ug.role
+      }));
```
LachlanStuart: These two `map`s could easily be combined

metaspace/graphql/dsMutation.js
```diff
@@ -187,15 +191,20 @@ const toSMAPIparam = (obj) => {
   return smAPIUpdate;
 };
 
+const isAuth = (user) => {
```
LachlanStuart: Two comments, though not necessarily requests for changes:
* It's more readable if you have a consistent and descriptive pattern for authentication checks, and readability helps to make sure you don't accidentally miss a check or use the wrong check. I personally use the pattern `(assert?)Can(action)(entity)`, e.g. `assertCanReadGroup` to throw an exception if the user can't read the group, or `canEditProject` to return `true`/`false`.
* When possible, it's better to build permissions checks into the functions used to access data. For a lot of TypeORM data access this isn't pragmatic as it would require an additional, unwanted abstraction layer over the ORM. But for data access where we have our own abstraction layer (e.g. context, mol-db/sm-api and ElasticSearch), it makes it a lot harder to forget authorization checks if the functions are structured so that you cannot access the data without making an authorization check, e.g. `context.getUserOrFail()` to get the current user (/`getUserOrNull()` for the non-asserting variant). `esDatasetByID` is a good example of this - there's no way to make it accidentally return data without an authorization check.

intsco: I was still looking for a way to implement this properly. Thanks for the guideline.

metaspace/graphql/resolvers.js
```diff
@@ -264,32 +290,17 @@ const Resolvers = {
...
-        .select('id')
-        .first();
-      if (lastDataset != null) {
-        return await esDatasetByID(lastDataset.id, user);
-      } else {
-        return null;
+      const results = await esSearchResults({
+        orderBy: 'ORDER_BY_DATE',
+        sortingOrder: 'DESCENDING',
+        limit: 1,
```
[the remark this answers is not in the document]
intsco: Actually, all requests sent to ES go through authorisation checks in the form of additional filters applied based on `user` object.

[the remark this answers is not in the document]
LachlanStuart: `esSearchResults` applies an "Only results that I'm allowed to see" filter. This function needs to add a "Only results that I submitted" filter. 

The issue isn't authorization. This function has to return the last dataset that was submitted by the current user. It is used for pre-filling the upload page metadata.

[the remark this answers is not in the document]
intsco: True, missed it.

PR #66, opened 2018-09-25
Read-only mode

LachlanStuart:
This adds a new "System Health" module to GraphQL, which allows admins to:
* Turn off all GraphQL mutations
* Turn off adding datasets to the processing queue
* Show a maintenance message

It also adds UI for setting the health, showing the health to users, and reacting to situations where the server rejects a request due to being in read-only mode.

It does not attempt to 

I've tried to keep this close to the patterns we have in `master`, but there will probably still be some pain merging, so I'll do that in a second PR after this is merged. In particular:
*  `graphqlClient.ts` uses `apollo-link` in `master`, which will require a completely different implementation. 
* the "Create Account", "Reset Password", etc. functions in the new authentication system will also need to be disabled. There's no point in doing it in this PR as it is significantly restructured in `master`.

# Screenshots

## Admin page ( http://localhost:8888/#/admin/health )
image
#### Maintenance message:
image
#### Read-only mode:
image
(This is the default message in Read-only mode, but it can be overridden. "Disable dataset upload/reprocessing" does not have any default message, but it's visible on the Upload page)

## Upload Page (existing behavior, new data source)

image

## Metadata Editor in "Disable dataset upload/reprocessing" mode

This only shows up if you make an edit that would trigger reprocessing:
image

## Errors

The global error handler in `graphqlClient.ts` can't easily suppress the error handlers that wrap most GraphQL code. I've tried to address this by making `reportError` detect and ignore errors that have already been handled. Usually you'll only see this popup:

image

However if I've missed any custom error handlers, it is potentially still possible to get two error messages at once e.g.:
image

### DIFF
--- a/metaspace/webapp/src/App.vue
+++ b/metaspace/webapp/src/App.vue
@@ -90,7 +90,6 @@
  }
 
  .main-content {
-   padding-top: 70px;     /* see MetaspaceHeader.vue: 70 = 62 height + 8 margin */
    /* padding-bottom: 130px; see MetaspaceFooter.vue: 130 = 30 height + 100 margin */
  }
 
--- a/metaspace/webapp/src/components/DatasetItem.vue
+++ b/metaspace/webapp/src/components/DatasetItem.vue
@@ -132,6 +132,7 @@
  import {deleteDatasetQuery, thumbnailOptImageQuery} from '../api/dataset';
  import {mdTypeSupportsOpticalImages} from '../util';
  import {encodeParams} from '../url';
+ import reportError from '../lib/reportError';
 
  function removeUnderscores(str) {
    return str.replace(/_/g, ' ');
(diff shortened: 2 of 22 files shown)


metaspace/webapp/src/App.vue
```diff
@@ -90,7 +90,6 @@
  }
 
  .main-content {
-   padding-top: 70px;     /* see MetaspaceHeader.vue: 70 = 62 height + 8 margin */
```
LachlanStuart: Instead of using padding here to prevent content appearing behind the header, I've added a non-floating element to `MetaspaceHeader.vue` that uses its height to push the rest of the content down the page.

metaspace/webapp/src/components/DatasetItem.vue
```diff
@@ -325,14 +326,8 @@
...
        catch (err) {
-         this.$message({
-           message: "Deletion failed :( Please contact us at contact@metaspace2020.eu",
-           type: 'error',
-           duration: 0,
-           showClose: true
-         });
          this.disabled = false;
-         throw err;
+         reportError(err, "Deletion failed :( Please contact us at contact@metaspace2020.eu");
```
LachlanStuart: I've left this message in because deletion is a bit sensitive - if someone is trying to delete their data, they probably are already angry. It's better to offer a slow-but-certain solution than the usual "refresh the page and try again".

PR #68, opened 2018-09-25
Feature/metadata section updates

Ren22:
There's still a problem which I haven't solved when the user changes the page in annotations table. Somehow all values are wiped out from the parent component or perhaps there's another reason why the accordion is expanded again.

### DIFF
--- a/metaspace/webapp/src/components/DatasetInfo.vue
+++ b/metaspace/webapp/src/components/DatasetInfo.vue
@@ -13,13 +13,14 @@
 <script>
   import {defaultMetadataType, metadataSchemas} from '../assets/metadataRegistry';
   import {get, flatMap} from 'lodash-es';
+  import { safeJsonParse } from '../util'
 
   export default {
     name: 'dataset-info',
     props: ['metadata', 'expandedKeys'],
     data() {
       return {
-        expandedTreeNodes: JSON.parse(localStorage.getItem('expandedTreeNodes')) || [],
+        expandedTreeNodes: safeJsonParse(localStorage.getItem('expandedTreeNodes')) || [],
       };
     },
     created() {
--- a/metaspace/webapp/src/components/DatasetInfo.vue
+++ b/metaspace/webapp/src/components/DatasetInfo.vue
@@ -31,19 +31,19 @@
 
     computed: {
       dsGroup() {
-        return this.$store.state.annotation.dataset.group
+        return this.metadata.Group
       },
 
       dsSubmitter() {
-        return this.$store.state.annotation.dataset.submitter
+        return this.metadata.Submitter
       },
 
       dsProjects() {
-        return this.$store.state.annotation.dataset.projects
+        return this.metadata.Projects
       },
 
       dsPI() {
-        return this.$store.state.annotation.dataset.principalInvestigator
+        return this.metadata.PI
       },
 
       schema() {
--- a/metaspace/webapp/src/components/DatasetInfo.vue
+++ b/metaspace/webapp/src/components/DatasetInfo.vue
@@ -1,25 +1,40 @@
 <template>
   <el-row>
-    <el-tree :data="treeData"
+    <el-tree :data="getTreeData()"
              id="metadata-tree"
              node-key="id"
-             :default-expanded-keys="defaultExpandedKeys">
+             default-expand-all>
     </el-tree>
   </el-row>
 </template>
 
 <script>
   import { defaultMetadataType, metadataSchemas } from '../assets/metadataRegistry';
-  import {get} from 'lodash-es';
+  import { groupsProjectsQuery } from '../../src/api/dataset';
+  import { get } from 'lodash-es';
 
  export default {
    name: 'dataset-info',
-   props: ['metadata', 'expandedKeys'],
+   props: ['metadata'],
    data() {
-     return {
-       defaultExpandedKeys: this.expandedKeys
-     }
+     	return {
+	      dsGroup: this.$store.state.annotation.dataset.group,
+        dsSubmitter: this.$store.state.annotation.dataset.submitter,
+	      dsProjects: this.$store.state.annotation.dataset.projects
+      }
    },
+	 apollo: {
+		 groupsProjects: {
+			 query: groupsProjectsQuery,
+			 skip: true,
+			 variables() {
+				 return {datasetId: this.$store.state.annotation.dataset.id}
+			 },
+			 update(data) {
+			 	 // console.log('Call')
+       }
+		 }
+	 },
    computed: {
      schema() {
        const metadataType = get(this.metadata, 'Metadata_Type')
@@ -27,12 +42,13 @@
          || defaultMetadataType;
        return metadataSchemas[metadataType];
      },
-     treeData() {
-       return this.objToTreeNode(null, this.metadata, this.schema);
-     }
-   },
-   methods: {
 
+	   isSignedIn() {
+		   return this.$store.state.authenticated;
+	   }
+   },
+
+   methods: {
      prettify(str) {
        return str.toString()
                  .replace(/_/g, ' ')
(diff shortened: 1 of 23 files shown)


metaspace/webapp/src/components/DatasetInfo.vue
```diff
@@ -1,37 +1,94 @@
...
+			   { id: "Submitter", label: `Submitter: ${this.dsSubmitter.name}`},
+			   { id: "Principal Investigator", label: `Principal Investigator: ${this.dsPI.name}`},
+			   { id: "Group", label: `Group: ${this.dsGroup.name}` }
+		   ];
+		   if (this.isSignedIn) {
+			   dataManagementChilds.push({ id: "Projects", label: `Projects: ${allProjects}` })
+		   }
+		   this.schemaBasedVals.push({id: "Data Management", label: "Data Management", children: dataManagementChilds});
+		   return this.schemaBasedVals;
+	   }
```
LachlanStuart: There are lots of mixed tabs and spaces throughout this file. What are you using to edit it?

metaspace/webapp/src/components/DatasetInfo.vue
```diff
@@ -1,37 +1,94 @@
...
        const metadataType = get(this.metadata, 'Metadata_Type')
          || this.$store.getters.filter.metadataType
          || defaultMetadataType;
        return metadataSchemas[metadataType];
      },
-     treeData() {
-       return this.objToTreeNode(null, this.metadata, this.schema);
-     }
+
+	   getTreeData() {
```
LachlanStuart: `computed` properties are accessed like properties, not methods, and thus should describe the data they hold, not the action they perform. The old name, `treeData` was better.

metaspace/webapp/src/components/DatasetInfo.vue
```diff
@@ -1,37 +1,94 @@
...
  export default {
    name: 'dataset-info',
    props: ['metadata', 'expandedKeys'],
    data() {
-     return {
-       defaultExpandedKeys: this.expandedKeys
+     	return {
+	      schemaBasedVals: {},
+	      expandAll: false,
+	      expandedTreeNodes: [] || this.expandedKeys,
```
LachlanStuart: This makes no sense. `[]` is never falsey, so `this.expandedKeys` is never used. If the prop isn't needed anymore, it should be removed.

metaspace/webapp/src/components/DatasetInfo.vue
```diff
@@ -1,37 +1,94 @@
...
+  import { get } from 'lodash-es';
 
  export default {
    name: 'dataset-info',
    props: ['metadata', 'expandedKeys'],
    data() {
-     return {
-       defaultExpandedKeys: this.expandedKeys
+     	return {
+	      schemaBasedVals: {},
```
LachlanStuart: The dataflow in this class is very confusing and needs to be refactored. The main problem is that there are a lot of values that can easily be `computed` that are instead stored as `data` and initialized at weird times. This contributes to a variety of other problems:
* `getTreeData` should not change the value of `this.schemaBasedVals`, because `computed` properties are cached and lazily updated. It's not predictable when they run. It also means that watched values change during render, which can trigger unnecessary re-renders.
* `schemaBasedVals` should not change type. It's initialized as `{}` but later set to be an array. 
* `mounted` should not change data if that data is already available when `data` is called.
* After `getTreeData` runs, `schemaBasedVals` contains more than just the schema-based nodes. Its name should reflect this. The original name `treeData` was fine.
* Because `dsSubmitter`, `dsPI`, `dsGroup`, and `dsProjects` are only set during `data`, they don't update when you change annotation. They should be `computed`s.
* `getChildsWOLeafs` should ideally not modify its arguments - the purpose of the function is to output an array, so ideally that array should be *returned*, not appended to one of the arguments. If it must modify its arguments (which I think is inappropriate here), then its name needs to make that obvious. `get` functions return values. A name like `appendNonLeafNodeIds` would be better.
    This is how it would look if it returned its output value:
    ```js
    import {flatMap} from 'lodash-es';

    // used like this:
    expandedTreeNodes = this.getNonLeafNodeIds(this.treeData);

    getNonLeafNodeIds(node) {
        if (Array.isArray(node.children) && node.children.length) {
            return [node.id, ...flatMap(node.children, this.getNonLeafNodeIds)];
        } else {
            return [];
        }
    }
    ```
    After doing this you could even wrap it into a `computed` so that you didn't have to worry about the order things were initialized in:
    ```js
    computed: {
        nonLeafNodeIds() {
            return this.getNonLeafNodeIds(this.treeData);
        }
    }
    ```

Ren22:
@LachlanStuart I think formatting should be better now

metaspace/webapp/src/components/DatasetInfo.vue
```diff
@@ -1,45 +1,102 @@
...
      schema() {
        const metadataType = get(this.metadata, 'Metadata_Type')
          || this.$store.getters.filter.metadataType
          || defaultMetadataType;
        return metadataSchemas[metadataType];
      },
-     treeData() {
-       return this.objToTreeNode(null, this.metadata, this.schema);
+
+     getTreeData() {
```
LachlanStuart: `computed` properties are accessed like properties, not methods, and thus should describe the data they hold, not the action they perform. The old name, `treeData` was better.

metaspace/webapp/src/components/DatasetInfo.vue
```diff
@@ -1,45 +1,102 @@
...
 
  export default {
    name: 'dataset-info',
    props: ['metadata', 'expandedKeys'],
    data() {
      return {
-       defaultExpandedKeys: this.expandedKeys
+       schemaBasedVals: {},
+       expandAll: false,
+       expandedTreeNodes: [] || this.expandedKeys,
```
LachlanStuart: This makes no sense. `[]` is never falsey, so `this.expandedKeys` is never used. If the prop isn't needed anymore, it should be removed.

metaspace/webapp/src/components/DatasetInfo.vue
```diff
@@ -1,45 +1,102 @@
...
+ import {defaultMetadataType, metadataSchemas} from '../assets/metadataRegistry';
+ import {get} from 'lodash-es';
 
  export default {
    name: 'dataset-info',
    props: ['metadata', 'expandedKeys'],
    data() {
      return {
-       defaultExpandedKeys: this.expandedKeys
+       schemaBasedVals: {},
```
LachlanStuart: The dataflow in this class is very confusing and needs to be refactored. The root cause is that there are a lot of values that could easily be `computed` but are instead stored as `data` and initialized at weird times. This contributes to a variety of other problems:
* `getTreeData` should not change the value of `this.schemaBasedVals`, because `computed` properties are cached and lazily updated. It's not predictable when they run. It also means that watched values change during render, which can trigger unnecessary re-renders.
* `schemaBasedVals` should not change type. It's initialized as `{}` but later set to be an array. 
* `mounted` should not change data if that data is already available when `data` is called.
* After `getTreeData` runs, `schemaBasedVals` contains more than just the schema-based nodes. Its name should reflect this. The original name `treeData` was fine.
* Because `dsSubmitter`, `dsPI`, `dsGroup`, and `dsProjects` are only set during `data`, they don't update when you change annotation. They should be `computed`s.
* `getChildsWOLeafs` should ideally not modify its arguments - the purpose of the function is to output an array, so ideally that array should be *returned*, not appended to one of the arguments. If it must modify its arguments (which I think is inappropriate here), then its name needs to make that obvious. `get` functions return values. A name like `appendNonLeafNodeIds` would be better.
    This is how it would look if it returned its output value:
    ```js
    import {flatMap} from 'lodash-es';

    // used like this, ideally in data(), not in mounted():
    expandedTreeNodes = this.getNonLeafNodeIds(this.treeData);

    getNonLeafNodeIds(node) {
        if (Array.isArray(node.children) && node.children.length) {
            return [node.id, ...flatMap(node.children, this.getNonLeafNodeIds)];
        } else {
            return [];
        }
    }
    ```

metaspace/webapp/src/components/DatasetInfo.vue
```diff
@@ -1,45 +1,102 @@
...
+       this.schemaBasedVals = this.objToTreeNode(null, this.metadata, this.schema);
+       let allProjects = this.dsProjects.map(e => e.name).join(', ');
+       let dataManagementChilds = [
+         {id: "Submitter", label: `Submitter: ${this.dsSubmitter.name}`},
+         {id: "Principal Investigator", label: `Principal Investigator: ${this.dsPI.name}`},
+         {id: "Group", label: `Group: ${this.dsGroup.name}`}
+       ];
+       if (this.isSignedIn) {
+         dataManagementChilds.push({id: "Projects", label: `Projects: ${allProjects}`})
+       }
```
LachlanStuart: A dataset's projects should be shown regardless of whether the current user is signed in. It would, however, be good to hide `Group` and `Projects` if they were empty.

metaspace/webapp/src/components/DatasetInfo.vue
Ren22: cool, thanks for lots of suggestions and comments, I'll fix them in code

metaspace/webapp/src/components/DatasetInfo.vue
Ren22: I think per task in Asana we wanted initially to hide the projects 
image

LachlanStuart: Yes, but only the private projects. That should be handled on the server side. If a dataset has a mix of public and private projects, graphql will return a list containing only the projects that the current user is allowed to see.

Ren22:
@LachlanStuart thanks for fixing this annoying test bug

PR #70, opened 2018-09-27
Fix addReadOnlyInterceptorToSchema

LachlanStuart:


metaspace/graphql/src/modules/system/addReadOnlyInterceptorToSchema.ts
```diff
@@ -49,7 +26,7 @@ const addReadOnlyInterceptorToSchema = (schema: GraphQLSchema) => {
         if (!canMutate || (!canProcessDatasets && mutationRequiresDatasetProcessing(mutationName, args))) {
           throw new UserError(JSON.stringify({ type: 'read_only_mode', message }));
         } else {
-          return originalResolve.apply(this, ...arguments);
+          return originalResolve.apply(this, arguments);
```
LachlanStuart: The issue was here. `apply` automatically spreads its 2nd parameter.

PR #73, opened 2018-10-01
Fix ES filters, user auth mutations

intsco:
* Support of information about PI, stored in graphql.dataset table.
* User/group updated authorisation assertions.
* Dataset mutations cleaned and updated.
* Fetching metadata and submitter suggestions from ES (not DB).
* Optical images authorisation and refactoring.

### DIFF
--- a/metaspace/graphql/resolvers.js
+++ b/metaspace/graphql/resolvers.js
@@ -2,7 +2,8 @@ import {UserError} from 'graphql-errors';
 import fetch from 'node-fetch';
 import * as _ from 'lodash';
 import * as config from 'config';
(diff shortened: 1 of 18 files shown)


metaspace/graphql/resolvers.js
```diff
@@ -357,19 +357,27 @@ const Resolvers = {
     },
 
     async principalInvestigator(ds, _, {connection}) {
```
intsco: @LachlanStuart does this resolver returns what you expect?

LachlanStuart: @intsco Yes

intsco:
@LachlanStuart please review when you have time

PR #74, opened 2018-10-02
UI Fixes 2

LachlanStuart:
This is best reviewed commit-by-commit, as they are mostly independent changes.

Many of the bugs were caused by Apollo Cache issues. The problem is that by default `InMemoryCache` will "normalize" objects based on their type and ID. So if one query gets `currentUser`, which returns `{id: '123', primaryGroup: {/*something*/}}`, then another query gets `dataset { submitter { primaryGroup } }`, which returns `{submitter: { primaryGroup: null } }`, then the initial `currentUser` query's result actually gets overwritten with the `primaryGroup: null` change. I've done several things to fix this:
* Explicitly exclude the `User` GraphQL type from normalization. `InMemoryCache` will instead cache users by their parent object or path in the GraphQL query used to fetch them. The above two examples will thus have different cache keys: `ROOT_QUERY.currentUser` and `Dataset(123).submitter`
* Disable loading data from the cache by default. This will mean that most pages re-fetch data when you open them. I've selectively added back caching to things that don't update as often.
* For optical images it was too much work to intelligently clear the cache for just the updated dataset, so I instead changed it to clear the entire cache when optical images are uploaded

Note that the cache issues only showed up in the `feature/user-auth-mutations` branch due to its `scopeRole` changes. It would be best to test in a local copy that has that branch merged in.

## Other reported bugs fixed:
* Fixed `urlSlug` being incorrectly sent when updating other fields in groups & projects
* Fixed the invalid submitter filter that appeared when clicking a submitter name in the datasets list
* Move success message popups so that they don't block the menu bar

## Other unreported bugs fixed:
* Hid the "Request Access" button that appeared when viewing a group/project as a logged-out user.
* Fixed CSV generation for datasets that had commas in their names.
* Added a change password button to My Account.
* Remove "Institution" filters. Instead of displaying "Institution", display "Group" instead. This also required a change to the aggregation fields on the GraphQL side for the dataset summary.
* Add missing `id` fields to queries, because InMemoryCache complains when a datatype has an ID in some queries but not others.


metaspace/webapp/src/api/annotation.ts
```diff
@@ -18,9 +18,8 @@ gql`query GetAnnotations($orderBy: AnnotationOrderBy, $sortingOrder: SortingOrde
         mz
         dataset {
           id
-          institution
           submitter { id name }
-          principalInvestigator { name }
+          principalInvestigator { id name }
```
intsco: What kind of ids should we have for `Person` type?

LachlanStuart:
TODO in this PR: Remove `authenticated` from Vuex to fix UploadPage behavior when user is not signed in

metaspace/webapp/src/api/annotation.ts
LachlanStuart: As discussed, I'll remove this field

PR #76, opened 2018-10-08
Invite not signed up user to group

intsco:
- Delete associated datasets on profile deletion
- Refactored user/credentials create/update. Users can sign up/log in with both local and google credentials
- All data mutation logging

### DIFF
--- a/metaspace/graphql/src/modules/auth/controller.ts
+++ b/metaspace/graphql/src/modules/auth/controller.ts
@@ -9,7 +9,7 @@ import {Connection} from 'typeorm';
 import config from '../../utils/config';
 import {User} from '../user/model';
 import {
-  createUserCredentials,
+  createUserCredentials, createGoogleUserCredentials,
   sendResetPasswordToken, resetPassword,
   verifyEmail, verifyPassword,
   initOperation,
@@ -163,7 +163,7 @@ const configureGoogleAuth = (router: IRouter<any>) => {
       },
       callbackify(async (accessToken: string, refreshToken: string, profile: any) => {
         return await findUserByGoogleId(profile.id)
-          || await createUserCredentials({
+          || await createGoogleUserCredentials({
             googleId: profile.id,
             name: profile.displayName,
             email: profile.emails[0].value,
--- a/metaspace/graphql/src/modules/auth/controller.ts
+++ b/metaspace/graphql/src/modules/auth/controller.ts
@@ -9,7 +9,7 @@ import {Connection} from 'typeorm';
 import config from '../../utils/config';
 import {User} from '../user/model';
 import {
-  createUserCredentials, createGoogleUserCredentials,
+  createUserCredentials,
   sendResetPasswordToken, resetPassword,
   verifyEmail, verifyPassword,
   initOperation,
@@ -163,7 +163,7 @@ const configureGoogleAuth = (router: IRouter<any>) => {
       },
       callbackify(async (accessToken: string, refreshToken: string, profile: any) => {
         return await findUserByGoogleId(profile.id)
-          || await createGoogleUserCredentials({
+          || await createUserCredentials({
             googleId: profile.id,
             name: profile.displayName,
             email: profile.emails[0].value,
--- a/metaspace/graphql/src/modules/auth/email.ts
+++ b/metaspace/graphql/src/modules/auth/email.ts
@@ -78,5 +78,5 @@ You have been invited to METASPACE by ${invitedBy}. Please sign up using this em
 Best wishes,
 METASPACE Team`;
   sendEmail(email, subject, text);
-  logger.info(`${invitedBy} invited user ${email}`);
+  logger.info(`Invitation email sent to ${email} by ${invitedBy}`);
 };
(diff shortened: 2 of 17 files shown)


metaspace/graphql/src/modules/auth/email.ts
```diff
@@ -34,19 +34,31 @@ export const sendVerificationEmail = (email: string, link: string) => {
...
+METASPACE Team`;
+  sendEmail(email, subject, text);
+};
+
+export const sendGoogleLoginEmail = (email: string) => {
+  const subject = 'METASPACE log in',
+    text =
+      `Dear METASPACE user,
+
+You are already signed up with our service. Please log in using your Google account.
```
LachlanStuart: Should we prevent email+password access if they've signed in with Google?

metaspace/graphql/src/modules/auth/controller.ts
```diff
@@ -163,7 +163,7 @@ const configureGoogleAuth = (router: IRouter<any>) => {
       },
       callbackify(async (accessToken: string, refreshToken: string, profile: any) => {
         return await findUserByGoogleId(profile.id)
-          || await createUserCredentials({
+          || await createGoogleUserCredentials({
```
LachlanStuart: We should also try searching by email address. We don't want to create 2 user accounts if they first sign up with username/password, then later use Google.

metaspace/graphql/src/modules/auth/operation.ts
```diff
@@ -113,25 +115,40 @@ const createLocalCredentials = async (userCred: UserCredentialsInput): Promise<C
   return cred;
 };
 
+export const createGoogleUserCredentials = async (userCred: UserCredentialsInput): Promise<void> => {
+  const existingUser = await findUserByGoogleId(userCred.googleId) as User;
+  if (existingUser) {
```
LachlanStuart: It's not actually possible for this branch to be taken - `createGoogleUserCredentials` is only called when there is no existing user with the same `googleId`

intsco: Good point!

metaspace/graphql/src/modules/auth/controller.ts
intsco: I thought about this situation. Do we want to show the user an error message if they are trying to use Google auth after email/password sign up?

metaspace/graphql/src/modules/auth/email.ts
intsco: Yes, I think we should show an error message in this case.

intsco:
PR is finished

metaspace/graphql/src/modules/group/controller.ts
```diff
@@ -266,10 +282,15 @@ export const Resolvers = {
...
 
-      const invUser = await findUserByEmail(connection, email);
-      if (!invUser)
-        // TODO: send sign up invitation
-        throw new UserError('Not Implemented Yet');
+      let invUser = await findUserByEmail(connection, email);
+      if (!invUser) {
+        // create not verified user
+        const userRepo = connection.getRepository(UserModel);
+        invUser = await userRepo.save({ notVerifiedEmail: email }) as UserModel;
```
LachlanStuart: This can create multiple records if run multiple times. `repo.save` relies on the primary key fields when deciding whether to `INSERT` or `UPDATE`. If no primary key fields are specified, it always `INSERT`s.

Also, should inactive users have credentials? I understand why they shouldn't logically have them, but it does simplify some of the other logic if you assume that the `credentials` relation doesn't need to be dynamically created.

metaspace/graphql/src/modules/user/model.ts
```diff
@@ -21,8 +21,8 @@ export class User {
   @Column({ type: 'text', default: 'user' })
   role: string;
 
-  @Column({ type: 'text', name: 'credentials_id' })
-  credentialsId: string;
+  @Column({ type: 'text', name: 'credentials_id', nullable: true })
+  credentialsId: string | null;
```
LachlanStuart: The below `credentials: Credentials` field should also be made nullable to match this.

metaspace/graphql/src/modules/auth/operation.ts
```diff
@@ -100,42 +91,84 @@ export const verifyPassword = async (password: string, hash: string|null|undefin
...
+    emailService.sendLoginEmail(existingUser.email!, link);
   }
   else {
-    const existingUserNotVerified = await findUserByEmail(userCred.email, 'not_verified_email');
-    if (!existingUserNotVerified) {
-      const newCred = userCred.googleId ?
-        await createGoogleCredentials(userCred) :
-        await createLocalCredentials(userCred);
-
+    if (userCred.googleId) {
```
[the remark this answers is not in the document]
intsco: I thought even inactive users will have `email`

metaspace/graphql/src/modules/group/controller.ts
intsco: True, I need to search `notVerifiedEmail` field too.
I assumed that inactive users don't have to have credentials.

PR #77, opened 2018-10-08
Projects

LachlanStuart:
It's reasonably stable and has enough functionality to demo. The main things missing are emails, and tests for the graphql queries.


metaspace/graphql/dsMutation.js
```diff
@@ -112,16 +114,26 @@ const isMemberOf = async (connection, user, groupId) => {
...
-    groupApproved,
+    groupApproved: groupId != null ? await isMemberOf(connection, submitterId, groupId) : false,
     piName: principalInvestigator ? principalInvestigator.name : undefined,
     piEmail: principalInvestigator ? principalInvestigator.email : undefined
   };
   await connection.getRepository(DatasetModel).save(dsUpdate);
+  if (projectIds != null && projectIds.length > 0) {
+    const userProjects = await connection.getRepository(UserProjectModel)
+      .find({ userId: submitterId, projectId: In(projectIds) });
+    const datasetProjectModel = connection.getRepository(DatasetProjectModel);
```
intsco: Why is `datasetProjectModel` name is better than 'datasetProjectRepo'?

LachlanStuart: That was a mistake. I'll change it to `datasetProjectRepo`.

metaspace/graphql/src/context.ts
```diff
@@ -1,22 +1,16 @@
...
-};
+export type UserProjectRoles = {[projectId: string]: ProjectRole | undefined}
 
 export interface Context {
-  connection: Connection,
-  user: JwtUser
+  // TODO: Replace connection with just EntityManager and rename it, as EntityManager is almost the same, but
+  // it can be swapped out by tests so that everything runs in a transaction.
+  connection: Connection | EntityManager;
+  user: JwtUser | null;
```
LachlanStuart: I'll do it in another PR after this is merged, but there's currently a lot of ambiguity on this field. `ctx.user` can have one of three forms: `null`, `{role: 'anonymous'}`, `{role: 'user/admin', id: '...', email: '...'}`. IMO, there's no value in having a `{role:'anonymous'}` user instead of just having a `null` user, and having a nullable `ctx.user.id` can cause easy-to-miss issues. I'd like to make anonymous users consistently have a `null` `ctx.user` field.

metaspace/graphql/src/modules/project/controller.ts
```diff
@@ -0,0 +1,310 @@
...
+
+    return true;
+  }
+};
+
+export const Resolvers = {
+  UserProject,
+  Project,
+  Query,
+  Mutation,
```
LachlanStuart: Due to the size of this file, I'm considering splitting it and putting each of these 4 groups of resolvers in their own file. I would do the same with the tests. Thoughts?

metaspace/graphql/src/modules/user/controller.ts
```diff
@@ -21,22 +24,20 @@ const assertCanEditUser = (user: JwtUser, userId: string) => {
...
+      scopeRole = SRO.PROFILE_OWNER;
     }
   }
   return scopeRole;
 };
 
 export const Resolvers = {
   User: {
-    async primaryGroup({scopeRole, ...user}: User & Scope, _: any,
+    async primaryGroup({scopeRole, ...user}: UserSource, _: any,
```
LachlanStuart: I've put in this `___Source` pattern so that we can easily check that the output of a parent resolver is compatible with a child resolver. I use "source" because it's the name of the first argument to resolvers. Not sure if there's any better name.

With this pattern we can customize the types a bit more easily, and won't have to use `LooselyCompatible` as often.

metaspace/graphql/tests/utils.test.js
```diff
@@ -1,139 +0,0 @@
-/**
```
LachlanStuart: The deleted tests were all failing and did not seem like they were still relevant, given that config processing has moved.

metaspace/graphql/datasetFilters.js
```diff
@@ -137,6 +137,7 @@ const datasetFilters = {
   submitter: new ExactMatchFilter('', { esField: 'ds_submitter_id' }),
   hasGroup: new NotNullFilter('', { esField: 'ds_group_id' }),
   group: new GroupMatchFilter('', { esField: 'ds_group_id' }),
+  project: new ExactMatchFilter('', { esField: 'ds_project_ids' }),
```
intsco: Will `term` query work on an array field?

metaspace/graphql/resolvers.js
```diff
@@ -357,7 +363,30 @@ const Resolvers = {
...
+      }
+
+      const userProjectRoles = await ctx.getCurrentUserProjectRoles();
+      let projectsQuery = ctx.connection.getRepository(ProjectModel)
+        .createQueryBuilder('project')
+        .innerJoin('project.datasetProjects', 'datasetProject')
+        .where(projectIsVisibleToCurrentUserWhereClause(ctx, userProjectRoles))
+        .andWhere('datasetProject.datasetId = :datasetId', {datasetId: ds._source.ds_id});
+      if (canSeeUnapprovedProjects) {
+        projectsQuery = projectsQuery.andWhere('datasetProject.approved')
```
intsco: If user can see unapproved projects, why to add this clause?

LachlanStuart: Oops, that should be `!canSeeUnapprovedProjects`

metaspace/graphql/src/modules/project/controller.ts
intsco: I don't have any ideas but I support it :)

metaspace/graphql/esConnector.js
```diff
@@ -107,17 +109,29 @@ const createBodyWithAuthFilters = (user) => {
...
       });
     }
+    const visibleProjectIds = Object.entries(userProjectRoles || [])
+                                    .filter(([id, role]) => [UPRO.MEMBER, UPRO.MANAGER].includes(role))
+                                    .map(([id, role]) => id);
+    if (visibleProjectIds.length > 0) {
+      filterObj.bool.should.push({
+        bool: {
+          must: [
+            { terms: { ds_project_id: visibleProjectIds } }
```
intsco: maybe ds_project_id**s**?

metaspace/graphql/src/modules/project/controller.ts
```diff
@@ -0,0 +1,313 @@
...
+    if (canViewProjectMembersAndDatasets(project.scopeRole)) {
+      return await ctx.connection
+        .getRepository(DatasetProjectModel)
+        .count({ where: { projectId: project.id, approved: true } });
+    } else {
+      return await ctx.connection
+        .getRepository(DatasetProjectModel)
+        .createQueryBuilder('dataset_project')
+        .innerJoinAndSelect('dataset_project.dataset', 'dataset')
+        .innerJoin('(SELECT id, is_public FROM "public"."dataset")', 'public_dataset', 'dataset.id = public_dataset.id')
```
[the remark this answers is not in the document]
LachlanStuart: Good point. Fixed.

metaspace/graphql/esConnector.js
LachlanStuart: Nice catch. I've pushed a fix.

PR #78, opened 2018-10-09
* Added a scale bar to annotations page

Ren22:
* Added vue-color that pops out a palette when user clicks on the scale bar
* Values are passed from the metadataJson

At the moment it looks like that: 
image
Values should be also properly updated if user scrolls the mouse wheel

_Minor problem that I discovered:_
If user saves the image the css styles are not properly handled
2018-10-01_19h10m13s_hmdb-v4_2018-03-08_c25h49no4_plus_h

Also the scale bar might appear in the Diagnostics tab
Going to fix this issues in the up-following commits

### DIFF
--- a/metaspace/webapp/src/components/ImageLoader.vue
+++ b/metaspace/webapp/src/components/ImageLoader.vue
@@ -165,20 +165,16 @@
    },
    computed: {
      scaleBarValX() {
-       if (this.pixelSizeIsActive) {
-         if (this.visibleImageWidth !== 0) {
-           return `${round((this.image.naturalWidth /
-             (this.zoom * this.visibleImageWidth)) * this.scaleBarSizeVal * this.pixelSizeX, 0)} µm`
-         }
+       if (this.pixelSizeIsActive && this.visibleImageWidth !== 0) {
+         return `${round((this.image.naturalWidth /
+           (this.zoom * this.visibleImageWidth)) * this.scaleBarSizeVal * this.pixelSizeX, 0)} µm`
        }
      },
 
      scaleBarValY() {
-       if (this.pixelSizeIsActive) {
-         if (this.visibleImageHeight !== 0) {
-           return `${round((this.image.naturalHeight /
-             (this.zoom * this.visibleImageHeight)) * this.scaleBarSizeVal * this.pixelSizeY, 0)} µm`
-         }
+       if (this.visibleImageHeight !== 0 && this.pixelSizeIsActive) {
+         return `${round((this.image.naturalHeight /
+           (this.zoom * this.visibleImageHeight)) * this.scaleBarSizeVal * this.pixelSizeY, 0)} µm`
        }
      },
 
--- a/metaspace/webapp/src/components/ImageLoader.vue
+++ b/metaspace/webapp/src/components/ImageLoader.vue
@@ -55,7 +55,6 @@
  import {quantile} from 'simple-statistics';
  import resize from 'vue-resize-directive';
  import config from '../clientConfig.json';
- import { Compact } from 'vue-color'
  import { round } from 'lodash-es';
  import Palette from './Palette.vue'
 
@@ -154,7 +153,6 @@
      }
    },
    components: {
-     'compact-picker': Compact,
      'palette': Palette
    },
    created() {
--- a/metaspace/webapp/src/components/ImageLoader.vue
+++ b/metaspace/webapp/src/components/ImageLoader.vue
@@ -22,10 +22,19 @@
              class="optical-image"
              :style="opticalImageStyle" />
       </div>
-      <div :style="cssProps" :class="{pixelSizeX: pixelSizeIsActive}" @click="onClickScaleBar()">{{scaleBarValX}}</div>
-      <div :style="cssProps" :class="{pixelSizeY: pixelSizeIsActive}" @click="onClickScaleBar()">{{scaleBarValY}}</div>
-      <compact-picker v-show="paletteIsVisible" :palette="palette"
-                      class="color-picker" @input="val=>updateColor(val)" v-model="scaleBarColor" />
+      <div :style="cssProps"
+           :class="{pixelSizeX: pixelSizeIsActive}"
+           title="Click to change the color"
+           @click="onClickScaleBar()">
+        {{scaleBarValX}}
+      </div>
+      <div :style="cssProps"
+           :class="{pixelSizeY: pixelSizeIsActive}"
+           title="Click to change the color"
+           @click="onClickScaleBar()">
+        {{scaleBarValY}}
+      </div>
+      <palette v-show="paletteIsVisible" class="color-picker" @colorInput="val=>updateColor(val)" />
     </div>
 
     <div ref="mapOverlap"
@@ -48,6 +57,7 @@
  import config from '../clientConfig.json';
  import { Compact } from 'vue-color'
  import { round } from 'lodash-es';
+ import Palette from './Palette.vue'
 
  const OPACITY_MAPPINGS = {
    'constant': (x) => 1,
@@ -138,14 +148,15 @@
        overlayDefault: true,
        overlayFadingIn: false,
        tmId: 0,
-       scaleBarColor: '#000',
-       scaleBarSizeVal: 25,
-       scaleBarShownSize: `${25}px`, // should be the same value as in scaleBarSizeVal
+       scaleBarColor: '#000000',
        paletteIsVisible: false,
-       scaleBarShadow: '#fff'
+       scaleBarShadow: '#FFFFFF'
      }
    },
-   components: { 'compact-picker': Compact },
+   components: {
+     'compact-picker': Compact,
+     'palette': Palette
+   },
    created() {
      this.onResize = throttle(this.onResize, 100);
    },
@@ -164,15 +175,27 @@
      this.isUnmounted = true;
    },
    computed: {
+     isIE() {
+       if (window.navigator.userAgent.indexOf('MSIE') > 0 ||
+       window.navigator.userAgent.indexOf('Trident/') > 0) {
+         return true
+       }
+       return false
+     },
+
+     scaleBarSizeVal() {
+       return 25
+     },
+
      scaleBarValX() {
-       if (this.pixelSizeIsActive && this.visibleImageWidth !== 0) {
+       if (this.pixelSizeIsActive && this.visibleImageWidth !== 0 && !this.isIE) {
          return `${round((this.image.naturalWidth /
            (this.zoom * this.visibleImageWidth)) * this.scaleBarSizeVal * this.pixelSizeX, 0)} µm`
        }
      },
 
      scaleBarValY() {
-       if (this.visibleImageHeight !== 0 && this.pixelSizeIsActive) {
+       if (this.pixelSizeIsActive && this.visibleImageHeight !== 0 && !this.isIE) {
          return `${round((this.image.naturalHeight /
            (this.zoom * this.visibleImageHeight)) * this.scaleBarSizeVal * this.pixelSizeY, 0)} µm`
        }
@@ -183,10 +206,6 @@
          return true
        }
        return false
-     },
-
-     palette() {
-       return ['#FFFFFF', '#999999', '#000000']
      },
 
      messageOS() {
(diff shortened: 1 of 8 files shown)


Ren22:
When user clicks on the scale bar:
image
Changed:
image

metaspace/webapp/src/modules/Annotations/AnnotationView.ts
```diff
@@ -190,6 +190,22 @@
...
+     JSON.parse(this.annotation.dataset.metadataJson).MS_Analysis.Pixel_Size !== undefined) {
+       return JSON.parse(this.annotation.dataset.metadataJson).MS_Analysis.Pixel_Size.Xaxis
+     }
+     return 0
+   }
+
+   get pixelSizeY() {
+     if (JSON.parse(this.annotation.dataset.metadataJson).MS_Analysis !== undefined &&
+     JSON.parse(this.annotation.dataset.metadataJson).MS_Analysis.Pixel_Size !== undefined) {
+       return JSON.parse(this.annotation.dataset.metadataJson).MS_Analysis.Pixel_Size.Yaxis
```
LachlanStuart: There's no need to parse `metadataJson` so many times. You could just use the `this.metadata` computed field.

metaspace/webapp/src/components/ImageLoader.vue
```diff
@@ -209,6 +256,43 @@
      }
    },
    methods: {
+     scaleBarValX() {
+       if (this.pixelSizeIsActive) {
+         const scaleBarXVal = 30;
+         if (this.$refs.visibleImage !== undefined) {
+           return `${round((this.image.naturalWidth /
+             (this.zoom * this.$refs.visibleImage.clientWidth)) * scaleBarXVal * this.pixelSizeX, 0)} µm`
```
LachlanStuart: `this.$refs.visibleImage.clientWidth` is 0 when the component first loads. Because it's a property on a DOM element, Vue doesn't turn it into a reactive property, so this never gets recalculated when the value changes. This leads to a divide-by-zero, so I see this when the component first loads: 

image

It looks like this problem affected other parts of this component, and so it already has a workaround in place: you can use `this.visibleImageWidth` and `this.visibleImageHeight` instead, as they're reactive properties that are manually updated whenever the `visibleImage` changes size.

Ren22:
So, it turns out the problem with css text style is Chrome related:
https://github.com/Sage-Bionetworks/Agora/issues/15#issuecomment-418536182

metaspace/webapp/src/components/ImageLoader.vue
Ren22: Thanks @LachlanStuart , there is also a wrong val for Yaxis as I discovered now which is because of copy-pasting the method scaleBarValX

Ren22: my workaround for this scenario was that I used methods and called the scaleBarValX() when the component got mounted and checked it later by `this.$refs.visibleImage !== undefined`. but if I use `this.visibleImageWidth` I can put them to `computed` I think and simply check that `this.visibleImageWidth !== 0`

metaspace/webapp/src/modules/Annotations/AnnotationView.ts
Ren22: ah, right, thanks

Ren22:
So, there is an unfortunate with Chrome, as I'm still puzzled how it renders css for text. Even though I tried these 2 packages 
`"dom-to-image": "https://github.com/MichalBryxi/dom-to-image/archive/v2.6.1.tar.gz",
    "dom-to-image-google-font-issue": "^2.6.2"`
it didn't work for the latest Chrome version.

Image on website:
image

Saved image:
image

Besides, Chrome doesn't render `position: absolute` properly(above),comparing to FF(below):

image

image

So I had to adjust css values to look it more or less nice on both Chrome and FF

metaspace/webapp/src/components/ImageLoader.vue
```diff
@@ -123,9 +137,15 @@
        dragThrottled: false,
        overlayDefault: true,
        overlayFadingIn: false,
-       tmId: 0
+       tmId: 0,
+       scaleBarColor: '#000',
+       scaleBarSizeVal: 25,
+       scaleBarShownSize: `${25}px`, // should be the same value as in scaleBarSizeVal
```
intsco: Why not use a variable instead?

metaspace/webapp/src/components/ImageLoader.vue
```diff
@@ -136,13 +156,43 @@
...
    },
    beforeDestroy() {
      window.removeEventListener('resize', this.onResize);
      this.onResize.cancel(); // If there's a pending throttled call, cancel it
      this.isUnmounted = true;
    },
    computed: {
+     scaleBarValX() {
+       if (this.pixelSizeIsActive) {
+         if (this.visibleImageWidth !== 0) {
```
intsco: The above two ifs can potentially be combined

metaspace/webapp/src/components/ImageLoader.vue
Ren22: It will be undefined if use just `scaleBarShownSize: ${this.scaleBarSizeVal}px`, this is a bit trickier with these data values on component start

Ren22: that's true, thanks

metaspace/webapp/src/components/ImageLoader.vue
Ren22: that's true, thanks

metaspace/webapp/src/modules/Annotations/AnnotationView.ts
```diff
@@ -190,6 +190,22 @@
      return Object.assign(safeJsonParse(this.annotation.dataset.metadataJson), datasetMetadataExternals);
    }
 
+   get pixelSizeX() {
+     if (this.metadata.MS_Analysis !== undefined &&
+       this.metadata.MS_Analysis.Pixel_Size !== undefined) {
```
intsco: `!= null` would be preferable here as it checks for both `null` and `undefined`, it's also more concise

metaspace/webapp/src/components/ImageLoader.vue
intsco: I thought more about a local variable not a field

metaspace/webapp/src/modules/Annotations/AnnotationView.ts
intsco: AFAIK, using long version `!==`  is preferable for comparing variables with types (e.g. strings). Here you compare to `undefined` anyway.
BTW, inequality (!=) operator tries to convert types before comparing values while !== doesn't.
Anyway, it's just general recommendation, in this particular case either is ok.

LachlanStuart: > I think checking only for != would mean value check but not the type if I'm not mistaken, 

You may be confusing JS's `==` and `===` for the "Value Equality" and "Reference Equality" checks that other languages have. JS doesn't actually have that. It just has a [collection of arbitrary rules for `==`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness#Loose_equality_using). `===` also has a couple weird rules (`NaN !== NaN`, `0 === -0`), but otherwise strictly checks value equality for primitives and reference equality for non-primitives.

My rule of thumb is to always use `===`/`!==` in boolean conditions except for these three cases:
* Checking whether something is `null` or `undefined` - prefer `== null` / `!= null` unless you specifically care about `null` vs `undefined`.
* Checking a string for `null`/`undefined`/`''`, I use `if (!str)` and either `if (str)` or `if (!!str)` depending on how explicit I feel it should be based on the situation.
* Null chaining. If you don't know when parts of a property access path are going to be null, `a && a.b && a.b.c` is acceptable. I'm eagerly awaiting a [TC39 proposal](https://github.com/TC39/proposal-optional-chaining#base-case) to add better syntax for this to ES6.

metaspace/webapp/src/components/ImageLoader.vue
```diff
@@ -191,7 +237,17 @@
...
+     },
+
+     cssProps() {
+       console.log(this.scaleBarShownSize)
+       return {
+         '--scaleBar-color': this.scaleBarColor,
+         '--scaleBarX-size': this.scaleBarShownSize,
+         '--scaleBarY-size': this.scaleBarShownSize,
+         '--scaleBarShadow-color': this.scaleBarShadow
+       }
```
LachlanStuart: [IE11 doesn't support CSS vars](https://developer.mozilla.org/en-US/docs/Web/CSS/var#Browser_compatibility) unfortunately. We should hide it for IE11, otherwise it looks bad:
image

Using the same logic as what we use to disable dom-to-image for IE would be fine

metaspace/webapp/src/components/ImageLoader.vue
```diff
@@ -191,7 +237,17 @@
 
      opticalImageUrl() {
        return (config.imageStorage || '') + this.opticalSrc;
-     }
+     },
+
+     cssProps() {
+       console.log(this.scaleBarShownSize)
```
LachlanStuart: Please clean up `console.log`s when you're done with them

metaspace/webapp/src/modules/Annotations/AnnotationView.ts
Ren22: hmm, then it's particularly confusing in this case. What I meant was exactly `undefined` type checking (not reference such as `is` in python). Following from
https://stackoverflow.com/questions/5076944/what-is-the-difference-between-null-and-undefined-in-javascript
`undefined is a type itself (undefined) while null is an object.` 
The potential problem I see here is comparing to empty string ,as this field `this.metadata.MS_Analysis` can be theoretically empty. in that case `this.metadata.MS_Analysis != null` is truthy, which shouldn't be

metaspace/webapp/src/components/ImageLoader.vue
LachlanStuart: It's always best to have as few `data` members as possible in a component, because the simpler the basic state of the component is, the easier it is to reason about changes to that state. Since `scaleBarShownSize` is supposed to be derived from `scaleBarSizeVal`, using a `computed` would be a more natural fit. But in this case, since it's a tiny amount of code, I'd prefer it if `cssProps` just had:
```
         '--scaleBarX-size': `${this.scaleBarSizeVal}px`,
         '--scaleBarY-size': `${this.scaleBarSizeVal}px`, 
```

You shouldn't need to protect against `undefined` here, since `cssProps` won't get used until the component is rendered, which happens after `data()` is applied to the component.

metaspace/webapp/src/modules/Annotations/AnnotationView.ts
LachlanStuart: >  undefined is a type itself (undefined) while null is an object.

You should not read too deeply into this - they're specifically talking about the implementation quirks of JS, not how those values should actually be used. `typeof null === 'object'` is only for [backwards compatibility](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof#Additional_information) because of a bug in the initial implementation of JS. I imagine that `undefined` not being a keyword while `null` is a keyword has a similar history.

The reason I advocate using `!= null` instead of `!== undefined` or `!== null` is because 90% of the time you should not care about the difference between the two. JS libraries sometimes use them inconsistently to mean different things, but in most code you only care if a value is there or not. If the difference between `null` and `undefined` matters, it needs to be made very obvious to the reader, because they need to know which to use. Using `!= null` by default is a good way of indicating to the reader that this code does not have any special behavior distinguishing the two, so they don't need to care which one gets used.

Ren22:
@intsco @LachlanStuart this PR was updated(see the latest commit)

metaspace/webapp/src/components/ImageLoader.vue
```diff
@@ -123,9 +147,16 @@
...
        overlayFadingIn: false,
-       tmId: 0
+       tmId: 0,
+       scaleBarColor: '#000000',
+       paletteIsVisible: false,
+       scaleBarShadow: '#FFFFFF'
      }
    },
+   components: {
+     'compact-picker': Compact,
```
LachlanStuart: This line and the `vue-color` import need to be removed, or else webpack fails to build as `vue-color` no longer is installed.

metaspace/webapp/yarn.lock
```diff
@@ -8365,6 +8365,10 @@ lodash.templatesettings@^4.0.0:
   dependencies:
     lodash._reinterpolate "~3.0.0"
 
+lodash.throttle@^4.0.0:
```
LachlanStuart: The yarn.lock changes need to be reverted to get the CircleCI tests to pass

Ren22:
@LachlanStuart thank you, it seemed not too much work to do, so I quickly introduced one more component :) Thanks for the tip of how to analyze bundles, I forgot about it, it's quite useful.

PR #85, opened 2018-10-24
Centralize Python dependencies in setup.py

LachlanStuart:
* Moves all dependencies out of the conda environment definitions into `setup.py`
* Removes the `pip uninstall` step. I'm not sure why it was needed, but pip seems to uninstall & reinstall `sm` (but not its unchanged dependencies) in the `pip install` step
* Adds `-e` to the `pip install` step so that `sm` is installed in develop mode, so that when modules are imported from `sm`, it uses the code in `/opt/dev/metaspace/metaspace/engine/sm/` instead of the copy that is made in `/opt/dev/miniconda3/envs/sm/lib/python3.6/site-packages/sm/` at the time of installation.
* Fixes pip package installation in docker


intsco:
All looks good to me. We used to prefer installing binary packages like numpy and scipy using conda because pip didn't support it before. But now I cannot find any reason not to use pip inside conda environment for that.
We should test this ways of installing dependencies on staging before though.

intsco:
One more caveat using dependencies in setup.py, the engine is installed automatically onto cluster machines when a dataset is submitted. If dependencies are not fixed, a new version of a library can be installed that is not compatible with engine code.

LachlanStuart:
Per discussion: on Monday we'll test deploying it to staging. This branch won't be merged until then.

We should make sure to test processing speed before & after. For [some tasks](https://www.anaconda.com/blog/developer-blog/tensorflow-in-anaconda/), Anaconda installs faster libraries. We currently specifically exclude MKL in favor of OpenBLAS because OpenBLAS is much smaller but apparently just as fast, it's worth checking.

PR #86, opened 2018-10-25
Nginx: serving https

intsco:
* forward all http traffic to https
* copy ssl cert and key to nginx config dir
* the cert and key encrypted using ansible-vault


ansible/aws/group_vars_all.yml.template
```diff
@@ -286,10 +293,6 @@ nginx_servers:
           - proxy_set_header Upgrade $http_upgrade;
           - proxy_set_header Connection "upgrade";
           - include proxy-params.conf;
-      - path: /mol_db/
-        params:
-          - proxy_pass http://moldb_server/;
-          - include proxy-params.conf;
```
LachlanStuart: I think `/graphiql` is also no longer needed.

ansible/roles/nginx/tasks/main.yml
```diff
@@ -36,6 +36,15 @@
...
   with_items: "{{ basic_auth_users }}"
 
+- name: Copy ssl certificate and key
+  become: yes
+  copy:
+    src={{ item }} dest=/etc/nginx/ssl/{{ item }}
+    owner=ubuntu group=ubuntu mode=0644
+  with_items:
+    - metaspace2020.eu-dfn-ca-chain.pem
+    - metaspace2020.eu.nopw.key
```
LachlanStuart: Does anything need to be added to decrypt these files after they're being copied?

PR #97, opened 2018-11-02
Admin tools

LachlanStuart:
* Adds `/admin/groups` page to see a full list of groups
* Allows admins to freely change group member roles
* Allows group admins/project managers to shift other members between Member and Group Admin/Manager. I intentionally don't allow changes to/from Pending and Invited to force managers to use the intended sign up flow. I intentionally don't allow changes to the current user to prevent groups becoming broken with no admin/manager.

### DIFF
--- a/metaspace/graphql/schemas/group.graphql
+++ b/metaspace/graphql/schemas/group.graphql
@@ -20,6 +20,10 @@ input UpdateGroupInput {
   urlSlug: String
 }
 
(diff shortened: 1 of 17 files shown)


intsco:
@Ren22 please review so that @LachlanStuart can merge it. So it can be deployed and tested

Ren22:
@intsco I'm reviewing it

PR #98, opened 2018-11-05
Python client refactoring

intsco:
* python-client sm_annotation_utils refactoring
* verify_certificate option
* check for response in GraphQLClient.(__init__ / get_jwt)
* removed not supported field: institution
* updateDatabasesQuery -> update_dataset


Ren22:
I suspect this is because the path `https://staging.metaspace2020.eu/api_auth/signin` doesn't work, which is basically used to login with email/password

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -562,8 +583,12 @@ def fetch_image(url):
 
 
 class SMInstance(object):
-    def __init__(self, config=DEFAULT_CONFIG):
-        self._config = config
+    def __init__(self, host='http://metaspace2020.eu', verify_certificate=True):
```
Ren22: perhaps it's better to set it to `False`?

LachlanStuart: Sorry, missed this in my first review. I think this should be HTTPS

PR #103, opened 2018-11-08
Allow admins to see users' email addresses in metadata

LachlanStuart:


metaspace/graphql/resolvers.js
```diff
@@ -172,7 +172,11 @@ const Resolvers = {
     rhoChaos: (hit) => hit._source.chaos,
 
     dataset(hit) {
-      return Object.assign({_id: hit._source.ds_id}, hit);
+
+      return {
+        _id: hit._source.ds_id,
+        _source: hit._source,
+      };
```
LachlanStuart: Unrelated, but this code was misleading - `hit` contains an `_id` field with the ID of the annotation, so it was always overwriting the dataset's ID with the annotation ID. Fortunately nothing actually used this `_id` field - the `_source.ds_id` field is always used for datasets, but I tidied this up anyway to save future headaches.

metaspace/webapp/src/components/DatasetInfo.vue
```diff
@@ -55,11 +55,16 @@
 
       treeData() {
         let schemaBasedVals = this.objToTreeNode(null, this.metadata, this.schema);
+        const submitterEmail = this.dsSubmitter.email ? ` (${this.dsSubmitter.email})` : '';
```
Ren22: why is this line in parentheses?
`(${this.dsSubmitter.email})`

PR #105, opened 2018-11-08
Post release fixes

intsco:
* `https` for the default host in python client
* `force` argument in sm api dataset update to re-index the dataset without any updates
* Increased `update_by_query` timeout in es_export to 60s


intsco:
Good catch. I'd better merge it into the release branch then.

PR #107, opened 2018-11-09
Fixing "Send request" button functionality on find a group

Ren22:


metaspace/webapp/src/modules/MetadataEditor/sections/FindGroupDialog.vue
```diff
@@ -31,7 +31,7 @@
       <el-button @click="handleClose">Cancel</el-button>
       <el-button
         type="primary"
-        :disabled="groupId == null"
+        :disabled="groupId == null || sameGroupRequest"
```
LachlanStuart: This should also show an error against the form item so that the user can tell why the button is disabled. I can't think of a one-size-fits-all error message, so I think you'd need several different error messages depending on their role.

metaspace/webapp/src/modules/MetadataEditor/sections/FindGroupDialog.vue
```diff
@@ -47,17 +47,22 @@
...
   } from '../../../api/dataManagement';
+  import { userProfileQuery, UserProfileQuery } from '../../../api/user'
   import reportError from "../../../lib/reportError";
   import './FormSection.scss';
 
   @Component<FindGroupDialog>({
     apollo: {
+      currentUser: {
+        query: userProfileQuery,
+        fetchPolicy: 'cache-first'
```
LachlanStuart: Please put `await this.$apollo.queries.currentUser.refetch()` in the code after the `requestAccessToGroupMutation` so that the cache is updated with the new list of the user's groups

Ren22:
@LachlanStuart updated the PR, I couldn't change Watcher to get/set prop correctly in the current frame of time, so I just left the watcher for now.

LachlanStuart:
@Ren22 I've changed the error message to be dependent on the user's existing role, as I don't feel that there was any natural sounding error message that covered all possible roles. I've also refactored the entire check into the getter because it was required to determine the user's existing role.

PR #110, opened 2018-11-12
Fixes

LachlanStuart:
Lots of small fixes. See individual commits.

### DIFF
--- a/metaspace/engine/scripts/update_es_index.py
+++ b/metaspace/engine/scripts/update_es_index.py
@@ -35,7 +35,6 @@ def _reindex_datasets(ds_ids, db, es_exp):
     logger.info('Reindexing %s dataset(s)', len(ds_ids))
     for idx, ds_id in enumerate(ds_ids):
         logger.info(f'Reindexing {idx+1} out of {len(ds_ids)}')
-        ds_name = None
         try:
             # Delete from ES regardless of whether the DS exists, so that this can clean up deleted datasets
             es_exp.delete_ds(ds_id)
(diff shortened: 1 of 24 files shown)


metaspace/graphql/package.json
```diff
@@ -67,7 +67,7 @@
     "typeorm": "0.2.7",
     "typescript": "^2.9.2",
     "uuid": "^3.3.2",
-    "winston": "^2.3.1"
+    "winston": "^2.4.4"
```
LachlanStuart: We were previously on 2.4.3. 2.4.4 adds TypeScript definitions

metaspace/webapp/src/lib/formatCsvRow.ts
```diff
@@ -1,7 +1,7 @@
 export default (values: string[]): string => {
   const escaped = values.map(v => {
     if (v != null) {
-      return `"${String(v).replace('"', '""')}"`;
```
LachlanStuart: `String.prototype.replace(string, string)` only replaces the first instance. Using a regex with the `g` option makes it replace every instance.

metaspace/engine/scripts/update_es_index.py
```diff
@@ -35,21 +35,24 @@ def _reindex_datasets(ds_ids, db, es_exp):
...
             # Delete from ES regardless of whether the DS exists, so that this can clean up deleted datasets
             es_exp.delete_ds(ds_id)
             ds = db.select_one("select name, config from dataset where id = %s", (ds_id,))
             if ds:
                 ds_name, ds_config = ds
                 for mol_db_name in ds_config['databases']:
-                    mol_db = MolecularDB(name=mol_db_name, iso_gen_config=ds_config['isotope_generation'])
-                    isocalc = IsocalcWrapper(ds_config['isotope_generation'])
-                    es_exp.index_ds(ds_id, mol_db=mol_db, isocalc=isocalc)
+                    try:
```
intsco: How do we find these failed dataset-moldb pairs afterwards? The dataset will have mol databases assigned to it with no results.

LachlanStuart: This is already a problem, because the datasets will often be indexed successfully on some databases and fail after there is already data in ElasticSearch. The only way we can currently find those datasets is by reading the logs. 

This change at least makes a missing or misconfigured mol_db (e.g. HMDB-v3.6) not break any legitimate annotations in the other mol_dbs.

I'm not sure whether there's anything we can do other than adding slack notifications of failures or something like that.

intsco: True, the engine takes care of dataset status update if indexing of a moldb failed. But here we don't handle it.

PR #111, opened 2018-11-12
More fixes

LachlanStuart:
This fixes two security issues:
* Importing datasets into a group or project didn't check who owned the dataset, which could allow access to private datasets
* Activating an inactive account using Google didn't remove any previously set password. This was a problem because setting a password on an inactive account can be done without any sort of verification by just creating an account with the same email address.

It also fixes a couple bugs:
* `"features"` section is missing from the docker `clientConfig.json`
* Signing up using email/password to activate an inactive account wouldn't save the password. This is the problem Veronica saw earlier today.

### DIFF
--- a/docker/sm-webapp/config/clientConfig.json
+++ b/docker/sm-webapp/config/clientConfig.json
@@ -15,6 +15,9 @@
     },
     "storage": "/opt/data/uploads"
   },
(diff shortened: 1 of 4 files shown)


intsco:
@LachlanStuart do you mind if I add it to the project https://github.com/metaspace2020/metaspace/projects/1?

LachlanStuart:
@intsco Go for it.

PR #115, opened 2018-11-14
Optimize dataset updates

LachlanStuart:
Review with white space comparison turned off, as most of the lines changed in es_export.py are only indentation changes.

Some compromises were needed. Discussion of alternatives is welcome:
* I had to [pick a different server backend](https://bottlepy.org/docs/dev/deployment.html#switching-the-server-backend) for bottle, because its default is single-threaded and can't dispatch messages fast enough to keep the update daemon workers working. I chose CherryPy because it is actively developed and doesn't rely on green threads or forks. CherryPy seems to default to 10 threads.
* I noticed that I now get a lot of these warnings about the Elasticsearch HTTP connections not being cleaned up:
    ```
    sm-api_1              | namedtuple_Column:14: ResourceWarning: unclosed <socket.socket fd=117, family=AddressFamily.AF_INET, type=2049, proto=6, laddr=('172.20.0.11', 34468), raddr=('172.20.0.3', 9200)>
    ```
    I compared against master - apparently I've always been getting warnings like this, but with CherryPy's multithreading they now come in large batches, all at once.
* Each `DatasetLocker` uses its own database connection. I found that psycopg2 can only apply transactions at the connection level, so I couldn't implement this using transactions. Also, Postgres 9.5 doesn't support the transaction-based advisory locks, so I was stuck using session-based advisory locks. Postgres defaults to a maximum of 100 connections, and I managed to hit that limit when I tried running with 32 worker threads. 
* Postgres' advisory locks don't support strings, only 32-bit and 64-bit integers. I've used a 32-bit hash of the dataset ID instead of locking on the exact ID itself.
* Using a single ElasticSearch pipeline name didn't scale to multiple threads, so now it uses a different name for each dataset and deletes the pipeline after use
* Updating a dataset doesn't change its status at all now. This means no status update is published, so the webapp won't automatically refresh the list after a name change. 

Locally, with these changes and 4 update daemon worker threads, it takes 160s to update the group name on 114 datasets. None of the Python processes use more than 1% CPU, and ElasticSearch uses 300-400%. I tried 8 and 16 threads and they took 120s but on 16 threads there was 1 request timeout failure and lagged everything else on my computer. I don't think the extra speed from 8 threads is worth the risk of lagging everything else on the system.

I also tried to fix the scientific test on CircleCI for extra certainty that I hadn't broken anything, but I couldn't get past this error: https://circleci.com/gh/metaspace2020/metaspace/1727

### DIFF
--- a/metaspace/engine/sm/engine/es_export.py
+++ b/metaspace/engine/sm/engine/es_export.py
@@ -5,6 +5,7 @@ import logging
 from collections import defaultdict, MutableMapping
 import pandas as pd
 
+from sm.engine.dataset_locker import DatasetLocker
 from sm.engine.util import SMConfig
 
 logger = logging.getLogger('engine')
@@ -215,6 +216,7 @@ class ESExporter(object):
         self._es = init_es_conn(es_config)
         self._ingest = IngestClient(self._es)
         self._db = db
+        self._ds_locker = DatasetLocker(self.sm_config['db'])
         self.index = es_config['index']
         self._get_mol_by_sf_dict_cache = dict()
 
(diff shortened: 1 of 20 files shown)


metaspace/engine/sm/engine/es_export.py
```diff
@@ -262,100 +265,103 @@ def _add_ds_fields_to_ann(self, ann_doc, ds_doc):
...
+                # Note: mind the time gap between the queries when called from different processes
+                self._es.update_by_query(
+                    index=self.index,
+                    body={'query': {'term': {'ds_id': ds_id}}},
+                    params={
+                        'pipeline': pipeline_id,
+                        'wait_for_completion': True,
+                        'request_timeout': 60,
+                    })
+                self._ingest.delete_pipeline(pipeline_id)
```
intsco: If an exception is thrown, the pipeline won't be deleted

intsco:
To fix https://circleci.com/gh/metaspace2020/metaspace/1727,
one can try to limit the number of spark executors by setting `spark.master = "local[2]",`

LachlanStuart:
The scientific test passed with the `local[2]` change: https://circleci.com/gh/metaspace2020/metaspace/1755

PR #117, opened 2018-11-15
Add indicators for group/project invitations/requests

LachlanStuart:
When you are invited to a group/project, or you manage a group/project and there is a pending request, this add a trail of indicators, some with tooltips, to guide you to where you can accept:

#### Menu bar:
image
image
#### My Account
image
#### Group Page
image

It also sorts the group/project members list so that pending requests come first, then invited people, then admins/managers, and finally members.

I had to add some extra code to `setupTestFramework`, because I found that after tests ended the components they create would continue to exist and would sometimes send graphql queries during other tests. This is why there are some seemingly unrelated HTML class changes in the snapshots.

### DIFF
--- a/metaspace/graphql/src/modules/group/util/resolveGroupScopeRole.ts
+++ b/metaspace/graphql/src/modules/group/util/resolveGroupScopeRole.ts
@@ -1,0 +1,21 @@
+import {Context} from '../../../context';
+import {ScopeRole, ScopeRoleOptions} from '../../../bindingTypes';
+import {UserGroup as UserGroupModel, UserGroupRoleOptions} from '../model';
+
+export const resolveGroupScopeRole = async (ctx: Context, groupId?: string): Promise<ScopeRole> => {
+  let scopeRole = ScopeRoleOptions.OTHER;
+  if (ctx.user != null && groupId) {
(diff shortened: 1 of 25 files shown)


metaspace/graphql/src/modules/group/util/resolveGroupScopeRole.ts
```diff
@@ -0,0 +1,21 @@
+import {Context} from '../../../context';
+import {ScopeRole, ScopeRoleOptions} from '../../../bindingTypes';
+import {UserGroup as UserGroupModel, UserGroupRoleOptions} from '../model';
+
+export const resolveGroupScopeRole = async (ctx: Context, groupId?: string): Promise<ScopeRole> => {
+  let scopeRole = ScopeRoleOptions.OTHER;
+  if (ctx.user != null && groupId) {
+    const userGroup = await ctx.connection.getRepository(UserGroupModel).findOne({
+      where: { userId: ctx.user.id, groupId },
```
LachlanStuart: Do comments work for me?

LachlanStuart:
@Ren22 It's [ES6 object literal property shorthand](https://ariya.io/2013/02/es6-and-object-literal-property-value-shorthand). `{ userId: ctx.user.id, groupId }` and `{ userId: ctx.user.id, groupId: groupId }` are identical.

PR #126, opened 2018-11-16
Add toggle to projects page to switch between All Projects and My Projects

### LINKED ISSUE #121: Add toggle to projects page to switch between All Projects and My Projects

LachlanStuart:
Fixes #121 

image

The dataflow is a bit awkward because "My Projects" is queried and paginated client-side, and "All Projects" is queried & paginated server-side. It's not worth the time to adapt the graphql API to clean this up.

Other changes:
* Sorts projects by `num_members * 20 + num_datasets` so that more prominent projects rise to the top.
* Fixes a bug in pagination on the projects page. Basically, it didn't work at all before. It seems like TypeORM ignores the QueryBuilder methods `skip` and `take` for Postgres, so I changed to `offset` and `limit`.
* Moves the "Create Project" button into the same row as the search box.

### DIFF
--- a/metaspace/webapp/src/modules/Project/ProjectsListPage.spec.ts
+++ b/metaspace/webapp/src/modules/Project/ProjectsListPage.spec.ts
@@ -40,6 +40,7 @@ describe('ProjectsListPage', () => {
     });
     const wrapper = mount(ProjectsListPage, { router, provide, store, sync: false });
     await Vue.nextTick();
+    await Vue.nextTick();
 
     expect(wrapper).toMatchSnapshot();
     const projectIds = wrapper.findAll({name:'ProjectsListItem'}).wrappers.map(item => item.props().project.id);
--- a/metaspace/webapp/src/modules/Project/ProjectsListPage.spec.ts
+++ b/metaspace/webapp/src/modules/Project/ProjectsListPage.spec.ts
@@ -1,4 +1,4 @@
-import { mount } from '@vue/test-utils';
+import {mount, Wrapper} from '@vue/test-utils';
 import Vue from 'vue';
 import ProjectsListPage from './ProjectsListPage.vue';
 import router from '../../router';
@@ -22,10 +22,10 @@ describe('ProjectsListPage', () => {
   const mockProject3: ProjectsListProject = { id: 'project 3', name: 'project three', urlSlug: null, isPublic: true, currentUserRole: 'MEMBER',
     numMembers: 10, numDatasets: 5, createdDT: '2018-04-30T11:00:00.000', latestUploadDT: '2018-05-15T13:00:00.000' };
 
-  const mockMyProjects: MyProjectsListQuery['myProjects'] = {
+  const makeMockMyProjects = (projects: any[]): MyProjectsListQuery['myProjects'] => ({
     id: 'id',
-    projects: [{project: mockProject1}, {project:mockProject3}],
-  };
+    projects: projects.map(project => ({project})),
+  });
   const mockAllProjects: ProjectsListQuery['allProjects'] = [mockProject1, mockProject2];
 
   sync(store, router);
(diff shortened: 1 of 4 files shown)


metaspace/webapp/src/modules/Project/ProjectsListPage.vue
```diff
@@ -102,22 +118,30 @@
     get query(): string {
       return this.$store.getters.filter.simpleQuery || '';
     }
+    get filteredMyProjects() {
+      if (this.query && this.myProjects != null) {
+        return this.myProjects.filter(p => p.name.toLowerCase().includes(this.query.toLowerCase()));
```
Ren22: Will it also handle the situation when multiple spaces by mistake might occur? say `Project(space)(space)X`

metaspace/webapp/src/modules/Project/ProjectsListPage.spec.ts
```diff
@@ -40,6 +40,7 @@ describe('ProjectsListPage', () => {
     });
     const wrapper = mount(ProjectsListPage, { router, provide, store, sync: false });
     await Vue.nextTick();
+    await Vue.nextTick();
```
Ren22: why this one more Tick is needed here?

metaspace/webapp/src/modules/Project/ProjectsListPage.vue
```diff
@@ -146,6 +147,13 @@
       this.page = 1;
     }
 
+    @Watch('currentUser')
```
Ren22: Just to understand: we use watchers for the data that comes with graphQL queries and for the rest we can use computed properties - is it correct?

LachlanStuart: No. We use computed properties for everything we possibly can, and watches only where computed properties aren't possible. The two main cases where computed properties aren't possible are: reacting to events in a way that does a one-way data change (e.g. user logs out, so reset the form), and updating data outside of the component (e.g. user updates a field, so tell ElementUI to hide the validation message).

In this case the behavior reacts to an event in a way that's not directly calculable from the existing properties. If I did something like `get visibleTab() { return this.currentUser != null ? this.tab : this.ALL_PROJECTS; }`, there would be an unwanted side effect of it remembering your previous filter if you sign out and sign in again.

metaspace/webapp/src/modules/Project/ProjectsListPage.spec.ts
LachlanStuart: It wasn't clear and I didn't put the time in to investigating what caused this particular case.

The basic problem is that async functions, timeouts, event handlers, etc. all get run in queues. Async functions and Vue component updates in the "microtask" queue and timeouts, event handlers, and `Vue.nextTick()` in the "macrotask" queue. 

The JavaScript engine always completes everything in the microtask queue before starting a new macrotask. This means that usually if an async function does multiple `await`s`, but those promises are already resolved, then all of those awaits will be run before `Vue.nextTick()` gets to run. However, if a test has run `Vue.nextTick()`, and then the component does a `setTimeout` or waits for an event, then that `setTimeout`/event handler is put into the macrotask queue *after* the test's `Vue.nextTick()`. 

The simplest fix is to just add another `await Vue.nextTick()`, which is why it worked in this case. But in some cases the component being tested will run a variable number of macrotasks, which would mean you'd need many `await Vue.nextTick()`s. Obviously not a good solution. I've encountered that once, and my solution was to add an extra field to the component containing a promise that resolved after all the macrotasks had finished.

metaspace/webapp/src/modules/Project/ProjectsListPage.vue
LachlanStuart: No. I don't expect people to use this for searching for multiple words...

LachlanStuart:
As this is already reviewed, I'm going to merge this and implement the quick-filter dropdown we discussed in a separate PR.

metaspace/webapp/src/modules/Project/ProjectsListPage.spec.ts
Ren22: @LachlanStuart Thank you for the explanation, it now makes better sense

PR #127, opened 2018-11-16
Bring back apollo client-side cache

LachlanStuart:
Originally we disabled the client-side cache because inaccurate data (usually from datasets from ElasticSearch) would sometimes overwrite accurate data (from Postgres) in the cache, causing e.g. the primary group in the header to change or disappear, or the user's email to appear or disappear depending on whether the last query about the user came from `dataset.submitter`, `group.members.user` or `currentUser`.

There were a couple issues with having the cache disabled:
* Vue-apollo inexplicably disables the loading status of the queries, so none of our loading indicators were working
* Updating the names of users/projects/groups wouldn't cause those names to be updated elsewhere on the page (e.g. in the menu bar, or the group/project page header)

The solution I chose was to change the name of the GraphQL type for data that was returned from ElasticSearch. e.g. `dataset.submitter` is now `DatasetUser` instead of `User`, `dataset.group` is now `DatasetGroup` instead of `Group`. This means that they get cached separately to the `user`/`currentUser`/`group`-type queries. 

There is at least one place where data still potentially conflicts, but I don't think it's an issue as only 1 view of the data appears at a time and it always refreshes from the server:
* On a group/project page where you are not a member, email addresses of managers will be `null`, but if you view a group/project where you are a member, you'll get the same data from the server with the email addresses populated.

I had to restructure one query used in MetadataEditor, for finding the submitter's groups & projects.

There were several other viable solutions that I chose not to implement:
* Renaming the `__typename` field in queries will cause Apollo to not find it, and not cache the data by type. I chose not to do this because it would need to be applied to all dataset queries, and if we ever forgot to apply it we'd get bugs again.
* Submitting a patch for the loading issue to vue-apollo. I chose not to do this because upgrading vue-apollo has been problematic in the past. We're intentionally running `v3.0.0-beta.13` because last time I upgraded, versions `v3.0.0-beta.14` to `v3.0.0-beta.19` had insurmountable issues.


LachlanStuart:
Python-client was querying for some of the fields that I had removed, but not using them. To prevent API breakages I added those fields back in, but deprecated them, and made them always null.

I also removed one notebook that contained user data. It was completely outdated so I removed it instead of fixing it.

PR #138, opened 2018-11-19
Tidy UX on My Account page

LachlanStuart:
image

* Replaced the "Change password" button with a (hopefully) clearer link button
* Disabled the "Save" button until the name/email has changed
* Made the Primary Group drop-down auto-save, because it's so far away from the save button that it wasn't clear the two were connected
* (unrelated) Fixed the `Dataset.opticalImage` and `Dataset.rawOpticalImageUrl` resolvers, and changed code to not use `Dataset.opticalImage` because it's deprecated. The issue was that `await rawOpticalImage(ds._source.ds_id, ctx).url` is the same as `await (rawOpticalImage(ds._source.ds_id, ctx).url)`, which is `undefined` because `.url` doesn't exist on `Promise`. This line was intended to be `(await rawOpticalImage(ds._source.ds_id, ctx)).url`.


metaspace/graphql/src/modules/dataset/controller/Dataset.ts
```diff
@@ -171,13 +171,13 @@ const DatasetResolvers: FieldResolversFor<Dataset, DatasetSource> = {
...
-    return await rawOpticalImage(ds._source.ds_id, ctx).url;
+    const opticalImage = await rawOpticalImage(ds._source.ds_id, ctx);
+    return opticalImage ? opticalImage.url : null;
   },
 
   async rawOpticalImageUrl(ds, _, ctx) {
-    // @ts-ignore
-    return await rawOpticalImage(ds._source.ds_id, ctx).url;
+    const opticalImage = await rawOpticalImage(ds._source.ds_id, ctx);
+    return opticalImage ? opticalImage.url : null;
```
Ren22: `async optical image` and `async rawopticalimage` have the same implementation?

Ren22:
Some conflicts detected

metaspace/graphql/src/modules/dataset/controller/Dataset.ts
LachlanStuart: Yeah, they're duplicates. `opticalImage` is deprecated in the `dataset.graphql` file. I've changed the UI to use `rawOpticalImageUrl` instead, and python-client has always used `rawOpticalImageUrl`. I'm going to leave `opticalImage` there for now just in case something is using the API, but it can probably be removed next time this is refactored.

PR #141, opened 2018-11-20
Add links to Gitter

### LINKED ISSUE #133: Add a gitter link to the website (Help)

LachlanStuart:
Resolves #133 

#### README.md badge
image

#### About page
image

#### Help page
image


LachlanStuart:
@Ren22 I went with text for consistency with what's already there. It's unsustainable to follow the current approach of adding paragraphs to the body of the page as if it were a document, so I don't want to try to evolve this design when we should be throwing it out and starting from scratch. 

If/when we decide to make changes to these pages, I feel like we should go with a footer similar to Stack Overflow or GitHub, where on the left we'd put: "Terms of Service/EULA", "Privacy Policy", "About Us", "Help", and on the right: "Twitter", "Gitter", "Email", "GitHub". Having about/help pages divided into sections ([like this](https://zenkit.com/documentation#/lists/HJQBQ1Hw/sections/ByS1_KfA/entries/B1yZtYf0?_k=nsqlsp)) would also help to clean the pages up.

PR #140, opened 2018-11-19
I had to change these lines to support earlier versions of python(3.5)

Ren22:

PR #142, opened 2018-11-20
Hotfix: Changed readme: added Python version 

Ren22:
+ corrected the link for pip install
Recreated PR

### DIFF
--- a/metaspace/python-client/README.md
+++ b/metaspace/python-client/README.md
@@ -10,7 +10,7 @@ Python module that provides programmatic access to a running instance of [`METAS
 
 ## Installation
 
-You need Python 3 to use this library.
+You need to have **Python 3.6 or higher** to use this library.
 
 `pip install git+https://github.com/metaspace2020/metaspace#subdirectory=metaspace/python-client`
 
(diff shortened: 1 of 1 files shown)


metaspace/python-client/README.md
```diff
@@ -10,15 +10,15 @@ Python module that provides programmatic access to a running instance of [`METAS
 
 ## Installation
 
-You need Python 3 to use this library.
+You need to have **Python 3.6 or higher** to use this library.
```
intsco: So Python 3.6 or 3.5?

Ren22: (even though it should work with 3.5 as well now)

LachlanStuart: Is Sergio on 3.6 now? Because we shouldn't say 3.6 if we're actually planning to support 3.5.

Ren22: he is on 3.5 with the changes from https://github.com/metaspace2020/metaspace/pull/140
Again, I cannot guarantee 100% support of 3.5 as I didn't test it. I personally used 3.6. If user hits the corner case again with 3.5 version, it will be our fault by misleading him with this readme

LachlanStuart: I've experienced this before with browser versions - it causes extra work if you're not clear what is supported and what isn't. If we decide to say we only support 3.6 here, PRs like #140 shouldn't happen again - the response should just be to tell the user to upgrade. Otherwise we'll get into an annoying cycle of cleaning up code by using new features, then messing it up again to support older versions.

Ren22: let's hope that it won't happen again, from now on we can use 3.6 syntax more freely. I'll ask Sergio to run it with 3.6 in the future

Ren22: merging this PR

intsco:
@Ren22 what about the branch, do you still need it?

Ren22:
@intsco nope, I'll delete it

PR #143, opened 2018-11-20
Added project manager(s)

Ren22:
image
I had to tweak some flex box values as the name could go beyond the allocated space

### DIFF
--- a/metaspace/webapp/src/modules/Project/ProjectsListItem.vue
+++ b/metaspace/webapp/src/modules/Project/ProjectsListItem.vue
@@ -13,6 +13,11 @@
           {{project.numMembers | plural('Member', 'Members')}}
         </div>
         <div class="info-line">
+          <span>Project manager<span v-if="projectManagers.length>1">s</span>:</span>
+          <span v-for="(manager, ind) in projectManagers">
+            <b>{{manager.user.name}}<span v-if="ind+1 < projectManagers.length" >, </span></b></span>
+        </div>
+        <div class="info-line">
           <span v-if="project.latestUploadDT != null">
             Last submission <b>{{formatDate(project.latestUploadDT)}}</b>
           </span>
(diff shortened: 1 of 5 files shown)


metaspace/webapp/src/modules/Project/ProjectsListItem.vue
```diff
@@ -146,8 +158,13 @@
     }
   }
 
+  .info {
+    flex-basis: 70%;
+  }
+
   .item-body {
     display: flex;
+    flex-wrap: wrap;
```
LachlanStuart: `.item-body` holds both the details on the left side, and the actions on the right side. Putting `flex-wrap` here causes the container to seriously break when the details get too big:

image

metaspace/webapp/src/modules/Project/ProjectsListItem.vue
```diff
@@ -157,7 +174,6 @@
 
   .info-line {
     overflow: hidden;
-    white-space: nowrap;
```
LachlanStuart: 1. Important note about CSS here: the combination of `overflow`, `white-space` and `text-overflow` is special. `text-overflow` does nothing without the other two. If you remove `white-space`, you should also remove `text-overflow`. [The explanation](https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow) isn't very exciting, but this effect has stumped me a few times.
2. I don't think you should actually remove `text-overflow` here. We tried having variable-height list items with the datasets page and it looked pretty bad. It's better to ensure these are always a consistent number of lines by truncating overflow like this:
image

3. You can get the above behavior by undoing your changes and adding these:
```
  .info {
    overflow: hidden; // The reason that .info-line could still overflow was that this .info had unrestricted width because it was also allowed to overflow.
    flex: auto; // auto allows both growing and shrinking. This means that the info area will expand to fill all remaining space
    padding-right: 12px;
  }
  .actions {
    flex: none; // none prevents growing and shrinking. This prevents the info area from trying to steal space from the actions area.
  }
```

Ren22: @LachlanStuart which browser do you use? I just got this on Chrome after these changes, I suspect that it caused by created span elements for each of the group manager
image

Ren22: Your Long text comes just as one line I guess

Ren22:
@LachlanStuart updated, but tests still fail..

PR #152, opened 2018-11-23
ELK logs collection

intsco:


ansible/aws/templates/master.yml.template
```diff
@@ -0,0 +1,73 @@
...
+              "patterns": [
+                "%{GREEDYDATA:path}/%{GREEDYDATA:file}",
+              ]
+            }
+          },
+          {
+            "grok" : {
+              "field": "raw_message",
+              "patterns": [
+                "%{TIMESTAMP_ISO8601:timestamp} - %{LOGLEVEL:level} - %{NOTSPACE:component}\\[%{GREEDYDATA:thread}\\] - %{NOTSPACE:file_line} - %{GREEDYDATA:message}",
```
LachlanStuart: I think `GREEDYDATA` used for `thread` should just be `DATA`. `GREEDYDATA` will grow to consume parts of the message if they match, whereas `DATA` will try to stop consuming the message as early it can, e.g. on [grokconstructor](http://grokconstructor.appspot.com/do/match#result) matching
```
2018-11-26 10:37:36,029 - INFO - engine[Thread-1] - search_job.py:201 - Dataset config: foo [] - foo - 
```
with pattern:
`%{TIMESTAMP_ISO8601:timestamp} - %{LOGLEVEL:level} - %{NOTSPACE:component}\[%{GREEDYDATA:thread}\] - %{NOTSPACE:file_line} - %{GREEDYDATA:message}`

image

vs with pattern:
`%{TIMESTAMP_ISO8601:timestamp} - %{LOGLEVEL:level} - %{NOTSPACE:component}\[%{DATA:thread}\] - %{NOTSPACE:file_line} - %{GREEDYDATA:message}`
image

intsco: Thanks, that's a good point! Somehow it worked when I tested it on the logs from the staging.

LachlanStuart: It generally would work. It's only if a log message contained `] - (some text without spaces) -` when it would break. Unlikely to happen, but annoying to deal with if it did happen.

PR #153, opened 2018-11-23
ELK compatible logs

intsco:
Changes needed for collecting logs from the prod instance


intsco:
Not needed because #152 is being merged and deployed to prod

PR #157, opened 2018-11-28
Fixed a proper path return for isotope image path

Ren22:
There was an obsolete call from `v0.11` release because we had defined
`DEFAULT_CONFIG = { ...
    'moldb_url': 'http://metaspace2020.eu/mol_db/v1' ...}`
But in principle this code line just returned the hostname , so I added it to `get_config`

### DIFF
--- a/metaspace/python-client/metaspace/sm_annotation_utils.py
+++ b/metaspace/python-client/metaspace/sm_annotation_utils.py
@@ -20,7 +20,7 @@ def _extract_data(res):
 
 def get_config(host, email=None, password=None, verify_certificate=True):
     return {
-        'host': '{}'.format(host),
+        'host': host,
         'graphql_url': '{}/graphql'.format(host),
         'moldb_url': '{}/mol_db/v1'.format(host),
         'signin_url': '{}/api_auth/signin'.format(host),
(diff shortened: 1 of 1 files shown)


metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -20,6 +20,7 @@ def _extract_data(res):
 
 def get_config(host, email=None, password=None, verify_certificate=True):
     return {
+        'host': '{}'.format(host),
```
LachlanStuart: I think this `.format` is redundant. Couldn't it just be `'host': host`?

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -518,7 +519,7 @@ def status(self):
 
     @property
     def _baseurl(self):
-        return self._gqclient.url.rsplit("/", 1)[0]
+        return self._gqclient._config['host']
```
LachlanStuart: I think [this](https://github.com/metaspace2020/metaspace/blob/c6d27649a304df3495b2c3aca9efecd325765531/metaspace/python-client/metaspace/sm_annotation_utils.py#L535) should be updated to use `self._baseurl` so that it gets the same fix.

intsco: Ideally one should not access a private field or method of a class, `_config` is private field of `GraphQLClient`.
How about adding a public method or just make a public field `host` in `GraphQLClient`?

Ren22:
@LachlanStuart thanks for a good catch, I was a bit hectic

metaspace/python-client/metaspace/sm_annotation_utils.py
Ren22: @intsco Thank you, this is an important note, should be fixed now

PR #159, opened 2018-11-28
Fix dataset change notifications

LachlanStuart:
#### Engine changes:

* Removes `NEW`, `INDEXING` and `DELETED` dataset statuses:
    * `NEW` was an unnecessary step because new datasets are immediately updated to `QUEUED`
    * Using `INDEXING` for metadata-only updates caused the problem of not knowing which status to set the dataset back to after the update was finished. (e.g. it could have been `QUEUED` or `ANNOTATING` prior to the update, but after the update it's unknown which status is correct)
    * `DELETED` wasn't a real status
* Adds `DaemonAction` enum to formalize the list of actions sent to daemons
* Restructure status updates so that they include the `action` that the daemon is processing, and the `stage` of processing. This allows the front-end to make better decisions about which updates to show.
* Add `'refresh': 'wait_for'` to ES updates. This will slow things down, but it will guarantee that ES changes have propagated before a status update is sent.
* Prevent the `dataset` ES document from being deleted during dataset reindexing. This prevents the dataset from briefly disappearing during reindexing, or permanently disappearing if reindexing fails.

#### GraphQL changes:

* For subscriptions, get the user's JWT from the payload & unpack it into context
* Refactor pubsub interface to enforce type safety
* Propagate all status updates to all clients that are allowed to see them, so that the datasets get updated in apollo cache without needing to refetch any queries
* [Fought against TypeScript](https://www.reddit.com/r/IdiotsFightingThings/comments/9zs3to/stupid_glass_sucks/) to try to get the apollo subscription to work nicely as an async generator, before giving up and implementing the async generator interface manually.
* Added "relationships" for datasets, so that notifications know if and how the current user is related to a dataset being updated

#### Webapp changes:

* Supply a JWT with each new GraphQL subscription, and add a bunch of logic to try to ensure that subscriptions are always kept alive and correctly authenticated
* Move around the dataset status updates responsibility:
    * `MetadataHeader.vue` now keeps a constant subscription active. This ensures that the local cache is kept up to date, and also allows it to show notifications if they're relevant to the current user.
    * `DatasetTable.vue` uses a subscription when you're on the Datasets page to track when new datasets have been added.
        * The apollo cache is now manually updated when datasets are added/removed
        * Things get weird when a dataset's status changes - the dataset stays in the same list but changes status, and then the downstream `computed` properties put them in the correct place.
        * The above two changes mean that this page doesn't ever need to refetch the datasets list, which should prevent the server from getting hammered with queries every time a dataset is updated. This makes it safe to send an update notification for every single dataset change.


intsco:
It looks good to me but there is no possible way for me to test in all in coming days

PR #164, opened 2018-11-29
Updated Jupyter notebooks for private datasets example

Ren22:
Removed `datasetByName` from queries on schemas
Added new Exception message if more than 1 dataset were found
Updated couple of notebooks, so they can work with private datasets after login to metaspace


metaspace/python-client/example/simple_database_queries.ipynb
```diff
@@ -20,25 +20,90 @@
...
-   },
-   "outputs": [],
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "401 <Response [401]>\n",
+      "Unauthorized. Only public but not private datasets will be accessible.\n"
```
LachlanStuart: Output should be cleared, at least on these cells. It's a bit confusing to open a notebook and immediately see errors.

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -42,8 +42,9 @@ def __init__(self, config):
             "password": self._config['usr_pass']
         })
         if res.status_code == 401:
-            print('Unauthorized')
+            print('Unauthorized. Only public but not private datasets will be accessible.')
         elif res.status_code != 200:
+            print('Authorized')
```
LachlanStuart: This didn't show up for me, because `res.status_code` was `200` when I ran it

Ren22: hmm, I removed it, dunnot why it still happened to be in the commit. thanks

LachlanStuart: IMO it's good to have this print statement - otherwise there's no indication that login was successful. It should just go outside of this `if`/`elif` block so that it shows correctly

Ren22: yep, I exactly did it, somehow I just committed earlier changes.
updated

PR #166, opened 2018-12-03
Dataset page optimizations and link to group page

### LINKED ISSUE #120: Add link to Group page from datasets list

LachlanStuart:
Resolves #120

#### Datasets page changes

* Link to group page from dataset items:
    image
    * This menu appears on after hovering over the group name
    * This menu is disabled on the group page, because it doesn't make sense there
* Move GraphQL `thumbnailOpticalImageUrl` into the `Dataset` type so that DatasetItems don't have to do a separate query for it. (Reduces number of GraphQL queries by 100 & increases page rendering speed by ~100ms)
* Move `currentUser` query from `DatasetItem` into parent `DatasetList` component (Removes 100 queries that would cost ~100ms even though Apollo deduplicates them into all into just 1 query)
* Start deferring rendering of datasets once more than 20 appear on the screen. This makes the page show datasets and become interactive about 500ms sooner. 
    * A negative side effect is that the page height jumps a few times while the page is loading
    * People with high-res screens (greater than 1400px high) will see past the initial 20 visible items, which looks a little bit worse because the page moves multiple times after loading
    * I had to use `requestAnimationFrame` to ensure that the datasets were evenly spread out. Otherwise Vue was clumping the 8 deferred batches of datasets to render into 1-3 janky updates instead of 8 smooth updates. This unfortunately adds about 100ms to the total time (but greatly improves interactivity while it is loading)

#### Other changes

* Add `white-space: nowrap` to the "My Account" menu items in the header, so that the notification icon doesn't wrap.
* Add default values for the `inpFdrLvls` and `checkLvl` args in the GraphQL side, because we actually only use a specific set of constants for this value and it's a pain to find out what they should be if they're needed in a nested query.

### DIFF
--- a/metaspace/graphql/src/modules/dataset/controller/Dataset.ts
+++ b/metaspace/graphql/src/modules/dataset/controller/Dataset.ts
@@ -7,6 +7,7 @@ import {Dataset} from '../../../binding';
 import {rawOpticalImage} from './Query';
 import getScopeRoleForEsDataset from '../util/getScopeRoleForEsDataset';
 import {logger} from '../../../utils';
+import {db} from '../../../utils/knexDb';
 
 const DatasetResolvers: FieldResolversFor<Dataset, DatasetSource> = {
   id(ds) {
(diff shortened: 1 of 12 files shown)


metaspace/graphql/src/modules/dataset/controller/Dataset.ts
```diff
@@ -186,6 +187,18 @@ const DatasetResolvers: FieldResolversFor<Dataset, DatasetSource> = {
   async rawOpticalImageUrl(ds, _, ctx) {
     const opticalImage = await rawOpticalImage(ds._source.ds_id, ctx);
     return opticalImage ? opticalImage.url : null;
+  },
+
+  async thumbnailOpticalImageUrl(ds, args, ctx) {
```
Ren22: I thought we don't plan to use knex anymore rather TypeORM instead?

LachlanStuart: Good point. I was just being lazy and copying existing code. I've replaced it with a TypeORM `connection.query()` call.

PR #173, opened 2018-12-05
Fix recently introduced bugs on ImageAlignmentPage

LachlanStuart:
* `this.opticalImageWidth`, etc. were never set because of the incorrect logic on `this.$refs.scan`, causing the image aligner to break
* It was not returning to the datasets list after completion, because opening the page would cause an extra entry to appear in the browser history for adding the `mdtype` filter. This meant that when `this.$router.go(-1)` was called, it would just undo that `mdtype` filter but stay on the page. This bug was created in [this commit](https://github.com/metaspace2020/metaspace/pull/149/commits/1dd2961ae42a7d2a03e784e6a27b61be217da309#diff-9b42577ce837c0f39ea6340a7bd93ca9L33), where I tried to fix browser history for filters.

Also fixes a bug on the Annotations page where it would never show the help text when no annotations are found for a particular dataset. The issue was that the `metadataType` filter (added for LC-MS) is always active, making the page assume that there were no annotations due to the filters being too strict.

image


LachlanStuart:
@Ren22 That works fine for me. What error does `sm-api` report?

Ren22:
again something like 
```2018-12-10 18:23:44,874 - INFO - api[CP Server Thread-1] - api.py:75 - Received ADD_OPTICAL_IMAGE request: {'url': 'http://localhost:4201/fs/raw_optical_images/e216ac90a73bfd901500e828c74e29be', 'transform': [[8.343763037129754, 0, 320.80100125156446], [3.2478369592023773e-15, 8.344466800804817, 249.9906103286385], [1.9346928527190762e-17, 0, 1]], 'doc': {}}
2018-12-10 18:23:44,878 - INFO - api[CP Server Thread-1] - dataset_manager.py:183 - Adding optical image to "2018-10-23_13h19m39s" dataset
2018-12-10 18:23:44,883 - ERROR - api[CP Server Thread-1] - api.py:100 - HTTPConnectionPool(host='sm-graphql', port=4201): Max retries exceeded with url: /fs/raw_optical_images/delete/8d999d6b04985fcee69615c9702ca36c (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fec1748cba8>: Failed to establish a new connection: [Errno 111] Connection refused',))
Traceback (most recent call last):
  File "/home/renat/miniconda3/envs/sm/lib/python3.6/site-packages/urllib3/connection.py", line 159, in _new_conn
    (self._dns_host, self.port), self.timeout, **extra_kw)
  File "/home/renat/miniconda3/envs/sm/lib/python3.6/site-packages/urllib3/util/connection.py", line 80, in create_connection
    raise err
  File "/home/renat/miniconda3/envs/sm/lib/python3.6/site-packages/urllib3/util/connection.py", line 70, in create_connection
    sock.connect(sa)
ConnectionRefusedError: [Errno 111] Connection refused```

PR #174, opened 2018-12-05
Fix reset password

### LINKED ISSUE #170: Reset Password does nothing if user doesn't exist / is inactive

LachlanStuart:
The issue is that Reset Password was silently failing for users that don't exist or have not been activated since the migration. GraphQL was only checking for users with a matching verified email address, and the web app wasn't handling server errors correctly.

This change does several things:
* Makes the Reset Password flow work for inactive users (as long as they already have a name). Once they've set their password, their email is automatically validated.
* Makes the Reset Password flow send an email directing the user to the Create Account page if the email is not recognized
* Adds error handling to the Reset Password dialog
* Adds more tests & groups the existing tests by function (hide whitespace changes - most of the file's changes are just indentation)

Resolves #170

### DIFF
--- a/metaspace/graphql/src/modules/auth/controller.ts
+++ b/metaspace/graphql/src/modules/auth/controller.ts
@@ -16,6 +16,7 @@ import {
   initOperation,
   resetPassword,
   sendResetPasswordToken,
+  validateResetPasswordToken,
   verifyEmail,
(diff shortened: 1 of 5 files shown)


LachlanStuart:
Test passed after rerunning

PR #177, opened 2018-12-10
Fix cluster stop behaviour

intsco:
- Check that queue is empty before stopping
- Stop all daemons on master before stopping machine


intsco:
Resolves #177

PR #182, opened 2018-12-13
Use AWS credentials from config

intsco:
Stop relying on credentials stored in ~/.aws/credentials

### DIFF
--- a/ansible/aws/aws_instance_manager.py
+++ b/ansible/aws/aws_instance_manager.py
@@ -6,19 +6,20 @@ from datetime import datetime, timedelta
 from os import path
 from subprocess import check_output
 import sys
-
 import boto3
 from yaml import load
 
 
 class AWSInstManager(object):
 
-    def __init__(self, key_name, conf, region, dry_run=False, verbose=False):
-        self.key_name = key_name
-        self.region = region
+    def __init__(self, conf, aws_conf, dry_run=False, verbose=False):
+        self.key_name = aws_conf['key_name']
         self.dry_run = dry_run
-        self.ec2 = boto3.resource('ec2', region_name=region)
-        self.ec2_client = boto3.client('ec2', region_name=region)
+        cred_dict = dict(aws_access_key_id=aws_conf['aws_access_key_id'],
+                         aws_secret_access_key=aws_conf['aws_secret_access_key'])
+        session = boto3.session.Session(**cred_dict)
+        self.ec2 = session.resource('ec2', region_name=aws_conf['region'])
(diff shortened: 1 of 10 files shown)


ansible/aws/aws_instance_manager.py
```diff
@@ -6,19 +6,20 @@
...
-        self.key_name = key_name
-        self.region = region
+    def __init__(self, conf, aws_conf, dry_run=False, verbose=False):
+        self.key_name = aws_conf['key_name']
         self.dry_run = dry_run
-        self.ec2 = boto3.resource('ec2', region_name=region)
-        self.ec2_client = boto3.client('ec2', region_name=region)
+        cred_dict = dict(aws_access_key_id=aws_conf['aws_access_key_id'],
+                         aws_secret_access_key=aws_conf['aws_secret_access_key'])
+        session = boto3.session.Session(**cred_dict)
```
LachlanStuart: `cred_dict` looks unnecessary - the two kwargs could be passed directly here. The same pattern is in `ansible/aws/cluster_auto_start_daemon.py` and `metaspace/engine/sm/engine/sm_daemons.py`. Is there something I'm missing?

ansible/roles/sm_cluster_autostart/tasks/main.yml
```diff
@@ -16,6 +16,7 @@
     state: latest
     executable: pip3
   with_items:
+    - pip
```
LachlanStuart: Not really an issue here, but just so you know, some future version of pip is going to remove [the `--process-dependency-links` feature that we rely on](https://github.com/metaspace2020/metaspace/blob/0b2e0d51ab7053b7925c25c826e7d146dbe98742/ansible/roles/sm_engine/tasks/main.yml#L7), according the deprecation warning:

```
pip install --process-dependency-links -e .
DEPRECATION: Dependency Links processing has been deprecated and will be removed in a future release. A possible replacement is PEP 508 URL dependencies. You can find discussion regarding this at https://github.com/pypa/pip/issues/4187.
```

Hopefully they'll actually have a solution in place before removing this feature. If not, then we'll have to start pinning the version of pip installed.

ansible/aws/aws_instance_manager.py
intsco: I think it was just copied from another place where an expression was already too big.

ansible/roles/sm_cluster_autostart/tasks/main.yml
intsco: That's actually a good point

PR #187, opened 2018-12-21
Several small changes

LachlanStuart:
See commit messages

### DIFF
--- a/metaspace/webapp/src/components/ImageLoader.vue
+++ b/metaspace/webapp/src/components/ImageLoader.vue
@@ -11,7 +11,7 @@
              :style="imageStyle"
              @click="onClick"
              @wheel="onWheel"
-             @mousedown="onMouseDown"
+             @mousedown.left.prevent="onMouseDown"
              ref="visibleImage"
              class="isotope-image"/>
       </div>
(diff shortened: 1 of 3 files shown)


metaspace/webapp/src/components/ImageLoader.vue
```diff
@@ -11,7 +11,7 @@
              :style="imageStyle"
              @click="onClick"
              @wheel="onWheel"
-             @mousedown="onMouseDown"
+             @mousedown.left.prevent="onMouseDown"
```
Ren22: haven't seen such a combination `mousedown.left.prevent` before, interesting

Ren22: so it prevents default ?

LachlanStuart: `.prevent` adds `event.preventDefault()` and `.left` adds the filter `if (event.button === 0)`.

Vue has lots of these weird modifiers. You can also do things like `@keyup.ctrl.enter`. There's a list [here](https://vuejs.org/v2/guide/events.html#Event-Modifiers)

PR #190, opened 2019-01-03
Queue exit message support

intsco:
* It seems like the only way to prevent new annotation jobs from starting while cluster is being stopped

### DIFF
--- a/ansible/aws/cluster_auto_start_daemon.py
+++ b/ansible/aws/cluster_auto_start_daemon.py
@@ -177,12 +177,13 @@ class ClusterDaemon(object):
         return True
 
     def cluster_stop(self):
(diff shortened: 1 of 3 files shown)


ansible/aws/cluster_auto_start_daemon.py
```diff
@@ -13,10 +14,56 @@
...
+
+    def send_queue_exit_message(self):
+        try:
+            ch = self._get_channel()
+            self._declare_queue(ch)
+            ch.basic_publish(exchange='',
+                             routing_key=self.qname,
+                             body=json.dumps({'action': 'exit'}),
+                             properties=pika.BasicProperties(
+                                 delivery_mode=2,  # make message persistent
```
LachlanStuart: I don't think this message needs to be, or should be persistent

PR #195, opened 2019-01-08
ECMDB description

intsco:


LachlanStuart:
@intsco  A snapshot test failed. It can be fixed by running `yarn run test -u` and committing the updated `.snap` file.

intsco:
Not sure why another snapshot got updated
https://github.com/metaspace2020/metaspace/pull/195/files#diff-11307891fa55da5f3cba1c83e04700a9

LachlanStuart:
@intsco `FilterPanel.spec.ts.snap` was updated because it depends on config (which is bad, but another matter). On your PC it doesn't show the Data Type filter because there's only one data type. On the CI server, Renat's PC and my PC we have LC-MS enabled in the webapp config.

The easiest way to get the CI to pass would be to revert the changes to `FilterPanel.spec.ts.snap` and keep the changes in `MetadataEditor.spec.ts.snap`.

PR #197, opened 2019-01-09
Improves the look of scale bar

Ren22:
It grows/shrinks proportionally to the zoomed image
Disables OY axis when pixel size OX = OY
Hides the bar if it grows too big

### DIFF
--- a/metaspace/webapp/src/components/ImageLoader.vue
+++ b/metaspace/webapp/src/components/ImageLoader.vue
@@ -25,14 +25,16 @@
       <div :style="cssProps"
            :class="{pixelSizeX: pixelSizeIsActive}"
            title="Click to change the color"
-           @click="onClickScaleBar()">
-        {{scaleBarValX}}
+           @click="onClickScaleBar()"
+           v-if="!notExceedsOxViewArea">
+        <div :class="{pixelSizeXText: pixelSizeIsActive}">{{scaleBarValX}}</div>
       </div>
       <div :style="cssProps"
            :class="{pixelSizeY: pixelSizeIsActive}"
            title="Click to change the color"
-           @click="onClickScaleBar()">
-        {{scaleBarValY}}
+           @click="onClickScaleBar()"
+           v-if="!yIsNotEqualX">
+        <div :class="{pixelSizeYText: pixelSizeIsActive}">{{scaleBarValY}}</div>
       </div>
       <palette v-show="paletteIsVisible" class="color-picker" @colorInput="val=>updateColor(val)" />
     </div>
@@ -55,7 +57,6 @@
  import {quantile} from 'simple-statistics';
  import resize from 'vue-resize-directive';
  import config from '../clientConfig.json';
- import { round } from 'lodash-es';
  import Palette from './Palette.vue'
 
  const OPACITY_MAPPINGS = {
@@ -149,7 +150,11 @@
        tmId: 0,
        scaleBarColor: '#000000',
        paletteIsVisible: false,
-       scaleBarShadow: '#FFFFFF'
+       scaleBarShadow: '#FFFFFF',
+       scaleBarOXVal: 50,
+       scaleBarOYVal: 50,
+       oXExceeding: false,
+       oYExceeding: false
      }
    },
    components: {
@@ -173,6 +178,18 @@
      this.isUnmounted = true;
    },
    computed: {
+     notExceedsOxViewArea() {
+       return this.oXExceeding
+     },
+
+     notExceedsOyViewArea() {
+       return this.oYExceeding
+     },
+
+     yIsNotEqualX() {
+       return this.pixelSizeX === this.pixelSizeY
+     },
+
      isIE() {
        if (window.navigator.userAgent.indexOf('MSIE') > 0 ||
        window.navigator.userAgent.indexOf('Trident/') > 0) {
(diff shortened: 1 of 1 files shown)


Ren22:
I'll add the option to hide the bar in the next commit

metaspace/webapp/src/components/ImageLoader.vue
```diff
@@ -173,6 +178,18 @@
...
+     notExceedsOxViewArea() {
+       return this.oXExceeding
+     },
+
+     notExceedsOyViewArea() {
+       return this.oYExceeding
+     },
+
+     yIsNotEqualX() {
+       return this.pixelSizeX === this.pixelSizeY
```
LachlanStuart: Don't be afraid to put simple logic like this directly in the template. As long as it's not too complicated, it's actually easier to read code like this inline in the template, because you don't have to jump down the file to find the `computed` property the template refers to.

metaspace/webapp/src/components/ImageLoader.vue
```diff
@@ -181,21 +198,35 @@
...
-         return `${round((this.image.naturalWidth /
-           (this.zoom * this.visibleImageWidth)) * this.scaleBarSizeVal * this.pixelSizeX, 0)} µm`
+         let ceiledVal = Math.ceil(Math.round((this.image.naturalWidth /
+           (this.zoom * this.visibleImageWidth)) * this.scaleBarSizeBasis * this.pixelSizeX) / 10) * 10;
+         let notCeiledVal = (this.image.naturalWidth /
+           (this.zoom * this.visibleImageWidth)) * this.scaleBarSizeBasis * this.pixelSizeX;
+         let diff = ceiledVal - notCeiledVal;
+         let addedVal = diff/this.pixelSizeX*((this.zoom * this.visibleImageWidth)/this.image.naturalWidth);
+         this.scaleBarOXVal = this.scaleBarSizeBasis + addedVal;
+         this.oXExceeding = Math.round(this.scaleBarOXVal) >= this.parentDivWidth/3;
```
LachlanStuart: `computed` functions should never change `data` variables. There are two reasons for this:

1. `computed` functions are executed somewhat unpredictably - Vue will remember their previous return value and only re-run it the function if some of their dependencies change. This means that if more than one place changes the `data` value, you can't rely on it having updated consistently because Vue may have skipped execution of the `computed`, or may have re-triggered the `computed` due to unrelated variable changes.
2. Because it's not obvious that the `data` values are updating dynamically based on other variables, it's easy to accidentally use them in other `computed` properties and make circular dependency loops. 

There are a couple other ways to achieve the same goal without risking these problems:

1. Return multiple values from the `computed`. Since `scaleBarOXVal`, `oXExceeding` and `scaleBarValX` all have a significant amount of common logic behind them, they could all continue to be calculated here, but instead of being assigned to an external `data` variable, they could be returned in an object that contains all 3 values.
2. Break up the `computed`s so that common calculations can be reused from other `computed`s.

Because there is so much common logic, I'd recommend returning multiple values.

metaspace/webapp/src/components/ImageLoader.vue
```diff
@@ -181,21 +198,35 @@
...
      scaleBarValX() {
        if (this.pixelSizeIsActive && this.visibleImageWidth !== 0 && !this.isIE) {
-         return `${round((this.image.naturalWidth /
-           (this.zoom * this.visibleImageWidth)) * this.scaleBarSizeVal * this.pixelSizeX, 0)} µm`
+         let ceiledVal = Math.ceil(Math.round((this.image.naturalWidth /
+           (this.zoom * this.visibleImageWidth)) * this.scaleBarSizeBasis * this.pixelSizeX) / 10) * 10;
+         let notCeiledVal = (this.image.naturalWidth /
+           (this.zoom * this.visibleImageWidth)) * this.scaleBarSizeBasis * this.pixelSizeX;
+         let diff = ceiledVal - notCeiledVal;
+         let addedVal = diff/this.pixelSizeX*((this.zoom * this.visibleImageWidth)/this.image.naturalWidth);
```
LachlanStuart: Please be consistent with code style. Here all of the `/` and `*` operators should have spaces on either side, to match the rest of the file.

metaspace/webapp/src/components/ImageLoader.vue
```diff
@@ -173,6 +178,18 @@
      this.isUnmounted = true;
    },
    computed: {
+     notExceedsOxViewArea() {
+       return this.oXExceeding
+     },
+
+     notExceedsOyViewArea() {
+       return this.oYExceeding
```
LachlanStuart: There's no need to wrap `data` properties with `computed` getters like this. Templates should just directly reference `oXExceeding` and `oYExceeding`.

metaspace/webapp/src/components/ImageLoader.vue
```diff
@@ -599,43 +630,43 @@
...
    position: absolute;
-   width: 100px;
-   height: 10px;
-   bottom: 58px;
-   left: 10px;
-   z-index: 3;
+   height: var(--scaleBarY-size);
+   bottom: 20px;
+   left: 20px;
+   box-shadow: 1px -1px 1px var(--scaleBarShadow-color);
```
LachlanStuart: To match the style used in scientific papers there shouldn't be any text shadow, at least by default. I also think that the bar should be a bit thicker.

metaspace/webapp/src/components/ImageLoader.vue
```diff
@@ -181,21 +198,35 @@
...
+     scaleBarSizeBasis() {
+       return 50
      },
 
      scaleBarValX() {
        if (this.pixelSizeIsActive && this.visibleImageWidth !== 0 && !this.isIE) {
-         return `${round((this.image.naturalWidth /
-           (this.zoom * this.visibleImageWidth)) * this.scaleBarSizeVal * this.pixelSizeX, 0)} µm`
+         let ceiledVal = Math.ceil(Math.round((this.image.naturalWidth /
+           (this.zoom * this.visibleImageWidth)) * this.scaleBarSizeBasis * this.pixelSizeX) / 10) * 10;
```
LachlanStuart: IMO, this way of choosing a scale is still sub-optimal. For large images or high pixel sizes it gives weird-looking numbers like `1480 µm`, and for small images it gives a `10 µm` bar that stretches awkwardly far across the screen.

It would be better if you manually chose a set of aesthetically pleasing numbers (e.g. 1, 2, 2.5, 5, 10, 20, 25, 50, 100, 200, 250, 500, 1000, etc. µm), check the width of each one and choose the one that is closest to the target width.

metaspace/webapp/src/components/ImageLoader.vue
Ren22: Thanks for a suggestion. I have a question: when created in a computed, object properties won't be reactive, is there a way to make them so?
```     scaleBarAxisObj() {
       return {
         scaleBarShownPxVal: 50,
         axisExceeding: false,
         scaleBarVal: pixelSizeAxis => {
           if (this.pixelSizeIsActive && this.visibleImageWidth !== 0 && !this.isIE) {
             let ceiledVal = Math.ceil(Math.round(this.image.naturalWidth /
               (this.zoom * this.visibleImageWidth) * this.scaleBarSizeBasis * pixelSizeAxis) / 10) * 10;
             let notCeiledVal = (this.image.naturalWidth /
               (this.zoom * this.visibleImageWidth)) * this.scaleBarSizeBasis * pixelSizeAxis;
             let addedVal = (ceiledVal - notCeiledVal) / pixelSizeAxis *
               (this.zoom * this.visibleImageWidth) / this.image.naturalWidth;
             this.scaleBarShownPxVal = this.scaleBarSizeBasis + addedVal;
             this.axisExceeding = Math.round(this.scaleBarSizeBasis + addedVal) >= this.parentDivWidth / 3;
             return `${ceiledVal} µm`
           }
         }
       }
     },```

**Comment by Ren22** [comment]
@LachlanStuart I've updated this PR, but the test fails, because of set up --frozen-lockfile. I changed the package dom-to-image and this causes this error I guess

File: metaspace/webapp/src/components/ImageLoader.vue [hunk 7]
```diff
@@ -285,6 +295,70 @@
      }
    },
    methods: {
+     scaleBarAxisObj(pixelSizeAxis) {
+       if (this.pixelSizeIsActive && this.visibleImageWidth !== 0 && !this.isIE) {
+         let notCeiledVal = (this.image.naturalWidth /
+           (this.zoom * this.visibleImageWidth)) * this.scaleBarSizeBasis * pixelSizeAxis;
+         let ceiledVal, notCeiledValToSwitch;
+         notCeiledVal !== undefined ? notCeiledValToSwitch = true : notCeiledValToSwitch = false;
```
Ren22: I know, could have replaced it by ```notCeiledVal = notCeiledVal !== undefined```

metaspace/webapp/src/components/ImageLoader.vue
```diff
@@ -285,6 +295,70 @@
      }
    },
    methods: {
+     scaleBarAxisObj(pixelSizeAxis) {
+       if (this.pixelSizeIsActive && this.visibleImageWidth !== 0 && !this.isIE) {
+         let notCeiledVal = (this.image.naturalWidth /
+           (this.zoom * this.visibleImageWidth)) * this.scaleBarSizeBasis * pixelSizeAxis;
+         let ceiledVal, notCeiledValToSwitch;
+         notCeiledValToSwitch = notCeiledVal !== undefined;
+         switch (notCeiledValToSwitch) {
```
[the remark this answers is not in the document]
Ren22: Thanks @LachlanStuart that's a very neat solution

Ren22:
@LachlanStuart thanks for all the effort put while reviewing this PR. I hope it should be ok now

Issue #196, opened 2019-01-09
Ion centroid generation: grouping by  adduct or adduct-neutral loss


intsco:
So that when a new adduct or neutral loss selected, not all theoretical centroids must be recalculated.

PR #206, opened 2019-01-28
Updated S3 link to new isotopic patterns

intsco:


LachlanStuart:
@intsco Has the file changed, or just moved to a different bucket to simplify access rights?

intsco:
@LachlanStuart the filed moved to another bucket for simplicity but the content also changed according to #196

PR #208, opened 2019-01-28
Feature/scale bar palette into tools

Ren22:
Updated tool icons as well 
image


metaspace/webapp/src/modules/Annotations/annotation-widgets/default/Palette.vue
```diff
@@ -1,34 +1,60 @@
...
+        <el-select :value="pickedColor"
+                   style="width: 120px;"
+                   title="Scale_bar_color"
+                   @input="handlerClick">
+          <el-option v-for="color in paletteColors"
+                     :value="color.colorName" :label="color.colorName" :key="color.code">
+          </el-option>
+        </el-select>
+        <el-form-item>
+          <el-checkbox @click="$event.stopPropagation()" @input="setBarVisibility">Disable</el-checkbox>
```
LachlanStuart: Unlike the colormap's "Invert" setting, disabling the scale bar is mutually exclusive with selecting a color. I feel it would be better if "Disabled" was an option in the `<el-select>` instead of a separate checkbox.

Ren22:
@LachlanStuart thank you for the review, moved the scale bar option under the image settings icon. Concerning the `disable` button I'd say that it would be easier and more intuitive to leave the checkbox, as it's straightforward and reduces the number of clicks. Otherwise a user needs to click settings icon, click the colors for scalebar and find transparent option (less intuitve and takes more time imo).

PR #219, opened 2019-02-06
Use codegen for creating graphql schema in engine tests

LachlanStuart:
* Adds a script to graphql to dump its SQL schema to a file
* Moves all 3 graphql scripts into the `bin` directory, and adds package.json scripts for them
* Moves the package.json scripts for `generate-json-graphql-schema` into the graphql project, so that all 3 scripts can be called from graphql
* Changes sm-engine tests to use the autogenerated SQL for setting up the graphql tables
* Changes CircleCI to dump the SQL schema in the `generate-graphql-schema` job and copy it into the correct place during the `test-engine` job


metaspace/engine/sm/engine/tests/graphql_sql_schema.py
```diff
@@ -0,0 +1,13 @@
+import os
+
+schema_path = os.path.join(os.path.dirname(__file__), 'graphql_schema.sql')
```
intsco: Btw, I prefer pathlib to work with files and directories, it's a bit newer API
```
from pathlib import Path
schema_path = Path(__file__) / 'graphql_schema.sql'
schema_path.open('r') as f:
    ...
```

Issue #163, opened 2018-11-29
Data analysis: Colocalized annotations


LachlanStuart:
* @LachlanStuart to start by consulting with Ver, D and Sha about what information & interaction is useful, then proceed with UI design. 
* @intsco has already done some work with correlating ion images
* TF-IDF and SSIM may be better algorithms for finding colocalization
* Clustering colocalized ion images should also be considered

LachlanStuart:
Feedback from T:

- [ ] Add colocalization coefficient to CSV
- [ ] Add comment to CSV if the annotations were truncated due to colocalization coefficient thresholding

*Colocalized annotations accordion section:*

- [ ] Use decimal number for colocalization coefficient
- [ ] Condense annotation subtitles to:  “0.8, 5%, 7.5e+4 | 0.6” with hover text: “MSM score, FDR value, max intensity | co-localization score”
- [ ] Use filter icon instead of list icon
- [ ] Rename "distinct localizations" to "representative spatial patterns" or "representative images" and move the link into the annotation view header (next to the share button)
- [ ] Move the "Colocalization Metric" into a cog icon popup on the colocalization tab, similar to the selector for choosing colormap

*Back-end:*

- [ ] Recalculate space requirements now that the `ion` table exists, and lower the threshold for cosine distance to 0.3

*Stretch goals:*

- [ ] Use the most representative ion image when optical images are missing
- [ ] Add a accordion section for the representative images & make them clickable so that it's easy to move between clusters

LachlanStuart:
Feedback from V: Use a configurable, limited list of algorithms so that we don't overwhelm people (similar to the adducts/dbs limited by GraphQL)

theodev:
Yes, very good point from Vitaly

I'd suggest for now to use only 1 simple co-localization: cosine/Pearson

Once we finish the research / evaluation with Katya, consider replacing it
with the best algorithm / score

LachlanStuart:
More feedback:

- [ ] Include the molecule being colocalized against in first place in lists of colocalizations (including CSV)
    - [x] GraphQL changes
    - [ ] UI changes
- [ ] When hovering over a molecular formula in a Related Annotations list, or in the MainImageHeader, show the list of putative molecule names
- [ ] Fix issue with coloc'd images having different heights (`/annotations?ds=2019-02-01_14h32m52s&sections=35&sort=-colocalization` @ 50% zoom)
- [ ] Feature flag
[...]

PR #220, opened 2019-02-07
Colocalization backend

LachlanStuart:
First piece of #163 

Includes the DB schema and annotation pipeline changes needed for colocalization.

There are 3 entry points to run colocalization:
* via `analyze_colocalization` for a completely self-contained interface, so that algorithms can be run/tested without needing the database, etc.
* via `Colocalization.run_coloc_job_for_existing_ds` (/`scripts/run_colocalization.py`) to add colocalization data to existing datasets, taking the data from the DB and downloading ion images
* via `Colocalization.run_coloc_job_for_new_ds` to take the data from the Spark/Pandas dataframes produced by the annotation pipeline

### DIFF
--- a/metaspace/engine/sm/engine/colocalization.py
+++ b/metaspace/engine/sm/engine/colocalization.py
@@ -11,18 +11,212 @@ from sklearn.cluster import spectral_clustering
 from scipy.stats import spearmanr, pearsonr
 
 from sm.engine.dataset import Dataset
+from sm.engine.ion_mapping import get_ion_id_mapping
 from sm.engine.mol_db import MolecularDB
 from sm.engine.util import SMConfig
 from sm.engine.png_generator import ImageStoreServiceWrapper
 
-logger = logging.getLogger('engine')
+COLOC_JOB_DEL = ('DELETE FROM graphql.coloc_job ' 
+                 'WHERE ds_id = %s AND mol_db = %s')
+
+COLOC_JOB_INS = ('INSERT INTO graphql.coloc_job (ds_id, mol_db, fdr, algorithm, start, finish, error, sample_ion_ids) ' 
+                 'VALUES (%s, %s, %s, %s, %s, %s, %s, %s) ' 
+                 'RETURNING id')
+
+COLOC_ANN_INS = ('INSERT INTO graphql.coloc_annotation(coloc_job_id, ion_id, coloc_ion_ids, coloc_coeffs) ' 
+                 'VALUES (%s, %s, %s, %s)')
 
 ANNOTATIONS_SEL = ('SELECT iso_image_ids[1], sf, adduct, fdr '
                    'FROM iso_image_metrics m '
                    'JOIN job j ON j.id = m.job_id '
                    'WHERE j.ds_id = %s AND j.db_id = %s')
 
-DATASET_CONFIG_SEL = "SELECT mol_dbs, config #>> '{isotope_generation,charge,polarity}' FROM dataset WHERE id = %s"
+DATASET_CONFIG_SEL = ("SELECT mol_dbs, config #>> '{isotope_generation,charge,polarity}' "
+                      "FROM dataset "
+                      "WHERE id = %s")
+
+logger = logging.getLogger('engine')
+
+
+class ColocalizationJob(object):
+    def __init__(self, ds_id, mol_db, fdr, algorithm_name=None, start=None, finish=None,
+                 error=None, ion_ids=None, sample_ion_ids=None, coloc_annotations=None):
+        """
+        Args
+        ----------
+        ds_id: str
+        mol_db: str
+        fdr: float
+        algorithm_name: str
+        start: datetime
+        finish: datetime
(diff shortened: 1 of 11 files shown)


metaspace/engine/sm/engine/png_generator.py
```diff
@@ -13,7 +13,7 @@ class ImageStoreServiceWrapper(object):
     def __init__(self, img_service_url):
         self._img_service_url = img_service_url
         self._session = requests.Session()
-        self._session.mount(self._img_service_url, HTTPAdapter(max_retries=5))
+        self._session.mount(self._img_service_url, HTTPAdapter(max_retries=5, pool_maxsize=100))
```
LachlanStuart: This was needed because the default pool size (10) is quickly exceeded while downloading ion images in parallel

metaspace/engine/sm/engine/colocalization.py
```diff
@@ -0,0 +1,310 @@
...
+        logger.warning('Warning: clustering failed on some cluster sizes', last_error)
+
+    # Find the best cluster, subtracting n/1000 to add a slight preference to having fewer clusters
+    best_cluster_idx = np.argmax([cs - n / 1000 for n, cs, l in results])
+    best_n, best_cluster_score, best_labels = results[best_cluster_idx]
+    logger.debug(f'best with {best_n} clusters (scores: {[(r[0], r[1]) for r in results]})')
+    return best_labels
+
+
+def _get_best_colocs(scores, labels, max_samples, min_score):
```
LachlanStuart: TODO: There are 2 problems in this function:

1. The "Give preference to item in the same cluster" line disregards the colocalization score.
2. `max_samples` is overridden if `num_in_cluster` is higher, which can result in e.g. clusters of 3000 annotations storing 2999 colocalization scores per annotation. 

For simplicity & usefulness of exported data, it would be better to ignore clusters (at least for PCA-based algorithms, where clustering is potentially misleading) and just pick up to `max_samples` IDs scoring above `min_score`.

metaspace/graphql/src/modules/annotation/model.ts
```diff
@@ -0,0 +1,97 @@
...
+  // /** Array of derivatizations, using '-' and '+' prefixes to indicate atoms lost/gained, e.g. ['-HO+HO3P'] */
+  // @Column({ type: 'text' })
+  // derivatizations: string[];
+
+  @Column({ type: 'text' })
+  adduct: string;
+
+  /** 1 or -1 */
+  @Column({ type: 'smallint' })
+  charge: number;
```
LachlanStuart: I'm including the charge here preemptively, because it will help for distinguishing molecules when we start listing structural isomers. Without considering charge, H2O+H and H4O-H have the same simplified formula H3O, but they cannot be considered structural isomers.

LachlanStuart:
@intsco Please review my last 2 commits for fixes to the selection criteria for colocalizations & performance improvements.

With 8000 annotations my computer would run out of memory & swap during `_preprocess_images` (meaning it required >24GB of memory), so I've made several changes to improve this:
* Refactor the way that raw_images are handled before `analyze_colocalization` is called so that the original per-image ndarrays can be garbage collected after they've been combined into one big ndarray
* Refactor `_preprocess_images` so that it does in-place modification of the images
* Added `_downscale_image_if_required` to reduce the resolution of images if they're too big
* Changed the pairwise comparison for Pearson to use `numpy.corrcoef`, because it's MUCH faster

These are the timings of each stage:
* Downloading 8302 images and resizing from 447x436 to 257x251: 37s
* Preprocessing images: 21s
  * Finding 99th & 100th percentile: 18s
  * Scaling & clipping: 3s
* PCA: 10s
* Cosine: 13.5s
* PCA-Cosine: 0.6s
* PCA-Pearson: 0.9s
* PCA-Spearman: 29s
* Clustering @ 50% FDR (8302 annotations): 2m21s
* Picking best colocalizations & saving to DB: 34s
* Total: 4m50s

After tuning it uses a maximum of 9.6GB of RAM on my computer

metaspace/engine/sm/engine/colocalization.py
```diff
@@ -146,51 +151,70 @@ def _get_best_colocs(scores, labels, max_samples, min_score):
...
-    fdrs = np.array([a[2] for a in filtered_annotations])
+    logger.debug(f'Preprocessing images (shape: {images.shape}, dtype: {images.dtype})')
+    _preprocess_images_inplace(images)
 
+    logger.debug('Calculating colocalization metrics')
     pca_images = PCA(min(20, *images.shape)).fit_transform(images)
     cos_scores = pairwise_kernels(images, metric='cosine')
     pca_cos_scores = pairwise_kernels(pca_images, metric='cosine')
-    pca_pear_scores = pairwise_kernels(pca_images, metric=lambda a, b: pearsonr(a, b)[0])
+    pca_pear_scores = np.corrcoef(pca_images)
```
intsco: Vectorized code rocks!

PR #232, opened 2019-02-14
Colocalization GraphQL changes

LachlanStuart:
The main changes from this PR are in `queryFilters.ts`, `schema.graphql` and `resolvers.js`. Almost everything else is just refactoring to support those changes.

With `queryFilters.ts` my thinking is that we're likely to add other types of filtering that can't be satisfied by ElasticSearch. Having `queryFilters` act as a sort of middleware, where it has a chance to intercept the parameters sent to ElasticSearch, and a chance to update the results, was the best solution I could come up with. I'm not overly happy about this architecture though. It seems like it might be a bit difficult to understand.

Refactors / non-"core" changes:
* Removed unused `sm_graphql_ws_public_endpoint` setting
* Moved common GraphQL config into `conf/default.js`. This allows me to add new config settings without breaking everybody else's local setups.
* Refactor `esConnector.js` to TypeScript. (This is split over 2 commits: [1](https://github.com/metaspace2020/metaspace/commit/fb4eb579af1f9756b903ba3f9492ee8fe92eee02), [2](https://github.com/metaspace2020/metaspace/commit/5c6ac48cf139e5a86719b67b4da630735fd6e82a) - I suggest you review them separately because GitHub's combined view makes it a lot harder to see the changes)
* Add a command to CircleCI's `generate-graphql-schema` job to fix installation of bcrypt in the `intsco/webapp` container. I'm not sure why it took so long for this problem to appear, but basically the `alpine` containers all use `musl` instead of `glibc`, and bcrypt only ships with binaries linking against `glibc`. The fix is to install Python so that there's a build environment for bcrypt to rebuild itself.
* Add a `contextCacheGet` function for memoizing repeated/common function calls during the lifespan of a graphql request.
* Remove `context.getCurrentUserProjectRoles()` as it's the same as `context.user.getProjectRoles()`


metaspace/engine/sm/engine/colocalization.py
```diff
@@ -167,8 +167,9 @@ def _downscale_image_if_required(img, num_annotations):
     if zoom_factor > 1:
         return img
     with warnings.catch_warnings():
-        warnings.filterwarnings('ignore', '.*the output shape of zoom\\(\\) is calculated with round\\(\\) '
-                                          'instead of int\\(\\).*')
+        #ignore "UserWarning: From scipy 0.13.0, the output shape of zoom() is calculated with round() instead of int()
```
intsco: Scipy 0.13.0 is an ancient version. We use 1.1.0 in production

LachlanStuart: This warning still exists in [1.2.1](https://github.com/scipy/scipy/blob/master/scipy/ndimage/interpolation.py#L603). It's quite annoying. The only other way of avoiding it is to find a zoom factor where `round()` and `int()` give the same result.

metaspace/engine/sm/engine/colocalization.py
```diff
@@ -313,9 +313,6 @@ def run_coloc_job_for_existing_ds(self, ds_id):
         for mol_db_name in mol_dbs:
             images, ion_ids, fdrs = self._get_existing_ds_annotations(ds_id, mol_db_name, image_storage_type, polarity)
             self._analyze_and_save(ds_id, mol_db_name, images, ion_ids, fdrs)
-            # Release references explicitly, because normally these fields wouldn't be released until
-            # after the next iteration has already allocated its memory inside _get_existing_ds_annotations.
-            del images, ion_ids, fdrs
```
intsco: So, Python preventively cleans these vars?

LachlanStuart:
@intsco The last 2 commits are all the changes I mentioned.

intsco:
https://github.com/metaspace2020/metaspace/blob/master/metaspace/engine/sm/engine/colocalization.py#L45

LachlanStuart:
@intsco Good catch. Fix pushed.

JS is the complete opposite. Default values are evaluated when the function is called, and only if needed.

PR #234, opened 2019-02-18
Group's description added to tabs

### LINKED ISSUE #168: Group/Project Descriptions
It would be cool if these were in Markdown, but that introduces several issues:

* Some markdown libraries are huge. It may be better to render it to HTML on the server
* Security is a big issue with user-submitted content:
    * All forms of scripting must be disabled
    * External images shouldn't be allowed

Ren22:
Markdown supported and sanitized on Graphql server
This adds one more tab (editable textarea), supporting markdown
image
image
I'm planing to connect hint's pop up to the session state, so it won't come up over and over again, also in my plan is to add support for auto save.  

Fixes #168


Ren22:
https://github.com/metaspace2020/metaspace/issues/168

Ren22:
https://github.com/metaspace2020/metaspace/projects/4#card-15628235

metaspace/webapp/src/modules/GroupProfile/ViewGroupPage.vue
```diff
@@ -207,6 +250,17 @@
...
     maxVisibleDatasets = 8;
 
+    @Watch('group')
+    setGroupDescription() {
+      this.groupDescription = this.group && this.group.groupDescription || '';
+    }
+
+    get markedConvert() {
+      if (this.group != null) {
+        return marked(this.group.groupDescription)
```
LachlanStuart: `marked` unfortunately allows images (e.g. with``), and does not let you disable them. 

Images are a problem because they can be used to track our users, or to show inappropriate content from our site. We may eventually support them, but only in a way where we can manually validate them, e.g. by only supporting images served from domains we control, and ask people to send us the images for us to validate & upload.

I suggest that we use `sanitize-html` on the output of `marked` to filter them out.

I had a look at using `sanitize-html` on the client side - it adds about [50kB](https://bundlephobia.com/result?p=sanitize-html@1.20.0) after minification & gzipping. When I saw that, I thought that surely the bundling of `source-map` and `postcss` must be a mistake because they're normally only used during webpack builds, but I've manually checked using `/metaspace/webapp/analyze-size.sh` and can confirm that they are included. I think it's still acceptable because that bundle is relatively small. The alternative would be to change the graphql API to have two fields: `groupDescriptionRawMarkdown` (the raw content for editing) and `groupDescriptionSafeHtml` (the rendered & sanitized version). 

If you are going to do a client-side preview, then I suggest using `sanitize-html` in the browser. If not, then I suggest moving all that logic into graphql.

metaspace/webapp/src/modules/GroupProfile/ViewGroupPage.vue
```diff
@@ -46,6 +46,49 @@
         </el-alert>
       </div>
       <el-tabs v-model="tab">
+        <el-tab-pane name="description" :label="'Description'" lazy >
+          <el-popover
+            placement="top-end"
+            v-if="groupDescrTextarea"
+            trigger="manual"
+            v-model="showHint">
```
LachlanStuart: This popup quickly gets annoying if you have to repeatedly save & edit to see your changes. I suggest using a static link like GitHub. Another option is to remember that the popup has been dismissed in a cookie, but that has the downside that you can't get to the link again after dismissing the popup.

metaspace/webapp/src/modules/GroupProfile/ViewGroupPage.vue
```diff
@@ -331,6 +385,36 @@
...
+      });
+      this.groupDescrTextarea = false;
+      this.refetchGroup()
+      this.$nextTick( ()=> {this.showHint = false;})
+    }
+
+    get linesLen() {
+      if (this.group != null) {
+        if (this.group.groupDescription.split(/\r\n|\r|\n/).length < 5)
+          return 10;
```
LachlanStuart: This is slightly awkward as having 5-9 lines results in a shorter text box than 1-4 lines due to the `< 5` in the line above.

~~I would suggest these changes:~~
* ~~Use `Math.min` and `Math.max` to keep lines between 10 and 50 (50 because it is roughly the most you can fit on a 1080p screen)~~
* ~~Use `this.groupDescription` instead of `this.group.groupDescription` so that it expands as the user is typing~~
* ~~Add +2 so that there is always a bit of space after the last line~~

`<el-input>` supports an attribute for automatically changing the size: `:autosize="{ minRows: 10, maxRows: 50 }"`. It would be better & cleaner to use this.

FYI no modern platforms use only `\r` for line breaks. I couldn't find any explicit acknowledgement of this, but I think that all browsers we support normalize line breaks into just `\n` across all platforms. In IE11, Chrome on Windows and Firefox on Windows, if you make a `<textarea>` and put a line break in it, its `value.length` will be `1` and `.value.charCodeAt` will only return `10`s (the code for `\n`).

metaspace/graphql/utils.js
```diff
@@ -61,6 +63,23 @@ async function wait(ms) {
...
 
+function sanitizeDescr(descriptionText) {
+  return sanitizeHtml(
+    descriptionText,
+    {
+      allowedTags: [ 'h3', 'h4', 'h5', 'h6', 'blockquote', 'p', 'a', 'ul', 'ol',
+        'nl', 'li', 'b', 'i', 'strong', 'em', 'strike', 'code', 'hr', 'br', 'div',
+        'table', 'thead', 'caption', 'tbody', 'tr', 'th', 'td', 'pre'],
+      allowedAttributes: {
+        'a': ['href', 'name', 'target', 'rel']
```
LachlanStuart: I would limit this to just `['href']`, because:
* `name` is obsolete and shouldn't be used by our users anyway (it has the same function as `id`, in that the browser will automatically scroll to it if there's a matching anchor in the URL. It's only useful for incoming links, not outgoing links)
* Allowing users to set `target` would allow annoying inconsistency in link behavior. 
* `rel` gets overwritten by the below tranform anyway, but even if it wasn't, being able to set `rel="preload"` has a similar same privacy issue as images do - it allows 3rd parties to know that a user has accessed the page. 

Also, please don't add to this `utils.js` file, as we're trying to convert graphql to TypeScript. Instead add a new file to the `/graphql/src/utils` directory.

Ren22: thanks, I'll add it to utils directory, for the `rel` attribute - as it doesn't come initially allowed from the lib, it won't work otherwise, I've tested it with several examples.  The rest of attributes `'href', 'name', 'target'` are the default ones if I'm not mistaken.

metaspace/graphql/src/modules/group/controller.ts
```diff
@@ -202,8 +219,8 @@ export const Resolvers = {
     async updateGroup(_: any, {groupId, groupDetails}: any, {user, entityManager}: Context): Promise<Group> {
       await assertCanEditGroup(entityManager, user, groupId);
       logger.info(`Updating '${groupId}' group by '${user!.id}' user...`);
-
       const groupRepo = entityManager.getRepository(GroupModel);
+      groupDetails.groupDescription = sanitizeGroupDescr(groupDetails.groupDescription);
       const group = {...(await groupRepo.findOneOrFail(groupId)), ...groupDetails};
       await groupRepo.save(group);  // update doesn't return updated object;
```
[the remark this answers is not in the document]
Ren22: Good point, I will fix it.

metaspace/webapp/yarn.lock
```diff
@@ -1176,6 +1183,11 @@
   resolved "https://registry.yarnpkg.com/@types/lodash/-/lodash-4.14.116.tgz#5ccf215653e3e8c786a58390751033a9adca0eb9"
   integrity sha512-lRnAtKnxMXcYYXqOiotTmJd74uawNWuPnsnPrrO7HiFuE3npE2iQhfABatbYDyxTNqZNuXzcKGhw37R7RjBFLg==
 
+"@types/marked@^0.6.0":
+  version "0.6.0"
+  resolved "https://registry.yarnpkg.com/@types/marked/-/marked-0.6.0.tgz#e4ac316144a84afda5c2474488d7b9fef3ab9995"
+  integrity sha512-TxwhgR9VsIfRDJ3WwFokG8Xu+ea0nYGDRHdI783WJ983uffatz0ytIeUEIBOwPvRy241KRSNVyywQltuTqDh0w==
```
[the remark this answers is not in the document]
Ren22: It seems that there was a bad revert that removed them from the package.josn list

Ren22:
@LachlanStuart updated this PR. If there are no major concerns, can we please close and merge it? I'm a bit tired of fixing the things.

Ren22:
+ Made the tab invisible for users if there's no description

Ren22:
@LachlanStuart it's not very obvious why tests are failing this time

LachlanStuart:
@Ren22 It's a build issue with CircleCI's docker containers that is affecting all branches. I'll look into it.

metaspace/webapp/src/util.ts
```diff
@@ -109,6 +110,21 @@ function safeJsonParse(json: string) {
   return undefined;
 }
 
+function sanitizeIt(descriptionText: string) {
+  return sanitizeHtml(
+    descriptionText,
+    {
+      allowedTags: [ 'h3', 'h4', 'h5', 'h6', 'blockquote', 'p', 'a', 'ul', 'ol',
+        'nl', 'li', 'b', 'i', 'strong', 'em', 'strike', 'code', 'hr', 'br', 'div',
+        'table', 'thead', 'caption', 'tbody', 'tr', 'th', 'td', 'pre'],
```
LachlanStuart: Some markdown styles get filtered out by this, e.g.
```markdown
# h1
## h2
### h3
~~strikethrough~~
```

Please add `'h1', 'h2', 'h3', 'del'` to this list.

Ren22:
@LachlanStuart is it ok to merge it, even though tests are failing? Snapshots test should be fine though

LachlanStuart:
@Ren22 Yep. I ran them locally and everything seemed fine.

Ren22:
@LachlanStuart cool, then I'm merging it. Thanks very much for the effort put into the review!

PR #237, opened 2019-02-20
Fix rare colocalization error, reduce memory consumption

LachlanStuart:
* Fixes an error that occurs when running against a dataset that has multiple jobs for the same mol_db. Previously it would select all annotations across all jobs, but with duplicate jobs this resulted in duplicate ions being selected. This happened against 4 datasets on Staging:
    * 2017-05-11_13h42m35s HMDB-v2.5
    * 2017-02-02_17h24m26s SwissLipids
    * 2017-05-02_15h37m42s SwissLipids
    * 2017-05-02_15h28m51s ChEBI

* Add config setting to disable colocalization as part of the annotation pipeline

* Reduce worst-case memory consumption by 2-3.6GB. On the dataset with 12k annotations (2018-10-09_03h48m12s), the "score" ndarrays have 12k*12k elements (=576MB each if 4 bytes per element). This PR saves up to:
    * 2GB by freeing `images` after it is no longer needed
    * 2 * 576MB by converting `pca_pear_scores` and `pca_sper_scores` from float64 to float32
    * Another 576MB by not creating a duplicate for `masked_scores` when the mask includes all annotations (i.e. for FDR 0.5)

The motivation for this is that when running against all of Staging the peak memory usage was 13.2 GiB. The spark master uses c4.xlarge instances, which only have 7.5GiB of RAM. While this won't completely stop swapping, it should reduce it substantially. Also, it makes the process less resource-intensive during development.

### DIFF
--- a/metaspace/engine/sm/engine/colocalization.py
+++ b/metaspace/engine/sm/engine/colocalization.py
@@ -77,6 +77,23 @@ class ColocalizationJob(object):
         self.coloc_annotations = coloc_annotations or []
 
 
+class FreeableRef(object):
+    def __init__(self, ref):
+        self._ref = ref
+        self._freed = False
+
+    def free(self):
+        self._ref = None
+        self._freed = True
+
+    @property
+    def ref(self):
+        if self._freed:
+            raise Exception('FreeableRef is already freed')
+        else:
+            return self._ref
+
+
 def _preprocess_images_inplace(imgs):
     """ Clips the top 1% (if more than 1% of pixels are populated),
     scales all pixels to the range 0...1,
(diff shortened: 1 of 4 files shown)


metaspace/engine/sm/engine/colocalization.py
```diff
@@ -180,27 +200,29 @@ def analyze_colocalization(ds_id, mol_db, images, ion_ids, fdrs):
...
-    _preprocess_images_inplace(images)
+    logger.debug(f'Preprocessing images (shape: {images.ref.shape}, dtype: {images.ref.dtype})')
+    _preprocess_images_inplace(images.ref)
 
     logger.debug('Calculating colocalization metrics')
-    pca_images = PCA(min(20, *images.shape)).fit_transform(images)
-    cos_scores = pairwise_kernels(images, metric='cosine')
+    pca_images = PCA(min(20, *images.ref.shape)).fit_transform(images.ref)
+    cos_scores = pairwise_kernels(images.ref, metric='cosine')
+    images.free()
```
intsco: I assume that `del images` didn't provide the needed effect?

intsco: Nice trick

metaspace/engine/sm/engine/colocalization.py
```diff
@@ -311,7 +333,7 @@ def run_coloc_job_for_existing_ds(self, ds_id):
         image_storage_type = Dataset(ds_id).get_ion_img_storage_type(self._db)
         mol_dbs, polarity = self._db.select_one(DATASET_CONFIG_SEL, [ds_id])
 
-        for mol_db_name in mol_dbs:
+        for mol_db_name in mol_dbs[1:2]:
```
intsco: It's not clear why this is needed

metaspace/engine/sm/engine/colocalization.py
LachlanStuart: It's unfortunate that it was needed in the first place, but it seemed to be the most straightforward solution.

metaspace/engine/sm/engine/colocalization.py
LachlanStuart: Oops, that was just for testing. I'll revert it.

PR #240, opened 2019-02-20
Tidy dataset lists & dataset page header

LachlanStuart:
* Removed the "Search results in reverse chronological order" text
* Moved the "Export CSV" button into the line above
* Adjusted the dataset item width so that the list takes the full width available to it
* Centered & greyed out the "No datasets found" message
* Moved the "No datasets found" message into DatasetsList so that it also appears on the Group/Project pages

## Before
image

## After
image

## Before
image

## After
image

## Before
image

## After
image

## Before
image
(Note that the line above the datasets has a 3px overhang on each side)

## After
image

### DIFF
--- a/metaspace/webapp/src/modules/Datasets/DatasetsPage.vue
+++ b/metaspace/webapp/src/modules/Datasets/DatasetsPage.vue
@@ -27,7 +27,6 @@
 
   /* 1 dataset per row by default*/
   #dataset-page-contents {
-    display: inline-block;
     width: 820px;
 
     @media (min-width: 1650px) {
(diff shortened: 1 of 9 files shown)


metaspace/webapp/src/modules/Datasets/DatasetsPage.vue
```diff
@@ -27,8 +27,9 @@
 
   /* 1 dataset per row by default*/
   #dataset-page-contents {
-    display: inline-block;
     width: 820px;
+    margin-left: 5px;
+    margin-right: 5px;
```
Ren22: You don't prefer ```margin: 0 5px;``` syntax?

Ren22:
I like it, and maybe I'd also align tick areas for processing/queued/failed to one line with button (export csv) and List/Summary

metaspace/webapp/src/modules/Datasets/DatasetsPage.vue
LachlanStuart: That was just me being lazy - the combined form overwrites all 4 margins, and I was too lazy to check if `margin-top`/`margin-bottom` were used. I've cleaned this up now.

PR #263, opened 2019-03-15
Off-sample API: basic implementation to be used as Docker container

intsco:
The README and example Jupyter notebook should provide enough information about how to use the API.

Design choices:
- Falcon as web API framework
- Gunicorn as application server
- Model inference (prediction) - Fastai on top of PyTorch
- Requests handled synchronously with two workers
- No queue is used, so the client has to wait until the model finishes prediction for all images in the request
- Parallel model inference is handled by Fastai/PyTorch


intsco:
@LachlanStuart the review is not urgent and can wait till the middle of the next week

metaspace/off-sample/docker/Dockerfile
```diff
@@ -0,0 +1,102 @@
...
+# config & cleanup
+# ------------------------------------------------------------------
+
+    ldconfig && \
+    apt-get clean && \
+    apt-get autoremove && \
+    rm -rf /var/lib/apt/lists/* /tmp/* ~/*
+
+WORKDIR /app
+COPY . .
```
LachlanStuart: This could be dangerous. If you publish this docker image, you need to make sure there are no passwords/API keys anywhere in this subdirectory, or else they will be included in the image.

I suggest only copying the `models` subdirectory.

metaspace/off-sample/docker/Dockerfile
```diff
@@ -0,0 +1,102 @@
...
+        pyyaml \
+        typing \
+        scikit-learn \
+        && \
+    $PIP_INSTALL \
+#        torch_nightly -f \
+#        https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html \
+        https://download.pytorch.org/whl/cpu/torch-1.0.0-cp36-cp36m-linux_x86_64.whl \
+        torchvision \
+        fastai==1.0.48 \
```
LachlanStuart: In case you ever have to reduce the size of this, the "spaCy" package is 392MB, and it's only required for `fastai.text`. fastai actually has [some documentation](https://docs.fast.ai/install.html) on how to do partial installs, but IMO it'd be easier to just uninstall it after this step. [spaCy is the only library they list in the `fastai.text` dependencies](https://github.com/fastai/fastai/blob/release-1.0.48/setup.py#L115)

metaspace/off-sample/off_sample/model.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class OffSamplePredictModel(object):
+
+    def __init__(self, model_path):
+        self.learn = load_learner('.', model_path)
+        self.target_class_idx = self.learn.data.classes.index('off')
+
+    def predict(self, path):
+        item_list = ImageList.from_folder(path)
```
LachlanStuart: I get the predictions back in an order that doesn't match the input images. I suspect that ImageList.from_folder doesn't do any sorting, and is just returning them in whatever arbitrary order the FS enumerates them in. If you have the full paths to the images available, using `ImageList(list_of_paths)` would preserve the order.

metaspace/off-sample/docker/Dockerfile
```diff
@@ -0,0 +1,102 @@
...
+    apt-get clean && \
+    apt-get autoremove && \
+    rm -rf /var/lib/apt/lists/* /tmp/* ~/*
+
+WORKDIR /app
+COPY . .
+COPY docker/config.ini ./config/config.ini
+COPY docker/start-all.sh ./start-all.sh
+
+EXPOSE 8000
```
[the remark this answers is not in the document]
intsco: I was actually thinking about somewhat opposite solution of using `80` port. Off-sample API is supposed to be the main process running in the container.

metaspace/off-sample/docker/Dockerfile
```diff
@@ -0,0 +1,102 @@
+# ==================================================================
+# module list
+# ------------------------------------------------------------------
+# python        3.6    (apt)
+# pytorch       latest (pip)
+# ==================================================================
+
+FROM ubuntu:16.04
```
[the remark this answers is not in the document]
intsco: Only because the Dockerfile I borrowed these commands from uses it. Will update to 18.04 if it doesn't cause major issues with DL libraries needed.

metaspace/off-sample/off_sample/model.py
intsco: That's a good catch

metaspace/off-sample/docker/Dockerfile
intsco: That's true but I didn't find any best practices on how to copy file you need in the image.
Otherwise I need to have multiple different `COPY` commands for models, configs, code, and scripts.

metaspace/off-sample/docker/Dockerfile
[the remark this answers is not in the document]
LachlanStuart: As the `EXPOSE`'d port inside the container, that's probably fine. I don't think I've ever seen it happen before, but I don't see any reason not to. 

For the external port (i.e. the `-p 8000:8000` in `README.md`), I'd recommend a better default. It becomes a hassle when people have to customize ports, because ports tend to be checked in in config files, etc.

metaspace/off-sample/docker/Dockerfile
LachlanStuart: There's relatively low overhead, so having many `COPY` commands isn't too big of a deal. It also works like `cp` where you can specify many source files and a destination directory e.g. `COPY ./config ./models ./start-all.sh .`

metaspace/off-sample/docker/Dockerfile
```diff
@@ -0,0 +1,102 @@
...
+        && \
+
+# ==================================================================
+# config & cleanup
+# ------------------------------------------------------------------
+
+    ldconfig && \
+    apt-get clean && \
+    apt-get autoremove && \
+    rm -rf /var/lib/apt/lists/* /tmp/* ~/*
```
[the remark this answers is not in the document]
intsco: With Docker 18.09 installed and all empty lines removed from the `RUN` command the warning is gone.

metaspace/off-sample/docker/Dockerfile
intsco: Apparently `COPY` copies only content of directories provided
>Note: The directory itself is not copied, just its contents.

PR #267, opened 2019-03-22
Off-sample API integration

intsco:
* Ion image classification is done after annotation indexing in the update daemon. This leads up to several additional minutes users have to wait before they get access to the annotation results
* Database changes: `ion_image_metrics` table updated, new PK, `off_sample` json field. Migration script provided


metaspace/engine/migrations/r1_5_20190321_iso_image_metrics_update.sql
```diff
@@ -0,0 +1,14 @@
...
+alter table iso_image_metrics
+  add constraint iso_image_metrics_pk
+	primary key (id);
+
+alter table iso_image_metrics
+  add constraint iso_image_metrics_id_pk
+	unique (job_id, db_id, sf, adduct);
+
+alter table iso_image_metrics
+  add off_sample JSON;
```
LachlanStuart: JSON is pretty space-inefficient, which makes a big difference on a table with so many rows. It's probably not worth fixing now, but in future I'd advise against using it on very large tables unless the data truly needs to have an extensible schema. The JSON overhead of just storing this as JSON instead of two columns seems to be about 6.4% of the total table size: 

```sql
WITH row_stats as (
  SELECT id,
         pg_column_size(iim) as total_row_size,
         pg_column_size(iim.off_sample) as json_off_sample_size,
         pg_column_size((iim.off_sample #>> '{prob}')::float) + pg_column_size((iim.off_sample #>> '{label}')::text) as unpacked_off_sample_size
  FROM iso_image_metrics iim
)
SELECT avg(rs.total_row_size), avg(rs.json_off_sample_size), avg(rs.unpacked_off_sample_size),
       (avg(rs.json_off_sample_size) - avg(rs.unpacked_off_sample_size)) * 100.0 / avg(rs.total_row_size) as "potential_savings_pct"
FROM job j
  JOIN iso_image_metrics iim on j.id = iim.job_id
  JOIN row_stats rs on iim.id = rs.id
WHERE iim.off_sample IS NOT NULL;
```
| Average row size | Average `off_sample` column size | Hypothetical size of `prop` and `label` extracted into columns | Potential table size saving |
| -- | -- | -- | -- |
| 465.7 | 44.8 | 14.8 | 6.4% |

intsco: That's a good point to keep in mind

PR #269, opened 2019-03-26
I made res of self.session.post() to be a property of gqclient class.

Ren22:
Added a method to return the status of the response which is a private prop.

### DIFF
--- a/metaspace/python-client/metaspace/sm_annotation_utils.py
+++ b/metaspace/python-client/metaspace/sm_annotation_utils.py
@@ -595,6 +595,7 @@ class SMInstance(object):
         self._config['usr_email'] = email
         self._config['usr_pass'] = password
         self.reconnect()
+        return self._gqclient.res.status_code == 200
 
     def reconnect(self):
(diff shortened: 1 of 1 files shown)


Ren22:
I think it's quite useful to have the response available externally. 
In my particular case, I use it for the spaceM pipe step to access Metaspace and check if a user logged in or not, so a user can grab annotations for private datasetes.
So far I don't understand why returning status from login/reconnect method is more preferable.

intsco:
It's not strictly required but for a good style it would make sense to handle a response in the code that initiates a request. Plus the client doesn't need to call a second method then.
Another case if you're planning to call `login_status` multiple times in different places. But then I'd suggest to have a method like `user_authorized` instead of returning `status_code` which is somewhat specific to `requests` package.

Ren22:
I still don't quite understand it. Could you provide me the snippet with a solution? the only changed I introduced was adding 
```self``` in front of ```res```, otherwise it's not possible to reach it when the object of class is instantiated.

PR #271, opened 2019-04-04
Post Off-sample integration fixes

intsco:
* A notebook for exploring off sample predictions. This PR should be updated after PR #272 is merged so that in the notebook off sample data is taken from `GraphQL` rather than from the DB.
* Removed `off_sample` module from `python-client` since it relies on the engine code anyway
* `python-client`: new methods and fixes


intsco:
@LachlanStuart here is the new notebook if you wanted to take a look before I merge the PR
https://github.com/metaspace2020/metaspace/blob/be3ba6b83435700ee8ecf65fb08112c6dfa454b3/metaspace/python-client/example/explore_off_sample_results.ipynb

LachlanStuart:
@intsco The new changes look good to me

PR #272, opened 2019-04-04
Off-sample front-end integration

LachlanStuart:
@intsco Please review engine & graphql changes

@Ren22 If you have time, please review the webapp changes. If not, don't worry about it.

In `engine`:

* Adds off-sample fields to ElasticSearch
* Changes update-daemon to do off-sample analysis in two daemon actions: one high-priority `INDEX`, and one low-priority `CLASSIFY_OFF_SAMPLE` (which also indexes afterwards)
* Updates `scripts.run_off_sample` to also reindex the dataset
* Added a test fixture to `test_sm_daemons` that resets the test queues before the tests. This was needed because `run_daemons` would only cause 1 message from each queue to be processed, and annotation jobs now produce 2 messages - `INDEX` and `CLASSIFY_OFF_SAMPLE`. This seemed to cause the second test to not process its intended annotation message.

In `graphql`:

* Adds API for viewing, filtering & sorting annotations by off-sample label/prob
* Ignores unrecognized status queue messages so that `CLASSIFY_OFF_SAMPLE` doesn't log errors

In `webapp`:

* Adds UI for viewing, filtering & sorting by the off-sample label
* Adds 3 feature flags:
    * `off_sample` for hiding/showing the off-sample filter & diagnostics indicator
    * `off_sample_col` that shows the off-sample `prob` in the annotations table so that we can easily it. We don't have much space in that table, so I don't want it permanently visible. It can be temporarily activated by adding `?feat=off_sample_col` to the URL
    * `new_feature_popup` that enables the "New Feature" popups. If this and `off_sample` are activated, the popup should appear when you first visit the Annotations page. If you dismiss it and want to see it again, run this in the browser console: `localStorage.removeItem('dismissedFeaturePopups')`

### DIFF
--- a/ansible/aws/templates/all.yml.template
+++ b/ansible/aws/templates/all.yml.template
@@ -191,8 +191,11 @@ sm_webapp_enable_uploads: true
 # "specialised metadata" subfolder
 sm_webapp_metadata_types: ["ims"]  # set ["ims", "lcms"] to accept both IMS and LC-MS datasets
 sm_webapp_features:
-  coloc: false
+  coloc: true
   ion_thumbs: false
+  off_sample: true
+  off_sample_col: false
+  new_feature_popups: true
 
 sm_activate_venv: source {{ miniconda_prefix }}/bin/activate {{ miniconda_env.name }}
 
--- a/metaspace/engine/scripts/run_off_sample.py
+++ b/metaspace/engine/scripts/run_off_sample.py
@@ -20,7 +20,7 @@ ORDER BY j.ds_id DESC;
 """
 
 
-def run_off_sample(ds_id, sql_where, fix_missing):
+def run_off_sample(ds_id, sql_where, fix_missing, overwrite_existing):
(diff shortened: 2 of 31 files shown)


intsco:
image
What about <10%, >90% options for extreme values of off sample probability? Or they are not supposed to be shown to users?

LachlanStuart:
@intsco That table column is not supposed to be visible to users, I just added it for debugging.

However, it also appears as a raw value in the CSV. I'm not sure what to do about that - I don't want to introduce string values that would mess up parsing to floating point. Do you think it'd be fine if I just added a note to the [CSV documentation](https://github.com/metaspace2020/metaspace/blob/master/metaspace/webapp/docs/csv_export.md)?

intsco:
@LachlanStuart adding a note to the CSV export documentation sounds like the best solution

theodev:
Maybe adding it as a comment in the CSV itself?

PR #279, opened 2019-04-15
Add 'Image_Generation' -> 'ppm' field to metadata. Other minor fixes

intsco:
* This way `ppm` can be specified in the dataset metadata in GraphQL
* Ansible metaspace_repo role: local deployment for engine and graphql
* Supervisor doesn't autorestart `sm-update-daemon`. `QueueConsumer` exists after 5 unsuccessful attempts to poll a message
* Updated timeout and number of threads for off-sample wrapper
* Use the same port inside and outside the off-sample container


metaspace/off-sample/docker/Dockerfile
```diff
@@ -97,7 +97,7 @@ COPY off_sample ./off_sample
 COPY app.py docker/start-all.sh ./
 COPY config/default.ini ./config/config.ini
 
-EXPOSE 8000
+EXPOSE 9876
```
LachlanStuart: Thank you! I was planning to make a PR to do this myself. Docker can't remap ports inside the docker network, so this was on port 8000 when I ran sm-engine inside docker, and 9876 when I ran it outside docker.

ansible/roles/sm_update_daemon/templates/sm-update-daemon.supervisor.j2
```diff
@@ -9,4 +9,4 @@ redirect_stderr = true
 stdout_logfile = {{ sm_home }}/logs/{{ sm_update_daemon_app_name }}.log
 numprocs = 1
 numprocs_start = 8000
-autorestart = true
\ No newline at end of file
+autorestart = false
```
LachlanStuart: What's the reasoning for this and the failed attempts limit? I'd prefer it if the system always tried to recover, or at least tried to reconnect for several minutes instead of just 5 seconds.

metaspace/off-sample/docker/Dockerfile
intsco: That's a pretty logical solution anyway

ansible/roles/sm_update_daemon/templates/sm-update-daemon.supervisor.j2
intsco: I decided to add it after I had to delete 700MB of logs of `sm-updated-daemon` trying to reach RabbitMQ. So far in my experience RabbitMQ was always either up and available or down and some manual interventions was needed. If you want, we can add longer waiting time between reconnect attempts.

PR #283, opened 2019-04-23
Separate segmentation of dataset and centroids dataframe

intsco:
This PR implements a new datasets and centroids segmentation strategy that should allow to scale dataset processing more easily by increasing the number of centroids segments beyond 128.

* Dataset is segmented into small pieces of 5MB in size. Centroids dataframe is segmented into 32 or more pieces. Each centroids segment takes multiple dataset segments when doing molecule search
* Runtime of dataset and centroids dataframe segmentation algorithm has been greatly improved
* `search_results` module generates png and sends them to the image service using Spark
* A list of formulas for a molecular db isn't stored in the database anymore


metaspace/engine/sm/engine/search_job.py
```diff
@@ -85,34 +84,38 @@ def store_job_meta(self, mol_db_id):
...
+            logger.info('Parsing imzml')
+            imzml_parser = ImzMLParser(self._ds_imzml_path)
             search_alg = MSMSearch(sc=self._sc, imzml_parser=imzml_parser, moldbs=moldbs,
                                    ds_config=self._ds.config, ds_data_path=self._ds_data_path)
             search_results_it = search_alg.search()
 
-            for moldb, moldb_ion_metrics_df, moldb_ion_images in search_results_it:
+            for moldb, moldb_ion_metrics_df, moldb_ion_images_rdd in search_results_it:
                 # Save results for each moldb
                 self.store_job_meta(moldb.id)
```
LachlanStuart: I should have caught this in the previous PR: `self.store_job_meta` sets the `job.start` time. Moving this call to after `MSMSearch.search` means that `job.finish - job.start` can no longer be used to see how long the job took. I use those stats on occasion for benchmarking and finding problematic datasets.

Could this be changed so that `start` is based on a timestamp taken before `MSMSearch.search` is called? No need to compensate for the fact that there will be multiple jobs overlapping - if you use the exact same value for `start` across multiple jobs, I can group them together in SQL and take the average duration.

metaspace/engine/sm/engine/search_job.py
```diff
@@ -85,34 +84,38 @@ def store_job_meta(self, mol_db_id):
...
                 coordinates = [coo[:2] for coo in imzml_parser.coordinates]
                 sample_area_mask = make_sample_area_mask(coordinates)
-                search_results.store(moldb_ion_metrics_df, moldb_ion_images, sample_area_mask,
+                search_results.store(moldb_ion_metrics_df, moldb_ion_images_rdd, sample_area_mask,
                                      self._db, self._img_store, img_store_type)
                 self._db.alter(JOB_UPD_STATUS_FINISH, params=(JobStatus.FINISHED,
                                                               datetime.now().strftime('%Y-%m-%d %H:%M:%S'),
                                                               self._job_id))
 
                 if self._sm_config['colocalization'].get('enabled', False):
```
LachlanStuart: This should be un-indented one level - `run_coloc_job_for_new_ds` runs against all of a DS's DBs every time, so it's only necessary to call it once.

metaspace/engine/sm/engine/search_job.py
intsco: Yes, that's a good comment. I didn't know what to do with what we used to call a job, since now jobs are executed together in the search step. Won't it be confusing to have start and finish time associated with particular dataset and database affected by all other databases the dataset was processed against?

intsco:
@LachlanStuart There are multiple issues that I'm fixing in the following PR

metaspace/engine/sm/engine/search_job.py
LachlanStuart: If the data is in the database, at least I can work around it. I generally don't need to know DB-specific stats, only average duration per DB. If the DB doesn't have correct start/end times, then I would have to dig through the logs.

intsco: Ok, will add it to the next PR

LachlanStuart:
Interesting - GitHub added a (poorly labelled) button to re-request review. The "Refresh" icon here:
image

Issue #281, opened 2019-04-18
Feature request: stop scaling optical images


LachlanStuart:
Due to interdependencies between features, this will actually be implemented as 3 parts, in this order:

- [x] Indicating whether hotspot removal has been applied to ion images
- [x] Removing optical image cropping, instead warping the optical image/ion image in the annotation display so that the whole optical image is visible
    - [x] As part of these changes, add an error handler so that if matrix inversion fails due to an error thrown by `numeric`, reset the transformation
- [x] Adding a display field that shows the intensity for the pixel under the cursor

PR #290, opened 2019-05-03
Refactor ion image loading & add info about hotspot removal to UI

LachlanStuart:
This refactors/rewrites most of the ion image-related code to rely less on the browser for image decoding & size calculation. PNG decoding is now done with the `upng-js` library, our code for intensity range compression & colormapping has been optimized, and the `ImageLoader` component has been split into two parts: The image loading & layout calculation have been extracted and put in `ImageLoader` and `MainImage`, and all the remaining code for showing the image & allowing dragging/zooming is now in `IonImageViewer`.

**Pros of doing all this stuff manually:**
* 16-bit PNGs are now properly decoded, which is going to be helpful for the task in #281 to show more accurate intensity values of specific pixels
* Making changes to the layout is easier - you don't have to dig through multiple layers of CSS & worry about breaking other places that use the same styles
* The coordinate system is now anchored in predictable units. E.g. `zoom=2.0` now means `1 ion image pixel == 2.0 browser pixels`, instead of `image width == 2.0 * whatever width the browser decided to give the image`. This is vital for the other parts of #281 , as it makes it possible to work backwards and determine exactly which ion image pixel corresponds to a given screen pixel.

**Cons:**
* `upng-js` + `pako` are an extra 20kB(gzipped) of JS needed to show the annotations page
* While image processing is now much faster (~1ms for doing hotspot removal and & colormapping a 350x100px image), decoding the PNG is slower as it no longer happens in a background thread. 
* There's more code

### Changes:
* Refactored/rewrote most ion image handling code & zoom/move math in `ImageLoader`
* Added tooltip about hotspot removal to the intensity scale
* Extracted `ScaleBar` into a separate component from `ImageLoader`
* Fixed bugs in `ImageAligner` that made it possible to irreversibly break the transform matrix with numerical errors
* Fixed css issue in `DatasetList` which caused IE11 to always show a single column of datasets (the bug is visible on Staging but it thankfully hasn't been deployed to prod)
* dom-to-image now captures the full image area instead of shrinking to only show the ion image. This means the scale bar is now always included.

### Testing:
As this change uses relatively new features that are quite browser sensitive, I've run through the following in IE11/Edge/Firefox/Chrome:
* Pan & zoom around ion images in the annotations page, with the Colocalized Annotations tab open (which contains lots of smaller ion images)
* Toggle the optical image & change the opacity
* Use the optical image alignment page to rotate, zoom, pan & warp the ion image

I've also checked that the optical image transforms and opacity mapping are the same before & after.


intsco:
Looks good to me. At least when testing locally, I couldn't spot any issues

PR #299, opened 2019-05-10
Transform optical image in webapp

LachlanStuart:
Part 2 of #281 

# Notable changes:
* Adds a `optical_transform` feature flag to webapp to toggle between the old and new mode (Can be quickly toggled on by loading the page with `?feat=optical_transform` in the URL)
* Updates the thumbnail generation script to reapply all optical image transformations instead of just adding thumbnails. Add command line options to make it easier to target specific datasets.
* Simplifies the `acq_geometry` code and uses metadata & ion images instead of imzML for pixel size/dimensions. I didn't end up using this for image rendering, but this will be really useful for me anyway, as I'll be able to query for large ion images via SQL
* Adds 3 new migration steps:
    * `r1_5_20190507_optical_image_types.sql` for the SQL schema changes
        * New column `type` for distinguishing types of optical image: `scaled` is the new type, `clipped_to_ion_image` is the old type
        * New column `transform` that stores the matrix needed to map the optical image to ion images. Note that the transform dependent on both zoom and the ratio of ion image pixel density to optical image pixel density, so it is different for every `optical_image` row.
    * `python -m migrations.r1_5_20190507_regenerate_acq_geometry --migrate` to regenerate `acq_geometry`
    * `python -m scripts.update_optical_images --sql-where "transform IS NOT NULL"` to regenerate the optical images
* Moves `add_optical_images` and `del_optical_images` out of `DatasetManager` into their own file. Also moves deletion of the optical images into update-daemon, where it was originally intended to go.
* Adds limits to how much an optical image can be scaled, both on client side and during optical image generation. 
    * In webapp it's now possible to zoom further out, and the zooming speed has been changed to not get slower as you zoom more
    * During image generation it previously could upscale the optical image at higher zoom levels, producing images as big as 8000x4000 even for small optical images, now it is capped at the maximum size of the optical image.
* Added support for PNG optical images

# Shortcomings

* The initial zoom in the annotations page is fitted to the ion image size, not the (transformed) optical image size. This means that the user will often need to zoom out to see the whole optical image.
* There's a bug that I haven't been able to track down - if you scroll too fast then the ion image stops zooming and the browser starts zooming instead. Really annoying. I don't think it's a new bug though.
* There is some duplicate code between `ImageLoader.vue` and `MainImage.vue`

# Main areas to review

* Do the SQL and GraphQL schema changes make sense
* Have `acq_geometry.py` and `optical_image.py` been refactored in an acceptable way
* Do the migrations work


metaspace/engine/sm/engine/optical_image.py
```diff
@@ -0,0 +1,165 @@
...
+    max_scale_factor = np.ceil(max(scan.width / dims[0], scan.height / dims[1]))
+    scale_factor = zoom * min(VIEWPORT_WIDTH / dims[0], VIEWPORT_HEIGHT / dims[1])
+    scale_factor = np.clip(scale_factor, 1, max_scale_factor)
+
+    transform = np.array(transform_)
+    assert transform.shape == (3, 3)
+    transform = transform / transform[2, 2]
+    transform[:, :2] /= scale_factor
+    coeffs = transform.flat[:8]
+    new_dims = int(round(dims[0] * scale_factor)), int(round(dims[1] * scale_factor))
```
LachlanStuart: Note: Previously `scale_factor` was rounded to an integer before multiplying against `dims`. Now the `new_dims` are rounded after scaling. This means that the generated images are no longer even multiples of the ion image size.

metaspace/webapp/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "allowJs": true,
-    "target": "es6",
+    "target": "es5",
```
LachlanStuart: This is an unrelated bugfix - the current code in `master` doesn't work in IE11 because *something* in the webpack pipeline is using this file (which is intended for tests & WebStorm code assistance) instead of the `tsconfig.client.json` file, despite the config file path being overridden in all possible places.

The downside of this is that some of the Jest tests no longer pass type checking, but this doesn't actually cause them to fail - WebStorm is the only thing that type-checks the `.spec.ts` files...

metaspace/engine/sm/engine/acq_geometry.py
```diff
@@ -0,0 +1,53 @@
...
+    #     grid_props = parser.imzmldict
+    # else:
+    #     grid_props = {}
+
+    return {
+        'length_unit': 'nm',
+        'acquisition_grid': {
+            'regular_grid': True,
+            'count_x': int(dims[0]),
+            'count_y': int(dims[1])
```
intsco:
```
row_n, col_n = dims
count_x = col_n
count_y = row_n
```
right?

metaspace/engine/scripts/create_schema.sql
```diff
@@ -33,8 +33,12 @@ CREATE INDEX ind_dataset_name ON dataset (name);
...
 CREATE TABLE optical_image (
   id      text PRIMARY KEY,
-  ds_id   text,
-  zoom    int,
+  ds_id   text NOT NULL,
+  type text NOT NULL,
+  zoom REAL NOT NULL,
+  width INT NOT NULL,
+  height INT NOT NULL,
+  transform REAL[][],
```
intsco: And `transform` field in the `dataset` table, it's not used anymore?

intsco:
Regarding the migration scripts. I'd suggest to test them on staging next week.

metaspace/engine/scripts/create_schema.sql
LachlanStuart: The `transform` matrices operate on pixel coordinates, which means that if the optical image is resized, then the `transform` matrix needs to be scaled appropriately. We used to handle this with a bunch of math on the client side, but now this `optical_image.transform` field contains the precalculated result of that math for the specific resolution of this image.

For the old images, now known as `type='clipped_to_ion_image'` images, the `transform` only does scaling, because the image is pre-transformed to have a direct mapping to ion image pixels.

For the new images, the `transform` is actually the inverse matrix from what is configured on the Upload Optical Image page. The Upload Optical Image page uses an `Ion Image->Optical Image` transformation matrix, but the annotation page needs an `Optical Image->Ion Image` transformation matrix.

metaspace/engine/sm/engine/acq_geometry.py
LachlanStuart: That's a bit annoying. Having coordinates/sizes in `x,y` / `w,h` order is almost universal in image processing, even if iteration/storage happens in the other order.

Fix pushed.

PR #301, opened 2019-05-13
Show ion intensity under cursor

### LINKED ISSUE #281: Feature request: stop scaling optical images

LachlanStuart:
Resolves #281 

image

Includes three unrelated minor fixes:
* Correctly calculate the change of offset when zooming in/out of ion images
* Fix an issue in the width of DatasetList which made Firefox always show it as only one column
* Fix the component name of `MainImageHeader`. This only showed up in the dev tools, but it was confusing.

### DIFF
--- a/metaspace/webapp/src/components/IonImageViewer.vue
+++ b/metaspace/webapp/src/components/IonImageViewer.vue
@@ -1,16 +1,20 @@
 <template>
   <div class="image-loader"
        v-loading="isLoading"
+       ref="imageLoader"
        :element-loading-text="message"
        :style="{width:width+'px', height:height+'px'}"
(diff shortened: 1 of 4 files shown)


intsco:
Btw, in Chromium the datasets page also has only one column

PR #339, opened 2019-06-07
Production hotfixes

LachlanStuart:
Several small fixes - see individual commits

The `jwt-simple` vulnerability *probably* doesn't affect us. Looking at the code it appears as though the issue in `jwt-simple` was [this issue](https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/): when an application uses the RSA public/private key algorithm for generating JWTs, a user can use the RSA public key to generate a valid signature for an HMAC/SHA-based JWT. I.e. if a developer publishes a public key used for JWTs, `jwt-simple` could be tricked into treating that public key as a `secret`. We don't use RSA JWTs and we've never published our `secret`, so we shouldn't be affected.

Since it's a security-critical dependency, I updated it as a precaution and also followed Auth0's advice on explicitly specifying the `algorithm` so that it couldn't be changed to a weaker algorithm in the JWT.

I also added Sentry error reporting to GraphQL. There will be a matching PR in the ansible repo to update some of the settings.

I will probably merge & deploy this hotfix on Tuesday once I'm happy it's stable, but a code review would still be appreciated when you get back.


intsco:
@LachlanStuart all looks good!

PR #344, opened 2019-06-14
Add emails for group requests/invitations & clean up graphql

LachlanStuart:
This adapts the "(user) invited you to a project", "(user) requested access to a project" and "Your project application was accepted" emails so that they can be used with both groups and projects. Going forward I'm going to try to keep consolidating code from the two and adding it to this `groupOrProject` module to reduce repetition.

Sorry for including unrelated changes in this PR. While updating the tests I found `imageUpload.test.js` had been failing, so I fixed it, and fell down a rabbit hole cleaning up the JS code & converting it to TS.


metaspace/webapp/src/router.ts
```diff
@@ -56,7 +56,11 @@ const router = new VueRouter({
     { path: '/admin/groups', component: async () => await import('./modules/Admin/GroupsListPage.vue') },
 
     { path: '/account/sign-in', component: DialogPage, props: {dialog: 'signIn'} },
-    { path: '/account/sign-in-success', redirect: redirectAfterSignIn },
+    { path: '/account/sign-in-success', redirect: () => {
+        debugger;
```
intsco: Is it a forgotten debug line?

LachlanStuart: Yes, good catch

PR #345, opened 2019-06-14
Ion image options improvements

### LINKED ISSUE #340: Log-scale colormaps

LachlanStuart:
Resolves #340 

This PR cleans up the ion image options panel and adds options for log-scale and rank-scale colormapping. 

Peek 2019-06-14 15-37

Full changes:
* Added `log` and `rank` ion image quantization modes
* Replaced `hotspotThreshold` querystring parameter with `scaleType`, which can be one of several presets that is used to determine the scale mode (linear/log/rank) and upper/lower percentiles for clipping (outlier clipping / full range)
* Simplified `IonImageSettings` component by moving `Inverted` into the colormap options dropdown and moving "hidden" into the scale bar color options. Also remove some redundant parameters (`colormapName` and `showScaleBar`) as they can be nicely encoded in their related parameters (`colormap` and `scaleBarColor`)
* Added `dontShowAfter` parameter to `NewFeaturePopup` configs so that we don't need to worry about manually turning the notices off. I've set the off-sample popup to stop appearing 2 months after we turned the feature on in production.
* Changed `ColorBar` to show an accurate scale in log/rank modes. Also support displaying inverted colormaps in the `IonImageSettings` dropdown (required a change to reverse array order in `getColorScale`, because CSS gradients must be in ascending order)
* Unrelated: fixed the z-order of `MetaspaceHeader` so that loading overlays don't appear over the top of the header bar.

### DIFF
--- a/metaspace/webapp/src/components/ImageLoader.vue
+++ b/metaspace/webapp/src/components/ImageLoader.vue
@@ -23,7 +23,7 @@
   import config from '../config';
   import {Component, Prop, Watch} from 'vue-property-decorator';
   import IonImageViewer from './IonImageViewer.vue';
-  import {IonImage, loadPngFromUrl, processIonImage} from '../lib/ionImageRendering';
+  import {IonImage, loadPngFromUrl, processIonImage, ScaleType} from '../lib/ionImageRendering';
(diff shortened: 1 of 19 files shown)


LachlanStuart:
Change needed before merging: Rename "Rank" scale to "Equalized Histogram"

PR #353, opened 2019-06-26
Move list of deprecated and custom mol DBs into config

intsco:
Once the list of deprecated mol DBs is loaded from the config, it's easier to update it


intsco:
Corresponding config updates https://github.com/intsco/metaspace-ansible-config/pull/4

Issue #362, opened 2019-07-01
Discuss standardization of adduct format


LachlanStuart:
With the change to add `[M]⁺`/`[M]¯` adducts I already found that our format of just storing e.g. `+H` introduces problems, as it would mean that both charge-only adducts would be stored ambiguously, as an empty string. As a workaround I added special-case checks for `"[M]+"`/`"[M]-"` strings, because it was less messy than trying to differentiate between empty strings based on polarity.

Skyline uses a [relatively consistent and comprehensive format](https://skyline.ms/wiki/home/software/Skyline/page.view?name=adduct_descriptions) for storing adducts.  It would be a good basis if we move to that format or something similar. The only apprehension I have is around the way that Skyline's format indicates charge - it seems to prefer to include it inside the brackets for some reason e.g. `[M+2]` even though that way is ambiguous with normal notation for mass shift, and it supports parsing the charge from outside the brackets e.g. `[M]2+`

If we intend to support multiple charges, isotopic labelling or unknown mass shifts, we should consider migrating to a system similar to Skyline's.

LachlanStuart:
I'm moving this to the backlog. I don't think there's short-term benefit anymore, so no point in keeping it in the triage list.

LachlanStuart:
Closing this as there is now no champion for it

PR #363, opened 2019-07-08
[M]+/[M]- support, bug fixes & ES scripts

LachlanStuart:
This contains several somewhat interrelated sets of changes:
* [M]+/[M]- support (Feature-flagged UI & back-end changes required to support the `"[M]+"`/`"[M]-"` strings. I've added some thoughts about adduct strings here: #362 362
* Bug fixes. There were several issues involving my use of a single string for "ion", which meant that if one dataset used a `chem_mod` that also happened to look like a `neutral_loss`, `adduct` or combination, then various bad things would happen causing the unique constraint to be violated. Now more places look ions up by their full `(formula, chem_mod, neutral_loss, adduct)` tuple. I'm not happy with this solution, because it isn't going to scale well, but I need to move on or else I'll never finish this.
* Updates to the ES scripts to allow `--inactive` to be used to create/update the `sm-yin`/`sm-yang` that isn't currently being used. I didn't add a second alias as I once said I would - it turned out to be cleaner to just have 1 alias for the active index and figure out which was the inactive index based on that.
* A small optimization to `update_es_index` - caching calculated centroids in `isocalc_wrapper`. I saw a ~25% performance increase when re-indexing 100 datasets, but I think it will be bigger overall on production, because the cache will be warmer on average with a bigger sample size, and because staging/production are using the slower/older version of cpyMspec whereas I'm locally running the latest version. Currently re-benchmarking against 500 datasets - will update when it's done.
    Note: this centroids cache is only enabled for `update_es_index`. I think it will be too much of a memory leak for production. We'll see once I reindex staging.


LachlanStuart:
Performance comparison for ES export of 500 datasets:

Without centroids cache:
```
	Command being timed: "python -m scripts.update_es_index --ds-name %' AND status = 'FINISHED' ORDER BY id DESC LIMIT 500; --"
	User time (seconds): 1013.14
	System time (seconds): 4.22
	Percent of CPU this job got: 58%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 28:58.47
	Maximum resident set size (kbytes): 562272
	Minor (reclaiming a frame) page faults: 285194
	Voluntary context switches: 182278
	Involuntary context switches: 63354
```

With cache:
```
	Command being timed: "python -m scripts.update_es_index --ds-name %' AND status = 'FINISHED' ORDER BY id DESC LIMIT 500; --"
	User time (seconds): 260.05
	System time (seconds): 3.76
	Percent of CPU this job got: 27%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 16:11.23
	Maximum resident set size (kbytes): 565232
	Minor (reclaiming a frame) page faults: 277238
	Voluntary context switches: 182655
	Involuntary context switches: 39011
```

The wall clock time (i.e. including ElasticSearch latency) only decreased 44% (28:58.47 -> 16:11.23). The CPU time used by Python decreased 74%, which suggests that multi-threading may be viable now because that gap between 74% and 44% means that Python is now spending a lot more time waiting for IO.

LachlanStuart:
Also worth noting, because I could no longer rely on the `ion` column to be unique, I had to change `chem_mod` and `neutral_loss` to be non-nullable. The reason is that in postgres, `null != null`, and this means that unique indices are basically useless, because e.g. you could have several copies of `('H2O',null,null,'+H')` without postgres considering them to be conflicts.

LachlanStuart:
> which suggests that multi-threading may be viable now because that gap between 74% and 44% means that Python is now spending a lot more time waiting for IO.

I tried reindexing 100 datasets with 1, 2 and 4 threads and got 3m09s, 2m23s and 2m10s. Multi-threading seems to make a bigger difference now, but it's still unlikely to translate to a big difference on the 2-core machines we have in production, so I won't go through with it.

PR #373, opened 2019-07-23
Bug fixes & tweaks

LachlanStuart:
Best reviewed commit-by-commit.

The first two commits are for handling changes in sm-engine behavior. I'm not sure what the cause for these changes is, and I couldn't actually reproduce the "completely black ion images" locally. Handling it on the client side seemed to be the easiest and most future-proof fix, though ideally we wouldn't generate empty PNG files.

Example DS with black images: https://metaspace2020.eu/annotations?db=ChEBI-2018-01&ds=2019-06-25_22h19m11s&cmap=Hot&scale=hist&sections=5 (Look in the Diagnostics tab)

"Prevent logging 'false' error" and "Ignore promise rejections when $alerts are closed" fix benign bugs.  ElementUI rejects promises with `false` when a form is invalid and `'cancel'` when a popup is closed by its "X" button. Since neither value inherits from `Error`, they don't get a stack trace which makes them a huge pain in Sentry to debug.

I'm really not proud of the "Make candidate molecule popups close faster" hack, but it makes the annotations page feel noticeably better. I submitted a PR to ElementUI to add an option that would allow the hack to be removed.

### DIFF
--- a/ansible/aws/templates/all/vars.yml.template
+++ b/ansible/aws/templates/all/vars.yml.template
@@ -199,6 +199,7 @@ sm_webapp_features:
   all_dbs: false
   all_adducts: false
   neutral_losses: false
+  neutral_losses_new_ds: true
   chem_mods: false
   advanced_ds_config: false
 
--- a/metaspace/engine/requirements.txt
+++ b/metaspace/engine/requirements.txt
@@ -8,7 +8,7 @@ pytest
 recommonmark
 boto3
 fabric3
-pypng
+pypng==0.0.19    # 0.0.20 introduced incompatible API changes
 pyyaml
 elasticsearch-dsl>=5.0.0,<6.0.0
 pika==0.11.0b1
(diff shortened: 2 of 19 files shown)


ansible/aws/templates/all/vars.yml.template
```diff
@@ -199,6 +199,7 @@ sm_webapp_features:
   all_dbs: false
   all_adducts: false
   neutral_losses: false
+  neutral_losses_new_ds: true
```
intsco: This feature flag should probably be added to the Ansible configs https://github.com/intsco/metaspace-ansible-config too

LachlanStuart: [I added it](https://github.com/intsco/metaspace-ansible-config/commit/06a56ce6ab588798606a777e1fefb601e4964329), it just didn't seem to be enough to warrant making a PR for

PR #390, opened 2019-08-09
Fix typo in neutral loss help text

LachlanStuart:
Also completely hide the section if the feature flag is enabled to allow people to set neutral losses on the initial upload.


intsco:
@LachlanStuart Two line PR? )

PR #391, opened 2019-08-09
Reorganize `.graphql` files

LachlanStuart:
No functionality/API changes, just cleaning up. 

I'm about to add new functionality to the annotation graphql schema and decided it was time to clean up the remaining declarations from the left-over `schema.graphql` file.


intsco:
@LachlanStuart probably `.circleci/config.yml` needs to be updated to fix `test-webapp` job

LachlanStuart:
@intsco Thanks, I hadn't noticed CircleCI failed. I had a suspicion that there was a 2nd place I had to update, but couldn't find anything with Find in Files. 

I've fixed the issue and added comments linking the two places that should be updated at the same time.

PR #393, opened 2019-08-13
Use database connection pool in sm.engine.db

intsco:
The only real change in this PR is use of database connection pool in `sm.engine.db` module. This module opens and keeps a singleton object that holds a pool of connection to the database. The code of `db_decor` and `DB` was also refactored to better handle transactions and exceptions.

All the other code changes were caused by the updated DB interface.

### DIFF
--- a/metaspace/engine/scripts/run_colocalization.py
+++ b/metaspace/engine/scripts/run_colocalization.py
@@ -74,7 +74,7 @@ def run_coloc_jobs(ds_id, sql_where, fix_missing, fix_corrupt, skip_existing):
 
     conf = SMConfig.get_conf()
 
-    db = DB(conf['db'])
+    db = DB()
 
     if ds_id:
         ds_ids = ds_id.split(',')
--- a/metaspace/engine/scripts/run_colocalization.py
+++ b/metaspace/engine/scripts/run_colocalization.py
@@ -1,9 +1,8 @@
 import argparse
-import logging
-from itertools import groupby
+from functools import partial
 
 from sm.engine.mol_db import MolDBServiceWrapper
-from sm.engine.util import init_loggers, SMConfig
+from sm.engine.util import bootstrap_and_run
 from sm.engine.db import DB
 from sm.engine.colocalization import Colocalization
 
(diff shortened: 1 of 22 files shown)


intsco:
@LachlanStuart did you mean something along those lines?

metaspace/engine/sm/engine/db.py
```diff
@@ -49,15 +48,24 @@ class DB(object):
...
 
     def __init__(self, config, autocommit=False):
-        self.conn = psycopg2.connect(**config)
-        if autocommit:
-            self.conn.set_isolation_level(psycopg2.extensions.ISOLATION_LEVEL_AUTOCOMMIT)
+        self._config = config
+        self._autocommit = autocommit
+        self.conn = None
         self.curs = None
         DB._dbs.add(self)
```
LachlanStuart: Because `_close` can now be called multiple times during a `DB`'s lifespan without invalidating the `DB` instalce, `DB._dbs.add(self)` should be moved into `_connect` so that it re-registers itself when it re-connects.

metaspace/engine/scripts/run_colocalization.py
```diff
@@ -74,7 +74,7 @@ def run_coloc_jobs(ds_id, sql_where, fix_missing, fix_corrupt, skip_existing):
 
     conf = SMConfig.get_conf()
 
-    db = DB(conf['db'])
+    db = DB()
```
LachlanStuart: I think these `scripts` also need to call `init_conn_pool`...

metaspace/engine/sm/engine/db.py
```diff
@@ -1,74 +1,69 @@
...
             logger.debug(debug_output[:1000])
-            res = func(self, *args, **kwargs)
-        except Exception as e:
-            # logger.error('SQL: %s,\nArgs: %s', args[0], str(args[1:])[:1000], exc_info=False)
-            self.conn.rollback()
-            raise Exception('SQL: {},\nArgs: {}\nError: {}'.format(args[0], str(args[1:])[:1000], e.args[0]))
-        else:
-            self.conn.commit()
+
+            with _conn_pool.getconn() as conn:
```
LachlanStuart: As far as I can tell from the docs/source, `_conn_pool.getconn()` returns a plain `connection` class, which means using this in a `with` statement always closes the `connection` directly upon leaving the `with` block. This `with` should be removed so that `conn` can be be reused by the pool.

intsco:
@LachlanStuart please review changes only in this commit https://github.com/metaspace2020/metaspace/pull/393/commits/2cac4f4f34697814fd4da3d926a9ca5dee505785

PR #394, opened 2019-08-15
Refactor sm_daemons module

intsco:
Refactored `sm_daemons` module by pulling operations on datasets from the daemon classes into `DatasetManager` class


LachlanStuart:
I've double-checked and can confirm `status` isn't used in the graphql/webapp side. It was previously used to detect when ES finished refreshing, but I removed that logic when I changed `ESExporter` to block until ES has confirmed the index has refreshed.

intsco:
@LachlanStuart thanks for checking that!

PR #397, opened 2019-08-23
Python client changes

LachlanStuart:
* Stops python-client from attempting to log in (and failing) before you call `SMInstance.login`. Also gets rid of the annoying `print`s when everything is working correctly
* Adds `only_first_isotope` and `scale_intensity` parameters to `SMDataset.all_annotation_images`, and adds an example of use to one of the example notebooks
* Removes the ElasticSearch-dependent `Top hits.ipynb` and add note discouraging use of ElasticSearch-dependent functions, as they won't ever work
* Adds a default annotation filter to exclude neutral losses, etc. as python-client has a `(sf, adduct)`-centric API
* Bumps python-client version to `1.5.1`. I had previously accidentally created and deleted `1.5.0`, so that version number is no longer available. I haven't pushed this version yet - I'll only do so after review.
* Removes unnecessary / not-often-needed dependencies


LachlanStuart:
@aeisenbarth As your code depends on `python-client`, could you please also review this?

PR #398, opened 2019-08-27
Code formatter and linter setup

intsco:
Install all necessary libraries into `sm-engine` docker image. Update circleci config. Quick fix of the sci test.


metaspace/engine/pylama.ini
```diff
@@ -0,0 +1,25 @@
+[pylama]
+format = parseable
+skip = */.tox/*,*/.env/*
+linters = pylint,pydocstyle,pyflakes,mccabe
+ignore = F0401,C0111,E731,E203,D100,D103,D104,D213,W1203
+
+[pylama:pylint]
+max_line_length = 88
```
LachlanStuart: Most of our code seems to target a 120 char line length...

intsco: I think it's because it's a PyCharm's default. But 88 seems to be a default value for most linters and Black as well. It's still an open question which line length to choose

PR #400, opened 2019-08-27
Black formatted code

intsco:
The code was formatted with the following command:
`black --line-length 88 --skip-string-normalization --target-version py36 sm tests scripts`
`--line-length 88` is Black's default


intsco:
Please take a quick a look what the new code style looks like.
For comparing the code formatting with 88 vs 120 line length check this commit 5658bbf473c49f34922637ca0f65c66de270ccc3

LachlanStuart:
@intsco It's ultimately your choice to make, but I strongly prefer 120 chars. I've tried to follow an 80 col limit in the past, and found that it slowed me down and bloated the code by requiring extra temporary variables, defs instead of lambdas, etc. to keep the code clean, without actually improving readability. Also, keep in mind that by limiting horizontal space, code will take more space vertically, which can be even worse on a smaller screen. 

Otherwise, I'm happy with Black's code style. It's not perfect in all situations, but at least it will be one less thing to need to think about.

intsco:
Okay, since it's ultimately my choice I'd like to try an intermediate solution of 100 chars. Quite subjectively, I feel like most of the time lines longer than that are slightly less comfortable to read.

PR #401, opened 2019-08-28
Refactoring parts of isomer display changes

LachlanStuart:
See individual commits.

The TypeORM migrations are necessary for the isomer display changes. The rest of the changes are developer quality-of-life improvements.

Parallel python tests still have issues, so I haven't fully integrated them. Some tests that use ElasticSearch or RabbitMQ fail sometimes due to interference from other tests. Regardless, running with pytest-xdist with 2 processes I can get through all the tests much faster.


metaspace/graphql/src/modules/engine/model.ts
```diff
@@ -0,0 +1,63 @@
...
+  @Column({ type: 'int' })
+  width: number;
+  @Column({ type: 'int' })
+  height: number;
+  @Column({ type: 'real', array: true })
+  transform: number[][];
+}
+
+
+export const ENGINE_ENTITIES = [
```
intsco: What about the other tables in the public schema? For them we won't be able to do TypeORM migrations, right?

LachlanStuart: @intsco I abandoned my changes for the `public` tables once I learned about TypeORM's problem with defaults. I could add them if you feel there's value though. There's no technical reason not to do so.

intsco: I don't have a strong opinion on how we do migrations. It just might be confusing if for some tables in `public` schema I need to write manual migration scripts while others will be automatically migrated by TypeORM.

LachlanStuart: Per IRL discussion, we agreed that there wasn't a good reason to split the schemas anymore, so we could probably merge them when we have time/desire to clean up technical debt.

We can add all the `public` schema tables into TypeORM's declarations, and start to use TypeORM for migrations in the `public` schema. Currently this is disabled with `synchronize: false` on the `@Entity` decorator. There's no pressing need to do that now though.

PR #402, opened 2019-08-28
Isomer display back-end changes

LachlanStuart:
This PR depends on the TypeORM changes from `feature/isomer-refactor` branch for the graphql changes to work. Tests won't pass until it is merged.

I decided against normalizing the `graphql.ion` table - it would have been too much work to clean up the database structure and migrate the data that depends on it.

I changed the `test_index_ds_works` test to get data from the database, as I found it would frequently break and didn't seem to actually be delivering much value because it mocked out so much.


metaspace/engine/sm/engine/es_export.py
```diff
@@ -350,13 +353,14 @@ def index_ds(self, ds_id, mol_db, isocalc):
 
             annotation_counts = defaultdict(int)
             fdr_levels = [5, 10, 20, 50]
+            isomer_groups = defaultdict(lambda: [])
```
intsco: One can also use `defaultdict(list)`

PR #413, opened 2019-09-05
Isomer display graphql and UI

LachlanStuart:
* I added a new ElasticSearch field `comps_count_with_isomers` to cache the total number of molecules across all isomers, just so that the UI didn't need to do many separate queries for a summary statistic.
   * In GraphQL I exposed this with a parameterized field `countPossibleCompounds(includeIsomers: Boolean)`. I'll have to add a new parameter later for isobars, when we get to that.
* In GraphQL I added new filters & lookups for Neutral Loss and Chem Mod. I intentionally made them as structures, even though they basically only contain one string and a formatted version of that string, because similar to `AdductSuggestion` there's a good chance new fields will eventually be needed.
* In the filters I removed the `allowEmptyString` flag, because it was actually unused. I accidentally left it in the `[M]+` change, because initially I tried representing adduct-less ionizations as empty strings.
* The UI is viewable on Staging: https://staging.metaspace2020.eu/annotations?ds=2019-09-04_15h08m15s
* Multi-filters are a bit of a hack. I didn't feel it was worth putting too much effort into this, because I only added 1 and I suspect that future multi-filters will have different APIs and requirements.
* I don't know why this happened in this branch (but it's probably related to upgrading Vue). There was some weird behavior causing tests to fail due to `orderedActiveFilters` not correctly triggering observers when I was using lodash's `pull` to modify it. The `state.orderedActiveFilters = without(state.orderedActiveFilters...` vuex changes fixed this. Basically instead of modifying the array in place, it now just replaces the whole array, which I guess is a better target as an observable.


LachlanStuart:
Not ready for review - several bug fixes needed

LachlanStuart:
@intsco This is ready for review

intsco:
@LachlanStuart how about the tests that failed?

LachlanStuart:
@intsco Damn, I didn't see those. That specific issue arises because running tests with coverage produces different snapshots than running tests without coverage. That `AdductFilter.vue` will need a small fix, but the rest of the code is ready.

PR #418, opened 2019-09-10
Misc changes

LachlanStuart:
See individual commits. These aren't actually related to the isomer changes, but it was easier to base them off the meta branch 

The font overrides change is a bug fix for an issue that became apparent due to `[M]-` adducts being added. Basically, the fallback font on OSX and iOS (Helvetica) renders a superscript `-` too small to be recognizable as a minus. This adds a CSS rule to override the font specifically for that one character. It also removes a bunch of places where we overrode ElementUI fonts (which would have also overridden the fix) - ElementUI no longer specifies fonts, so these rules aren't needed anymore.

The dataset thumbnails change makes optical images visible in the datasets list again, but only as a background to the ion image, meaning it's only visible if the ion image is masked. IMO a slight improvement to not being able to find which datasets have optical images...

I don't know if there are drawbacks to using `ts-node-dev` yet, but it's ~5-10x faster than the nodemon way of restarting the whole process on code change. It goes from ~7s to less than a second...

### DIFF
--- a/docker/sm-graphql/docker-entrypoint.sh
+++ b/docker/sm-graphql/docker-entrypoint.sh
@@ -20,7 +20,8 @@ if [ "$SM_DOCKER_ENV" = "development" ]; then
   yarn install
   yarn run deref-schema
   nodemon -e graphql -q --exec "yarn run gen-binding" &
-  exec nodemon -e js,json,ts --require ts-node/register server.js
+  # exec nodemon -e js,json,ts --require ts-node/register server.js
+  exec yarn exec ts-node-dev -- --respawn server.js
 else
   export NODE_ENV=production
   cd /opt/metaspace/metaspace/graphql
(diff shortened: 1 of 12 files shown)


docker/sm-graphql/docker-entrypoint.sh
```diff
@@ -20,7 +20,8 @@ if [ "$SM_DOCKER_ENV" = "development" ]; then
   yarn install
   yarn run deref-schema
   nodemon -e graphql -q --exec "yarn run gen-binding" &
-  exec nodemon -e js,json,ts --require ts-node/register server.js
+  # exec nodemon -e js,json,ts --require ts-node/register server.js
+  exec yarn exec ts-node-dev -- --respawn server.js
```
intsco: The use of double `exec` looks quite exotic here. Just curious why?

LachlanStuart: The first `exec` is just `sh`'s `exec` - it recycles the current `sh`/`bash` process for running the command in order to save memory. 

`yarn exec` is used for running scripts from node modules. Basically it just adds `./node_modules/.bin` to the `PATH` and then executes the arguments. The `--` is needed because otherwise named arguments go to `yarn` instead of `ts-node-dev`.

PR #421, opened 2019-09-11
Isomer display

LachlanStuart:
The meta-PR is finally complete.

Because codecov isn't great, here's the coverage comparing `master` -> `feature/isomer-meta`:

* engine: 82.29% -> 82.34%
* graphql: 52.43% -> 53.35%
* webapp: 50.57% -> 51.11%

The tests haven't really improved, it's just that the code that was already covered by tests has grown substantially.

Outside of the sub-PRs I've added one commit for excluding migrations from the test coverage statistics. Technically they are tested during the tests during database setup, but that runs before the tests start so they aren't included in the coverage.


intsco:
@LachlanStuart the PR looks good to me. Due to #424 merge into master there are minor conflicts in the engine code. Please also format python files with Black as the code in master currently fails the build. I was too quick to merge #424

PR #426, opened 2019-09-13
Dataset molecule filter

LachlanStuart:
This contains substantial refactoring, so I suggest reviewing each commit individually:

* Commit 1 is mainly splitting up a big file into smaller files. Some annotation-specific data processing needed to be moved out of `applyQueryFilters` into `applyColocalizedWithFilter` and `applyColocalizationSamplesFilter`. 
* Commit 2 has refactoring to tidy up the existing filters and add a better `QueryFilterArgs` type that isn't derived from the graphql schema and can be more easily shared between annotations and datasets
* Commit 3 has the new functionality - graphql and webapp changes to add a `hasAnnotationsMatching` filter. This filter uses ElasticSearch to find all datasets that contain at least 1 annotation matching the supplied filter (which can be any annotation filter in the GraphQL API, but in the front-end it's limited to the `compoundQuery` filter)

image

### DIFF
--- a/metaspace/graphql/datasetFilters.js
+++ b/metaspace/graphql/datasetFilters.js
@@ -139,6 +139,7 @@ export const datasetFilters = {
   group: new GroupMatchFilter('', { esField: 'ds_group_id' }),
   project: new ExactMatchFilter('', { esField: 'ds_project_ids' }),
   metadataType: new ExactMatchFilter('Data_Type', {}),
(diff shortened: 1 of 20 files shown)


LachlanStuart:
Retracting the request for review - I'm going to add some webapp changes to the Dataset Summary page to make it more useful when viewing filtered data.

LachlanStuart:
@intsco Ready for review now

PR #442, opened 2019-10-01
Changes required to support M4I databases

intsco:
* Read and write centroids parquet files in chunks to prevent OOM errors in Spark
* Set `request_timeout` to 300s in `update_by_query` to prevent timeout interruptions for long updates. For ~80k documents it takes 2-3 minutes
* Settings of nginx should also be updated to allow requests longer than 60s

### DIFF
--- a/metaspace/engine/sm/engine/formula_centroids.py
+++ b/metaspace/engine/sm/engine/formula_centroids.py
@@ -124,6 +124,7 @@ class CentroidsGenerator(object):
     def _restore_df_chunks(self, spark_df, chunk_size=20 * 10 ** 6):
         total_n = spark_df.rdd.count()
         chunk_n = ceil(total_n / chunk_size)
+        logger.debug(f'Restoring peaks chunks, chunk_size={chunk_size / 1e6}M, chunk_n={chunk_n}')
         spark_dfs = spark_df.randomSplit(weights=np.ones(chunk_n))
         pandas_dfs = [df.toPandas() for df in spark_dfs]
         return pd.concat(pandas_dfs).set_index('formula_i')
@@ -169,10 +170,21 @@ class FormulaCentroids(object):
     def __init__(self, formulas_df, centroids_df):
         u_index_formulas = set(formulas_df.index.unique())
         u_index_centroids = set(centroids_df.index.unique())
-        assert u_index_formulas == u_index_centroids
+        index_intersection = u_index_formulas.intersection(u_index_centroids)
+        index_union = u_index_formulas.union(u_index_centroids)
+        if len(index_union) > len(index_intersection):
+            mismatch_row_n = len(index_union) - len(index_intersection)
+            logger.warning(
+                f'Index mismatch between formulas and centroids dataframes. '
+                f'Ignoring {mismatch_row_n} rows'
+            )
 
-        self.formulas_df = formulas_df.sort_values(by='formula')
-        self._centroids_df = centroids_df.sort_values(by='mz')
+        self.formulas_df = formulas_df[
+            formulas_df.index.isin(index_intersection)
+        ].sort_values(by='formula')
+        self._centroids_df = centroids_df[
+            centroids_df.index.isin(index_intersection)
+        ].sort_values(by='mz')
 
     def centroids_df(self, fixed_size_centroids=False):
         """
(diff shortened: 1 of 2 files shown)


metaspace/engine/sm/engine/formula_centroids.py
```diff
@@ -121,16 +121,25 @@ def _saved(self):
             return (Path(self._ion_centroids_path + '/formulas/_SUCCESS').exists() &
                     Path(self._ion_centroids_path + '/centroids/_SUCCESS').exists())
 
+    def _restore_df_chunks(self, spark_df, chunk_size=20 * 10 ** 6):
+        total_n = spark_df.rdd.count()
+        chunk_n = ceil(total_n / chunk_size)
+        spark_dfs = spark_df.randomSplit([1.0 for _ in range(chunk_n)])
+        dfs = [_.toPandas() for _ in spark_dfs]
```
LachlanStuart: I'm surprised pylint lets you use `_` as a variable name. I've seen it before as a common shortcut for variables that aren't used (so the `randomSplit` usage above is fine), but I don't think it's a good idea for variables that are actually used...

metaspace/engine/sm/engine/formula_centroids.py
```diff
@@ -121,16 +121,25 @@ def _saved(self):
             return (Path(self._ion_centroids_path + '/formulas/_SUCCESS').exists() &
                     Path(self._ion_centroids_path + '/centroids/_SUCCESS').exists())
 
+    def _restore_df_chunks(self, spark_df, chunk_size=20 * 10 ** 6):
+        total_n = spark_df.rdd.count()
+        chunk_n = ceil(total_n / chunk_size)
+        spark_dfs = spark_df.randomSplit([1.0 for _ in range(chunk_n)])
```
LachlanStuart: `randomSplit` seems like a good way to randomly get OOM errors when the random numbers are generated too far apart.

Did you try `.repartition(chunk_n).toLocalIterator()`? See the first code block [here](https://stackoverflow.com/a/53013814/432578) for an example

metaspace/engine/sm/engine/formula_centroids.py
intsco: Pylint complains about any 1-2 symbol variable names. But we have a white list of names and `_` is on it. Will fix it

metaspace/engine/sm/engine/formula_centroids.py
intsco: E.g. `randomSplit([1, 1, 1])` should generate 3 dataframes of approximately the same size, so I'm not sure if there is a danger of getting OOM errors.
I'm always a bit cautious about using `repartition` as it leads to extra shuffling of data over network. Also `toLocalIterator` returns an iterator over rows.

LachlanStuart: > randomSplit([1, 1, 1]) should generate 3 dataframes of approximately the same size,

That's alarmingly poorly documented in the PySpark documentation. They call it random, but actually all the chunk sizes seem to be +/- 10% over 10 runs. I guess that's fine then. Is there enough of a safety margin on the chunk size? Like, would a freak 30M-row chunk cause an OOM? Because there would be very little performance loss reducing the chunk size to something like 1M rows...

> I'm always a bit cautious about using repartition as it leads to extra shuffling of data over network.

It's not significantly expensive here in comparison - `toPandas` is already moving the entire DF over the network to a single node...

> Also toLocalIterator returns an iterator over rows.

Yes, you can first `mapPartitions` to convert the partitions into pandas DFs on the executors so that each row is a DF, and using `toLocalIterator` has the advantage that only one partition is kept in Java memory at once (even though they are all merged in Python memory for the `pd.concat`).

intsco: >That's alarmingly poorly documented in the PySpark documentation. They call it random
I guess this function is meant to be used mostly by people building models. They can use `randomSplit` the same way as `sklearn.model_selection.train_test_split`.
>Because there would be very little performance loss reducing the chunk size to something like 1M rows
Unfortunately there is a performance loss as the whole dataframe is read for each random selection, `chunk_n` times.
>Yes, you can first mapPartitions to convert the partitions into pandas DFs 
It turned out that converting to pandas dataframes this way is much slower than calling `toPandas` on a Spark dataframe.

Anyway, it feels like these changes are temporary as we'll need to update it all during the custom molecular databases feature development.

PR #450, opened 2019-10-16
Add experimental change to mass accuracy calculation

LachlanStuart:
This adds a new option to select analysis version on the upload page when the `advanced_ds_config` feature flag is enabled. Switching version 2 makes these changes:

* Use cpyMSpec==0.4.2
* Take analyzer type into account when generating isotope centroids
* Better parsing of analyzer type from metadata
* Change mass accuracy "ppm" window to scale differently based on analyzer type
* Continuous FDR values (they're useful for debugging. I'm not intending to show them to people when this is released though)

Additionally, I fixed several issues:
* Don't crash on datasets with <20 spectra (the 5% sample taken during segmentation would try to pick 0 spectra, causing an exception)
* Add `replace=False` in the 5% sample selection, so that spectra can't be picked multiple times
* Fix counting of annotations in FDR levels during elasticsearch export (was broken by continuous FDR values)
* Round FDR% to an integer in AnnotationsTable so that long fractional numbers don't break the UI

This PR is feature complete, but I need to do more testing on it before moving it out of draft. It's especially important for me to make sure that analysis_version=1 behavior is unchanged.


LachlanStuart:
Notes to Future @LachlanStuart:

This isn't ready for merging. There are ~3~ 2 unknowns that need investigating:

1. ~The continuous FDR change caused some annotations to actually change FDR level (localhost/Untreated_lite dataset). This should not be possible. Maybe the weird differing mz values(unknown #2) caused some unseen difference in the decoy set.~ POTENTIALLY SOLVED: v1 uses `df.fdr < fdr_threshold`, when it should use `<=`, which would potentially cause FDR values that were exactly `0.1` to be put in the `0.2` bin.
2. The isotopic peak m/z values differ by a  between the discrete FDR and continuous FDR datasets. This needs further investigation. It seems to vary relative to m/z, so it may be a PPM offset.
3. The Bovine dataset saw a significant decrease in FDR10% annotations compared to both v1 and "v1 1ppm 500k". I really need to make a histogram of m/z values at specific FDRs to see what's happening here.

Also, there are a couple things to do:

1. `mass_accuracy_bounds` is incorrect. `ppm` is the correct unit for mass accuracy, and the new code only tries to scale by mass precision, whereas this function must compensate for both. It would be best to compute this as something like e.g. 1-2 ppm + fwhm.
2. v1 results need to be re-tested to ensure there are no changes

LachlanStuart:
> Notes to Future @LachlanStuart:
> 
> This isn't ready for merging. There are ~3~ 2 unknowns that need investigating:
> 
> 1. ~The continuous FDR change caused some annotations to actually change FDR level (localhost/Untreated_lite dataset). This should not be possible. Maybe the weird differing mz values(unknown #2) caused some unseen difference in the decoy set.~ POTENTIALLY SOLVED: v1 uses `df.fdr < fdr_threshold`, when it should use `<=`, which would potentially cause FDR values that were exactly `0.1` to be put in the `0.2` bin.

I processed a rich dataset and the only cases of FDRs changing were several datasets with exactly 0.5 FDR (continuous) were included in the continuous FDR dataset and not the discrete FDR dataset. The `<=` vs `<` explanation seems extremely likely. `<=` is clearly the correct way to do it (i.e. an annotation that has 2 decoys and 7 real annotations above it should be FDR (2 / (2+7+1) <= 20%) not (`2 / (2+7+1) < 50%`) , so I will leave it as `<=` in v2.

> 2. The isotopic peak m/z values differ by a  between the discrete FDR and continuous FDR datasets. This needs further investigation. It seems to vary relative to m/z, so it may be a PPM offset.

I found and fixed two potential causes:
* When re-indexing datasets via script, `isocalc_wrapper`'s cache keys didn't include the `analysis_version`, meaning that `cpyMSpec==0.3.5` and `cpyMSpec==0.4.2` results were mixed into the same cache. I remember re-indexing these datasets several times to troubleshoot cpyMSpec version issues
* Incorrect versions of `cpyMSpec` may have been imported on my local computer due to not correctly resetting the environment when switching branch. I've added assertions that the correct versions are imported.

> 3. The Bovine dataset saw a significant decrease in FDR10% annotations compared to both v1 and "v1 1ppm 500k". I really need to make a histogram of m/z values at specific FDRs to see what's happening here.

Density map of MSMs and FDRs with respect to m/z:
* Columns are datasets, top row is MSM, bottom row is FDR
* 2D histogram color: log(count of annotations at this MSM/FDR and m/z)
* White line: average MSM/FDR value at this m/z
* Red line (same data for both plots): total count of annotations at this m/z

image

The "v1/1ppm/500k RP"(middle) and "v2"(right) show reduced average MSM scores (slightly expected), as well as significantly reduced number of annotations at the upper end of the m/z range. This could be a mass accuracy issue with the dataset. I'm deferring further investigation until we look at recalibration again.


> Also, there are a few things to do:
> 
> 1. `mass_accuracy_bounds` is incorrect. `ppm` is the correct unit for mass accuracy, and the new code only tries to scale by mass precision, whereas this function must compensate for both. It would be best to compute this as something like e.g. 1-2 ppm + fwhm.
 
If mass accuracy is fixed (i.e. with recalibration), this won't be a problem. I'm leaving this as-is.

> 2. v1 results need to be re-tested to ensure there are no changes

Done. No changes.

> 3. Some TOF data should be tested with v2.

Skipping this.

LachlanStuart:
@intsco I've done the testing I wanted to do, and added a few small changes. Please review the new commits.

PR #466, opened 2019-11-04
Use pickle instead of msgpack for serialization of arrays and dataframes

### LINKED ISSUE #420: Change Pandas serialization library: msgpack -> pyarrow
FutureWarning: The read_msgpack is deprecated and will be removed in a future version.

intsco:
Pandas recently deprecated msgpack use for data serialization so using pyarrow+parquet seemed to be the best alternative but didn't live up to the expectations, pickle turned out to be the fastest and easiest option.

* The code responsible for dataset and centroids segmentation switched to using pickle for serialization.
* The in-memory spectra chunk size was reduced to 500MB to speed things up.
* Centroids cache used parquet before. But I took the opportunity to switch it from Spark's implementation to pyarrow as it's Python only solution which should solve annoying Java OOM issues we've seen before.

Test dataset (CT26_xenograft) segmentation runtime.

msgpack:
1 GB chunk size, 33 sec
5 GB chunk size, 52 sec

pyarrow:
1 GB chunk size, 1 min 4 sec
5 GB chunk size, 1 min 9 sec

pickle:
0.5 GB chunk size, 28.9 s
1 GB chunk size, 32.1 s
5 GB chunk size, 52.2 s

Resolves #420

### DIFF
--- a/metaspace/engine/sm/engine/msm_basic/segmenter.py
+++ b/metaspace/engine/sm/engine/msm_basic/segmenter.py
@@ -4,6 +4,8 @@ from shutil import rmtree
 
 import numpy as np
 import pandas as pd
+import pyarrow as pa
+import pyarrow.parquet as pq
 
 from sm.engine.errors import SMError
 from sm.engine.msm_basic.formula_imager import get_pixel_indices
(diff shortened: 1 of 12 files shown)


metaspace/engine/sm/engine/msm_basic/segmenter.py
```diff
@@ -74,9 +76,11 @@ def segment_spectra_chunk(sp_mz_int_buf, mz_segments, ds_segments_path):
...
 
     for segm_i, (start, end) in enumerate(zip(segm_starts, segm_ends)):
-        pd.to_msgpack(
-            ds_segments_path / f'ds_segm_{segm_i:04}.msgpack', sp_mz_int_buf[start:end], append=True
-        )
+        table = pa.Table.from_pandas(pd.DataFrame(sp_mz_int_buf[start:end]))
+        with pq.ParquetWriter(
+            ds_segments_path / f'ds_segm_{segm_i:04}.parquet', table.schema
+        ) as writer:
+            writer.write_table(table)
```
LachlanStuart: As far as I can tell, `ParquetWriter.write_table` overwrites files instead of appends. This means for datasets with more than 1000 spectra, only the last chunk of spectra would survive.

PR #476, opened 2019-11-08
Fix formatting according to new version of black

intsco:
Only formatting changed so no need for a review


intsco:
Only formatting changes so no need for a review

PR #484, opened 2019-11-26
Add 'stopping' to list of valid statuses of existing instance

intsco:


ansible/aws/aws_instance_manager.py
```diff
@@ -31,7 +31,10 @@ def find_inst_by(self, host_group, first=False):
             self.ec2.instances.filter(
                 Filters=[
                     {'Name': 'tag:hostgroup', 'Values': [host_group]},
-                    {'Name': 'instance-state-name', 'Values': ['running', 'stopped', 'pending']},
+                    {
+                        'Name': 'instance-state-name',
+                        'Values': ['pending', 'running', 'stopping', 'stopped'],
```
LachlanStuart: > instance-state-name - The state of the instance (pending | running | shutting-down | terminated | stopping | stopped ).

Should `shutting-down` also be included?

intsco: According to [this](https://docs.aws.amazon.com/AWSEC2/latest/WindowsGuide/ec2-instance-lifecycle.html) diagram, an instance in this state won't be able to start later.

Issue #434, opened 2019-09-19
Spark cluster stability testing for large datasets


intsco:
Dataset:
https://metaspace2020.eu/datasets?q=2017-07-28_12h59m41s

Error:

September 19th 2019, 17:29:52.418 | 19/09/19 17:29:48 ERROR TaskSchedulerImpl: Lost executor 2 on 172.31.3.92: worker lost 

  | September 19th 2019, 17:29:45.411 | 19/09/19 17:29:35 WARN TransportChannelHandler: Exception in connection from /172.31.5.121:34812 | -

  | September 19th 2019, 17:29:38.408 | 19/09/19 17:29:35 ERROR TaskSchedulerImpl: Lost executor 6 on 172.31.7.163: worker lost | -

  | September 19th 2019, 17:21:12.199 | 19/09/19 17:21:06 WARN TaskSetManager: Lost task 91.1 in stage 10.0 (TID 1080, 172.31.1.155, executor 4): org.apache.spark.api.python.PythonException: Traceback (most recent call last): | -

  | September 19th 2019, 17:18:43.222 | 19/09/19 17:18:38 WARN TaskSetManager: Lost task 9.1 in stage 10.0 (TID 1056, 172.31.3.92, executor 2): org.apache.spark.api.python.PythonException: Traceback (most recent call last): | -

  | September 19th 2019, 17:12:40.906 | 19/09/19 17:12:40 ERROR TaskSchedulerImpl: Lost executor 1 on 172.31.5.121: Remote RPC client disassociated. Likely due to containers exceeding thresholds, or network issues. Check driver logs for WARN messages. | -

  | September 19th 2019, 17:11:45.904 | 19/09/19 17:11:41 WARN HeartbeatReceiver: Removing executor 3 with no recent heartbeats: 417549 ms exceeds timeout 360000 ms | -

  | September 19th 2019, 17:10:50.902 | 19/09/19 17:10:41 WARN HeartbeatReceiver: Removing executor 1 with no recent heartbeats: 373298 ms exceeds timeout 360000 ms | -

  | September 19th 2019, 17:03:20.849 | 19/09/19 17:02:37 WARN HeartbeatReceiver: Removing executor 0 with no recent heartbeats: 360607 ms exceeds timeout 360000 ms | -

  | September 19th 2019, 16:43:54.693 | Processing segments... | INFO

Dataset:
2019-11-22_08h57m39s

Error:
```Traceback (most recent call last):
[...]

PR #496, opened 2019-12-06
Memory optimized iso image generation

intsco:
The objective of this PR is to reduce memory consumption by the worker instances due to too many dataset segments being loaded into the memory for one centroids segment. Previously, this behavior resulted in OOM errors in Python workers for large datasets (>20GB)

- The simplest solution was to load only one dataset segment at a time. Unfortunately, it required additional tricks to be implemented to prevent centroid peak images from being clipped as a result of falling too close to a boundary between two dataset segments. 
- To reduce the number of dataset segments for the large datasets, the dataset segment size was bumped up to 20MB.
- Dataset segments are stored as pandas dataframes to reduce memory consumption by using different precision for different columns.
- `ImzMLParserWrapper` class was added to handle a few conversions between pyimzML and the engine. The main one is reading spectra while removing peaks with zero intensities.
- A few unrelated bug fixes were also added on the way, see the commit messages.

Resolves: #434


intsco:
@LachlanStuart please wait before reviewing the PR. The memory issue still persists for some datasets. I'm investigating it.

metaspace/engine/sm/engine/msm_basic/segmenter.py
```diff
@@ -47,13 +50,18 @@ def spectra_sample_gen(imzml_parser, sample_size):
         yield sp_idx, mzs, ints
 
 
-def define_ds_segments(sample_mzs, total_mz_n, mz_precision, ds_segm_size_mb=5):
+def define_ds_segments(sample_mzs, sample_ratio, imzml_parser, ds_segm_size_mb=5):
     logger.info(f'Defining dataset segment bounds')
+    sp_arr_row_size_b = (
+        np.dtype(SP_IDX_DTYPE).itemsize
+        + np.dtype(imzml_parser.mzPrecision).itemsize
+        + np.dtype(INT_DTYPE).itemsize
```
LachlanStuart: `INT_DTYPE` looks like "integer datatype", and is a bit confusing here. `SP_INT_DTYPE` or `INTENSITY_DTYPE` would be better.

intsco:
@LachlanStuart the PR is finally ready. It has grown probably too big at this point. Sorry about that.

intsco:
@LachlanStuart should we merge this PR into the release branch so that users can start benefiting from the changes soon?

metaspace/engine/sm/engine/msm_basic/formula_imager.py
```diff
@@ -102,17 +107,10 @@ def read_centroids_segment(segm_path):
...
-    if sp_arr:
-        sp_arr = np.concatenate(sp_arr)
-        sp_arr = sp_arr[sp_arr[:, 1].argsort()]  # assume mz in column 1
-    else:
-        sp_arr = np.empty((0, 3))
-    return sp_arr
+    for ds_segm_i in range(first_segm_i, last_segm_i + 1):
+        segm_path = get_file_path(f'ds_segm_{ds_segm_i:04}.pickle')
+        ds_segm_df = read_ds_segment(segm_path)
+        yield ds_segm_df.sort_values(by='mz')
```
LachlanStuart: If `read_ds_segment` takes the `if not sp_chunk_list` path and returns an empty DataFrame, this will raise an exception about the `mz` column not existing. I suggest either making `read_ds_segment` return `None` and not yielding `None` values from `read_ds_segments`, or adding column definitions to `read_ds_segment`'s empty DataFrame.

metaspace/engine/sm/engine/msm_basic/segmenter.py
```diff
@@ -66,16 +73,15 @@ def define_ds_segments(sample_mzs, total_mz_n, mz_precision, ds_segm_size_mb=5):
...
-    segm_starts = np.searchsorted(sp_mz_int_buf[:, 1], segm_left_bounds)
-    segm_ends = np.searchsorted(sp_mz_int_buf[:, 1], segm_right_bounds)
+    segm_starts = np.searchsorted(sp_chunk_df.mz.values, segm_left_bounds)
+    segm_ends = np.searchsorted(sp_chunk_df.mz.values, segm_right_bounds)
 
     for segm_i, (start, end) in enumerate(zip(segm_starts, segm_ends)):
         segment_path = ds_segments_path / f'ds_segm_{segm_i:04}.pickle'
         with open(segment_path, 'ab') as f:
-            pickle.dump(sp_mz_int_buf[start:end], f)
+            pickle.dump(sp_chunk_df.iloc[start:end], f)
```
LachlanStuart: Have you checked the segment file sizes? `sp_chunk_df.iloc[start:end]` should return a view, not a copy...

intsco: Yeah, I did check that. Actually, this time the segment sizes are much more homogeneous (at least for the dataset I tested). What makes you think that views should pull the whole array when being serialized?

LachlanStuart: What made me think that was the memory pressure we had caused by using views instead of copies in `gen_iso_images`. I just tested it though, and pickles of views and copies seem to be the same size, so I guess ndarrays and DataFrames override the default pickling behavior so that they only pickle the included slices of data.

PR #506, opened 2020-01-08
Update python-client upload_dataset_api notebook

intsco:
Also add improved `SMInstance.submit_dataset_v2` method

### DIFF
--- a/metaspace/python-client/metaspace/sm_annotation_utils.py
+++ b/metaspace/python-client/metaspace/sm_annotation_utils.py
@@ -1127,11 +1127,13 @@ class SMInstance(object):
     ):
         """
         Submit a dataset for processing on the SM Instance
+
         :param imzml_fn: file path to imzml
         :param ibd_fn: file path to ibd
         :param ds_name: dataset name
         :param metadata: a properly formatted metadata json string
-        :param s3_bucket: boto3 s3 bucekt object, both the user has write permission to and METASPACE can access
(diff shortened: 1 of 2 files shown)


metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -1111,6 +1113,46 @@ def submit_dataset(
...
+        """
+        Submit a dataset for processing on the SM Instance
+        :param imzml_fn: file path to imzml
+        :param ibd_fn: file path to ibd
+        :param is_public: make dataset public
+        :param ds_name: dataset name
+        :param moldbs: list molecular databases
+        :param metadata: a properly formatted metadata json string
+        :param adducts: list of adducts
+        :param s3_bucket: this should be a bucket that both the user has write permission to and METASPACE can access
```
LachlanStuart: This should probably mention that it specifically needs to be a boto `s3.Bucket` instance.

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -1111,6 +1113,46 @@ def submit_dataset(
...
+    def submit_dataset_v2(
+        self,
+        imzml_fn,
+        ibd_fn,
+        ds_name,
+        metadata,
+        is_public=None,
+        moldbs=None,
+        adducts=None,
+        s3_bucket=None,
```
LachlanStuart: This function can't run without `s3_bucket`. I don't think it should have a `None` default value.

PR #508, opened 2020-01-17
API key authentication

### LINKED ISSUE #350: Force Google login to show account switcher screen
Currently if you're logged in to a single Google account the doing Google login won't give you the ability to change to a different Google account.

Seems like there could be an easy fix: https://stackoverflow.com/questions/44267756/how-to-skip-the-choose-account-pop-up-with-google-sign-in-when-requesting-an-add

LachlanStuart:
* Switches to use Passport for authentication across all of graphql. Previously Passport was only used in the `/api_auth` routes that were used for signing in and generating JWTs, and `jwt-simple` was used for other routes. This was a bit annoying as there were effectively 2 competing mechanisms for authentication that could overlap, and this PR would have had to add a third. 
    * The downside of this is that passport-jwt's error handling is garbage. I had to add two workarounds for JWT and API-Key errors to ensure that the errors from invalid tokens are correctly sent back to the client. Passport doesn't normally differentiate between "You didn't try to authenticate" and "You tried to authenticate but did it wrong"...
    * I've reduced the scope of JWT tokens so that only `/graphql` can use JWTs and API keys to authenticate. This prevents a few security edge cases by ensuring the `/auth_api` routes can only be accessed through password/Google authentication.
    * I've now allowed `/graphql` to be accessed by the same cookie used to generate the JWT in the first place, because it was a tiny tweak and we may consider removing JWTs in the future
* Replaces `jwt-simple` with `jsonwebtoken` as it seems to be better supported
* Adds `context.user.authMethod` field so that resolvers can adjust behavior based on how the user authenticated
* Adds `credentials.apiKey` and `credentials.apiKeyLastUpdated` database columns
* Adds `User.apiKey` graphql query, and `resetUserApiKey(userId: ID!, removeKey: Boolean! = False)` mutation. `User.apiKey` can't be queried if you're using an API key.
* Adds `addApiKeyInterceptorToSchema` method that logs all mutations performed via API key, and only allows whitelisted mutations
* Updates python-client to support api-key authentication, and updates the example notebooks to use it instead of email/password authentication
* Adds UI to the "My Account" page to create/delete your API key:

#### My Account page (no key yet)
image

#### Generating the key:
image

#### After creating the key
image

#### Revoking the key
image

This isn't the full set of changes for MetaboLights integration - I still need to add support for linking to external integrations. That will come in a separate PR.

Fixes #350


metaspace/graphql/src/modules/auth/controller.ts
```diff
@@ -195,23 +250,33 @@ const configureGoogleAuth = (router: IRouter<any>) => {
...
+          done(err);
         }
-        return user;
-      })
+      }
     ));
 
     router.get('/google', Passport.authenticate('google', {
-      scope: ['profile', 'email']
+      prompt: 'select_account',
```
LachlanStuart: Unrelated change: adding `prompt: 'select_account'` fixes #350

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -841,11 +859,15 @@ def __init__(self, host='https://metaspace2020.eu', verify_certificate=True):
...
+    def login(self, email=None, password=None, api_key=None):
+        assert (
+            email and password
+        ) or api_key, 'Either email and password, or api_key must be provided'
         self._config['usr_email'] = email
         self._config['usr_pass'] = password
+        self._config['usr_api_key'] = api_key
         self.reconnect()
-        return self._gqclient.logged_in
+        assert self._gqclient.logged_in, 'Login failed'
```
LachlanStuart: @aeisenbarth Please review changes to python client

metaspace/graphql/src/getContext.ts
```diff
@@ -8,7 +9,7 @@ import {Request, Response} from 'express';
 import * as _ from 'lodash';
 import * as DataLoader from 'dataloader';
 
-const getBaseContext = (jwtUser: JwtUser | null, entityManager: EntityManager,
+const getBaseContext = (jwtUser: JwtUser | UserModel | null, entityManager: EntityManager,
```
intsco: Isn't is confusing that the `jwtUser` variable can contain `UserModel` object?

metaspace/graphql/src/modules/auth/NoisyJwtStrategy.ts
```diff
@@ -0,0 +1,24 @@
...
+export default class NoisyJwtStrategy extends JwtStrategy {
+  authenticate(req: Request, options: StrategyOptions) {
+    const oldFail = this.fail;
+    this.fail = function (challenge: any, status?: number) {
+      const reason = challenge instanceof Error ? challenge.message : challenge;
+
+      if (reason && typeof reason !== 'number' && reason !== 'No auth token') {
+        return this.error(challenge);
+      } else {
+        return (oldFail as any).apply(this, arguments);
```
intsco: I have really hard time understanding this piece of code. What do `this` and `arguments` refer too?

LachlanStuart: There's 3 bits of JavaScript magic here:

* `this` in JS is funky. If you call `foo.bar()`, then inside the `bar` function `this` will be `foo` because of the way it was called. However, unlike Python, if you do `const foobar = foo.bar; foobar();`, `foobar` will see `this` set to `undefined` or `global` depending on if you're in "strict mode", because `foo.` wasn't used when it was called. There are several ways to circumvent this behavior: `func.call` and `func.apply` call the function with a specified value as `this`, and `func.bind` can be used to wrap the function to force a certain `this` value.
* `anyFunction.apply(thisVar, someArray)` calls `anyFunction` with `thisVar` as its `this` context and `someArray` expanded out into its arguments. There's a related `anyFunction.call(thisVar, argOne, argTwo, argThree)` which does the same but doesn't expand its arguments as an array.
* `arguments` is an implicit variable, similar to `this`, which is an Array-like that contains the arguments to the current function supplied by the caller. Similar to `this`, it only works in normal `function`s, not in "arrow functions". It also bases its length on how many arguments were supplied by the caller, regardless of how many arguments the function has declared. e.g. with `function returnArgs(a) { return arguments; }`, `returnArgs()` will return `[]` and `returnArgs(1,2,3)` will return `[1,2,3]`.

So basically this code calls `oldFail` with the exact same `this` and arguments that were used to call the current function.

metaspace/graphql/src/getContext.ts
LachlanStuart: Good point. I've renamed it to `userFromRequest`. I couldn't rename it to just `user` as that name is already in use inside the function.

metaspace/graphql/src/modules/user/controller.ts
```diff
@@ -104,24 +111,27 @@ export const Resolvers = {
       }
       return null;
     },
+
+    async apiKey({scopeRole, id: userId}: UserSource, args: any, ctx: Context): Promise<string|null> {
+      const allowedAuthMethods: AuthMethod[] = [AuthMethodOptions.JWT, AuthMethodOptions.SESSION];
+      if ((scopeRole === SRO.PROFILE_OWNER || ctx.isAdmin)
+        && allowedAuthMethods.includes(ctx.user.authMethod)) {
+        const userModel = await findUserById(userId, true);
+        return userModel && userModel.credentials && userModel.credentials.apiKey || null;
```
intsco: I don't quite see how this line returns something that corresponds to the type `Promise<string|null>`. To me, it looks more like it returns a boolean or null.

LachlanStuart: `apiKey` is a `string`, and like Python, JS preserves "truthy" and "falsy" values in boolean logic. `userModel.credentials.apiKey || null` will return `userModel.credentials.apiKey` if it is truthy, otherwise `null`. e.g. `'123' || null === '123'` and `'123' && '456' && '789' === '789'`.

metaspace/graphql/src/modules/auth/NoisyJwtStrategy.ts
intsco: Thanks for explaining! This definitely a tricky part of JS.

PR #514, opened 2020-01-27
Move webapp to vue-cli

LachlanStuart:
Moving webapp into the vue-cli framework makes it easier to keep everything up to date and start using new features that come with Vue: 
* [Modern Mode](https://pusher.com/tutorials/vue-cli-3#modern-mode) makes two copies of the compiled JS assets - one for old browsers, and one for new browsers. The new browser version is ~33% smaller, meaning faster page loads for everybody who isn't still on IE11.
* Components using TypeScript + JSX never worked in our project due to weird webpack issues, but they work out of the box with vue-cli. They're much more readable than vue templates, and would clean up our code if we started using them.
* [Vue 3.0's Composition API](https://vue-composition-api-rfc.netlify.com/) should further clean up code, but really only makes sense to use with TypeScript + JSX.

One large aspect of this change is that vue-cli assumes that in production your API server will be a separate application. By default vue-cli just outputs static files and assumes you'll configure nginx, etc. to serve them. To work around this, I've moved the server-side code for uploading datasets to GraphQL.
* To simplify the nginx config, I moved the upload API to `/dataset_upload` instead of `/upload`, because `/upload` overlaps with the URL for the upload page
* I also added nginx rules for the static content so that browsers can cache them forever. This is safe because they now include a hash in their filename, so their filename will change when they are updated. For extra safety, I configured deployment/nginx to keep the last deployed version of the files, just in case someone has cached their `index.html`.

This change does not include code linting fixes. I'll add those in a separate PR, because they're mostly automated and won't need as much review.

I decided not to add Vue-SSR (Server-side rendering), because it would have required a lot more changes than I was comfortable including in this PR.

Some of the tests required additional type annotations. Previously they weren't actually type-checked during testing, but now they are.


ansible/aws/ansible.cfg
```diff
@@ -7,7 +7,7 @@ vault_password_file = ./vault_pass.txt
 interpreter_python = auto
 
 [ssh_connection]
-ssh_args = -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null
+ssh_args = -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o ControlMaster=auto -o ControlPersist=60s
```
LachlanStuart: [Multiplexing](https://en.wikibooks.org/wiki/OpenSSH/Cookbook/Multiplexing) speeds up SSH connections a bit.

ansible/aws/templates/all/vars.yml.template
```diff
@@ -262,12 +255,12 @@ nginx_servers:
...
           - include proxy-params.conf;
 
-  - desc: HTTP redirect
+  - desc: HTTPS redirect
     params:
       - listen 80 default_server;
       - listen [::]:80 default_server;
       - server_name _;
-      - return 301 https://$host$request_uri;
+      - return 301 https://{{ sm_webapp_hostname }}$request_uri;
```
LachlanStuart: Redirecting to `$host` causes certificate errors for other domains, because we don't have certs for all the domains that nginx listens to, e.g. http://metasp.eu . It's easier to just redirect to our main domain.

ansible/roles/base/tasks/main.yml
```diff
@@ -6,7 +6,7 @@
 
 - name: Install python-simplejson
   become: yes
-  raw: apt-get install -y python-simplejson
+  raw: apt-get install -y python3-simplejson
```
LachlanStuart: I didn't get a chance to fully debug this, but Ansible couldn't find pip in the "Update pip" step. Python 2 seemed to be interfering with Python 3. `python-simplejson` seemed to be the reason Python 2 was installed...

ansible/roles/mol_db/defaults/main.yml
```diff
@@ -4,5 +4,4 @@ mol_db_import_databases: false
 mol_db_import_mol_images: false
 mol_db_port: 5001
 mol_db_miniconda_env:
-  name: mol-db
```
LachlanStuart: I removed `mol_db_miniconda_prefix` because it was redundant to `miniconda_prefix`, and provisioning would fail if the two were out of sync (which happened in my case).

ansible/roles/mol_db/tasks/main.yml
```diff
@@ -33,6 +36,23 @@
...
+    executable: /bin/bash
+  shell: "source activate {{ mol_db_miniconda_env.name }} && pip install -e ."
+
+- name: Pull MolDB config from the remote host
+  fetch: src={{ mol_db_home }}/conf/local.ini.template dest=/tmp/local.ini.template
+         flat=yes fail_on_missing=yes
+
+- name: Put MolDB config to ./conf/local.ini
+  template: src=/tmp/local.ini.template dest={{ mol_db_home }}/conf/local.ini
+            owner=ubuntu group=ubuntu mode=0600
```
LachlanStuart: I copied these steps from the "deploy" task because into the "provision" task because they're needed for `import_mol_databases.yml` to run.

ansible/roles/sm_web/tasks/main.yml
```diff
@@ -45,12 +45,8 @@
   file: dest=/opt/data/sm_data state=directory owner=ubuntu group=ubuntu mode=0700
   become: yes
 
-- name: Copy sm-api supervisor config file
-  template: src=sm-api.supervisor.j2 dest=/etc/supervisor/sm-api.supervisor
```
LachlanStuart: This step was duplicated from the "sm-api" role. `sm-api.supervisor.j2` is missing in this role's files, but is present in sm-api's.

docker/sm-graphql/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:8.11.3
+FROM node:8.17.0
```
LachlanStuart: One of the vue-cli dependencies wouldn't work with 8.11, so I had to update. 

Prod is on 8.16.2 which should work. Staging was on 8.11.0 so I had to update it (now on 8.17.0 because 8.16.2 wasn't available on apt)

docker/sm-webapp/config/clientConfig.json
```diff
@@ -1,9 +1,8 @@
 {
-  "graphqlUrl": "http://localhost:8999/graphql",
-  "wsGraphqlUrl": "ws://localhost:8999/ws",
+  "graphqlUrl": null,
+  "wsGraphqlUrl": null,
```
LachlanStuart: `null` autodetects the url based on current host & port

metaspace/engine/sm/engine/es_export.py
```diff
@@ -108,15 +108,20 @@ def __init__(self, es_config=None):
...
 
-        index = next(iter(indices.keys()))
+        index = next(iter(indices.keys()), None)
         if len(indices) > 1:
             logger.warning(
                 f'Multiple indices mapped on to the same alias: {indices}. '
                 f'Arbitrarily choosing {index}'
             )
+        elif index is None:
+            index = yin
```
LachlanStuart: These changes are to prevent exceptions when no indexes/aliases exist

metaspace/graphql/src/modules/webServer/fineUploaderLocalMiddleware.ts
```diff
@@ -0,0 +1,89 @@
...
+import {Router, Request, Response, NextFunction} from 'express';
+import * as mkdirp from 'mkdirp';
+import * as multer from 'multer';
+import * as bodyParser from 'body-parser';
+
+const readFileAsync = promisify(fs.readFile);
+const openAsync = promisify(fs.open);
+const closeAsync = promisify(fs.close);
+const writeAsync = promisify(fs.write);
+const unlinkAsync = promisify(fs.unlink);
```
LachlanStuart: 2 significant changes in these two middleware files: 
* Converted to TypeScript
* `combineChunks` in this file has been rewritten to not use `concat-files`, as that dependency didn't have type annotations and used the synchronous node filesystem APIs, which block other JS code from running in the same process.

metaspace/webapp/src/lib/sanitizeIt.ts
```diff
@@ -0,0 +1,17 @@
+import sanitizeHtml from 'sanitize-html';
```
LachlanStuart: Unrelated change - I moved this out of `util.ts` because `sanitize-html` has about 150kB (65kB gzipped) of dependencies. It's only used on the group/project pages.

metaspace/webapp/src/modules/Account/components/__snapshots__/CreateAccountDialog.spec.ts.snap
```diff
@@ -12,8 +12,7 @@ exports[`CreateAccountDialog should match snapshot 1`] = `
                 <!---->
                 <div class="el-form-item__content">
                   <div class="el-input">
-                    <!---->
-                    <input type="text" autocomplete="off" placeholder="First name" required="required" class="el-input__inner">
+                    <!----><input type="text" autocomplete="off" placeholder="First name" required="required" class="el-input__inner">
```
LachlanStuart: No actual change to the rendered content here. One of the updated packages caused whitespace to be handled differently in test snapshots.

metaspace/webapp/src/modules/Annotations/__snapshots__/AnnotationTable.spec.ts.snap
```diff
@@ -163,65 +152,22 @@ exports[`AnnotationTable should match snapshot 1`] = `
...
-          <!---->
-        </ul>
-        <button type="button" disabled="disabled" class="btn-next">
-          <i class="el-icon el-icon-arrow-right"></i>
-        </button><span class="el-pagination__sizes"><div class="el-select el-select--mini"><!----><div class="el-input el-input--mini el-input--suffix"><!----><input type="text" readonly="readonly" autocomplete="off" placeholder="请选择" class="el-input__inner"><!----><span class="el-input__suffix"><span class="el-input__suffix-inner"><i class="el-select__caret el-input__icon el-icon-arrow-up"></i><!----><!----><!----><!----><!----></span>
-        <!---->
-        </span>
-        <!---->
-        <!---->
+      <mock-el-pagination total="1" page-size="15" page-sizes="15,20,25,30" current-page="1" layout="prev,pager,next,sizes"></mock-el-pagination>
```
LachlanStuart: `el-pagination` suddenly started being problematic in tests. Unsure why, but the easy fix was to just globally mock it out.

metaspace/webapp/vue.config.js
```diff
@@ -0,0 +1,45 @@
...
+      const compressionTest = /\.(js|css|json|txt|html|ico|svg)(\?.*)?$/i;
+      config.plugins.push(
+        new CompressionWebpackPlugin({
+          test: compressionTest,
+          minRatio: 0.9
+        }),
+        // new BrotliPlugin({
+        //   test: compressionTest,
+        //   minRatio: 0.9
+        // })
```
LachlanStuart: Nginx still doesn't have Brotli support :(
It would reduce the page size by ~15-20%.

ansible/roles/mol_db/tasks/main.yml
```diff
@@ -14,9 +14,12 @@
   become_user: postgres
   postgresql_db: name=mol_db owner=mol_db encoding=UTF8 template=template0
 
+- name: Fix conda
+  shell: "conda config --set restore_free_channel true"
+
```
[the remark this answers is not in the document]
LachlanStuart: We do need it - conda wouldn't run the next step without this. 

I didn't check which specific package required it, but since we're removing mol-db soon I didn't think it was worth an investigation

ansible/aws/templates/all/vars.yml.template
```diff
@@ -305,6 +312,10 @@ nginx_servers:
         params:
           - proxy_pass http://isotope_image_server/fs/ion_thumbnails/;
           - limit_except GET {deny all;}
+      - path: /dataset_upload
+        params:
+          - proxy_pass http://isotope_image_server/dataset_upload;
```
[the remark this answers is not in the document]
LachlanStuart: I've renamed it to `storage_server` as discussed

PR #516, opened 2020-01-29
Replace moldb service with database tables

intsco:
This PR is the first step towards the custom databases feature.

* All the code that communicates with the moldb service is removed. Instead, molecular databases and their molecules are loaded directly from the database.
* `/isotopic_patterns` endpoint moved from the moldb service to sm-api.
* The molecular database import code moved to engine. Only import with a script is supported so far.


metaspace/graphql/src/modules/dataset/controller/Mutation.ts
```diff
@@ -77,15 +76,17 @@ function validateMetadata(metadata: MetadataNode) {
...
-      throw new UserError(JSON.stringify({
-        'type': 'wrong_moldb_name',
-        'moldb_name': name
-      }));
+async function molDBsExist(entityManager: EntityManager, molDBNames: string[]) {
+  const foundMolDBNames = (await entityManager.getRepository(MolecularDB)
+    .find({ where: { name: In(molDBNames) } }))
+    .map(moldb => moldb.name);
+
+  if (foundMolDBNames.length < molDBNames.length) {
```
LachlanStuart: Comparing length here isn't enough because db names aren't necessarily unique - multiple dbs can exist with the same version, which could cause `foundMolDBNames.length` to be higher than expected due to duplicates.

metaspace/graphql/src/modules/moldb/model.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    @Column({ type: 'text' })
+    mol_name: string;
+
+    @Column({ type: 'text' })
+    formula: string;
+
+    @ManyToOne(type => MolecularDB)
+    moldb: MolecularDB;
+}
```
LachlanStuart: Unique constraints on these two tables would probably save some confusion down the line.

It looks like the old code ensured unique `Molecule`s by using a composite primary key over `db_id, inchikey`

metaspace/graphql/src/modules/moldb/model.ts
```diff
@@ -0,0 +1,43 @@
...
+
+@Entity({ schema: 'public' })
+export class Molecule {
+
+    @PrimaryGeneratedColumn()
+    id: number;
+
+    @Column({ type: 'text' })
+    mol_id: string;
```
LachlanStuart: I added `SnakeCaseNamingStrategy` to TypeORM, so you can use `molId` and `molName` here and the DB will still have `mol_id` and `mol_name`.

metaspace/graphql/src/modules/moldb/model.ts
```diff
@@ -0,0 +1,43 @@
...
+  @Column({ type: 'text' })
+  version: string;
+
+  @OneToMany(type => Molecule, molecule => molecule.moldb)
+  molecules: Molecule[];
+
+}
+
+@Entity({ schema: 'public' })
+export class Molecule {
```
LachlanStuart: What happened to the `inchikey` column?

metaspace/graphql/src/modules/moldb/model.ts
```diff
@@ -0,0 +1,43 @@
...
+    mol_id: string;
+
+    @Column({ type: 'text' })
+    mol_name: string;
+
+    @Column({ type: 'text' })
+    formula: string;
+
+    @ManyToOne(type => MolecularDB)
+    moldb: MolecularDB;
```
LachlanStuart: Please add the associated `moldbId` integer-typed column as well.

metaspace/graphql/src/modules/dataset/controller/Mutation.ts
intsco: I think at the moment we don't have databases with the same name. I expect this code to change in the following PRs, since we would either want to check databases by id or use versions as a part of the database "name" used within graphql.

metaspace/graphql/src/modules/moldb/model.ts
intsco: I don't completely understand what type of constraint you actually mean. Now, instead of composite primary key we have a integer `id` key that uniquely identifies a molecule across all databases.

metaspace/graphql/src/modules/moldb/model.ts
intsco: I actually forgot about it. I'm still not sure how we'll be generating molecule images, using inchi, smiles or even something else, `inchikey` by itself was not particularly useful though.

Issue #227, opened 2019-02-13
Add linting to enforce code style in sm-graphql


LachlanStuart:
Consistent code style makes it easier for both newcomers and regular coders to understand the codebase. It also helps to identify some classes of bugs, and reduces the distraction of "ugly code" while working on existing code.

There are tools available that will not only enforce consistent code style, but also automatically fix existing code. In addition there are several pre-made code style guides that we could copy & tweak.

For TypeScript/JavaScript, ESLint is the most common, and is built in to Vue CLI (#211). Recent plugins to it allow it to support TypeScript.

For Python there are many options including Pylint and Black, although some of them enforce an opinionated code style, so research is needed before choosing one.

Ideally any linter we implement should be configured based on our existing code style, though one exception has been identified:
* In JS, there is a general consensus that removing semicolons would make it easier to switch between JS and Python

Ren22:
sounds good, shall we also refactor/rewrite the rest of components in web-app into TS as part of this issue? If that's the case I'd be happy to have my hands on it.

LachlanStuart:
@Ren22 Preferably not as part of this issue. ESLint should be able to apply the same rules to both TS and JS, so there shouldn't be a need for it. The type of code issues that ESLint finds are usually very self-contained, so I wouldn't expect any major refactoring to be needed.

On the JS side, I envision ~50% of this task being configuring ESLint to get it targeting our existing code style, and the other ~50% manually fixing code that ESLint doesn't have an automatic fix for.

On the Python side, I envision significantly more time evaluating linters & how well they work with PyCharm and our codebase. Unlike WebStorm, PyCharm doesn't have built-in support for any linters, so it's still unknown whether a linter would fit well into our workflow.

TBH, I'm still on the fence about the value of TypeScript in webapp. Every interaction with Vue forces data to go through an untyped interface, causing TypeScript to miss a huge proportion of issues. I don't think it's worth the time to do a full rewrite - I'd prefer to keep going on a convert-as-needed basis.

PR #518, opened 2020-01-30
Add eslint to webapp

LachlanStuart:
I evaluated the [Standard](https://github.com/standard/standard) and [Airbnb](https://github.com/airbnb/javascript) presets. Standard seemed to be the least objectionable. With Airbnb I gave up after adding ~20 global overrides against rules that seemed to cost time but not actually improve the code. Standard just needed a few cosmetic tweaks.

I did not evaluate Prettier. I'm not a fan of most of the default "opinionated"-type rules that Prettier uses (no trailing commas, 80-col width, putting operators at the ends of lines when a line break is needed), so it seemed like a bad fit even if most of those rules could be overridden.

As we [discussed a long long time ago](https://github.com/metaspace2020/metaspace/issues/227), I removed semicolons.

I ran across some noteworthy issues:
* Our test in the `VisibilityOptionsSection` was actually wrong as it didn't mention people in a dataset's groups/projects being able to see private datasets. I fixed that
* Two eslint rules had unresolvable issues:
    * `no-unused-vars` complains about imported TypeScript types
    * `import/no-duplicates` broke code when trying to merge import statements
* The `curly` rule's automatic fixes would squash blocks that didn't have `{}`s into a single line, e.g. 
    ```
        if (foo)
            bar()
   ```
    became
    ```
        if (foo) { bar() }
    ```
    These changes often broke `max-len` and always reduced readability. I put them in a separate commit as I had to go through them all and make them multi-line again.
*  eslint has no rule for ensuring consistent indentation in multi-line expressions. Someone suggested it as an extension to the `indent` rule, but they closed the issue because nobody was interested in implementing it. This means this code still unfortunately passes lint:
    ```
   const foo = a
   + b
        + c
    + d
    ```

This PR would be best reviewed commit-by-commit, skipping the "Automatic eslint fixes" commit.


metaspace/webapp/vue.config.js
```diff
@@ -1,45 +1,66 @@
...
+  chainWebpack: config => {
+    config.module
+      .rule('vue')
+      .use('vue-loader')
+      .loader('vue-loader')
+      .tap(options => ({
+        ...options,
+        compilerOptions: {
+          ...options.compilerOptions,
+          whitespace: 'preserve',
```
LachlanStuart: This fixes some small whitespace issues introduced by adding vue-cli.

intsco:
Is there a particular reason to prefer
```
if (foo)
    bar()
```
over
```
if (foo) {
    bar()
}
```
?

PR #522, opened 2020-01-31
Isobar tweaks

### LINKED ISSUE #497: Isobar tweaks
- [x] Add relative m/z to headings in Molecules section
- [x] Add New Feature popup

LachlanStuart:
### New feature popup (only shows up once you've selected an annotation that shows the warning icon)
image

Wording:
> Isomers and isobars
> METASPACE now warns when there are isomeric and isobaric annotations.
>
> When you see this icon , it means that that there are multiple candidate annotations for this set of peaks. Open the Molecules section for more information.
>
> If isobaric annotations have been identified, the Diagnostics section allows comparisons between overlapping annotations to help you to determine which annotation is more likely. The ion formulas of isomers and isobars are also available in the CSV export.

### Isobars how show mass difference in the "Molecules" page
image

### Diagnostic ion images now show an indicator against the peak that overlap
image

### Behind the scenes
* If an annotation is missing ion images for some of its peaks, those peaks won't count toward isobar detection. In extreme cases it was previously possible for two annotations to be declared as isobars, but not actually have any overlapping signal.
* (Unrelated change) Print the error if something goes wrong during the moldb migration

Resolves #497

### DIFF
--- a/metaspace/engine/.pylintrc
+++ b/metaspace/engine/.pylintrc
@@ -276,7 +276,7 @@ function-naming-style=snake_case
 #function-rgx=
 
 # Good variable names which should always be accepted, separated by a comma.
-good-names=_,i,j,f,k,n,a,e,m,v,h,w,x,db,df,ds,db,id,es,fn,fp,on,xs,Run,logger,parser,args,sm_config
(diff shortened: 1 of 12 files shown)


LachlanStuart:
@theodev Please review the screenshots & wording in the new feature popup

PR #525, opened 2020-02-10
Fixes for "Datasets containing molecule" filter

### LINKED ISSUE #432: Fixes for dataset molecule search
- [x] It is currently limited to 10 results - it shouldn't be limited
- [x] It is currently quite slow (2-3 seconds per search) - either fix or mitigate it
    * The input isn't debounced, meaning that typing in the search box causes requests to stack. This probably contributes to the lag as well.
    * There's no loading spinner on the datasets list, so it's not clear that the search is still running
    * All queries are run 3/4 times due to the status filters - possibly ElasticSearch can't easily filter out the empty statuses on the aggregate searches...
- [x] The MSSetupSummary graph doesn't show any data, because its threshold is set too high

LachlanStuart:
Resolves #432 

This is a collection of tweaks around the datasets list to improve performance/UX. The biggest cause of performance problems with the "datasets containing molecule" filter was that typing into the filter box would trigger ~6 graphql requests per keystroke, each taking about 400ms. These would queue up, meaning it would sometimes take 5+ seconds for the datasets list to update. During this time there was no indicator that anything was happening.

Changes:

* Consolidate the API calls into a single pair of `allDatasets` and `countDatasetsPerGroup` calls instead of having a call to `allDatasets` and `countDatasets` with filters for each of the 3/4 visible statuses.
    * In order to replicate the original status-based ordering, I've added a new column to the database: `Dataset.status_update_dt` - the datetime of the last status update for the dataset.
* Debounce & throttle filter changes so that requests are less likely to stack up
* Add a loading spinner to the datasets list
* Tidy up the summary plots, mainly the one for instrument types & MALDI matrices so that it can be used for checking "which instruments/matrices can detect this molecule".
* Change the structure of the ElasticSearch aggregation query so that it wouldn't limit itself to 10 results

### DIFF
--- a/metaspace/engine/sm/engine/es_export.py
+++ b/metaspace/engine/sm/engine/es_export.py
@@ -113,7 +113,7 @@ class ESIndexManager:
             indices = self._ind_client.get_alias(name=alias)
         except NotFoundError:
             indices = {}
-        logger.warning(f'Could not find ElasticSearch alias "{alias}"')
+            logger.warning(f'Could not find ElasticSearch alias "{alias}"')
 
         index = next(iter(indices.keys()), None)
         if len(indices) > 1:
--- a/metaspace/engine/sm/engine/es_export.py
+++ b/metaspace/engine/sm/engine/es_export.py
@@ -67,6 +67,7 @@ FROM (
     d.input_path AS ds_input_path,
     d.upload_dt AS ds_upload_dt,
     d.status AS ds_status,
+    d.status_update_dt as ds_status_update_dt,
     to_char(max(job.finish), 'YYYY-MM-DD HH24:MI:SS') AS ds_last_finished,
     d.is_public AS ds_is_public,
     d.config #> '{databases}' AS ds_mol_dbs,
--- a/metaspace/engine/tests/test_colocalization.py
+++ b/metaspace/engine/tests/test_colocalization.py
@@ -44,7 +44,14 @@ def mock_get_ion_images_for_analysis(storage_type, img_ids, **kwargs):
 
 def test_new_ds_saves_to_db(test_db, metadata, ds_config):
     db = DB()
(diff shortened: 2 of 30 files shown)


metaspace/engine/sm/engine/es_export.py
```diff
@@ -112,7 +113,7 @@ def internal_index_name(self, alias):
             indices = self._ind_client.get_alias(name=alias)
         except NotFoundError:
             indices = {}
-        logger.warning(f'Could not find ElasticSearch alias "{alias}"')
+            logger.warning(f'Could not find ElasticSearch alias "{alias}"')
```
LachlanStuart: Unrelated fix - this log message somehow lost its indentation and it would always warn even when nothing was wrong

metaspace/engine/tests/test_colocalization.py
```diff
@@ -44,7 +44,14 @@ def mock_get_ion_images_for_analysis(storage_type, img_ids, **kwargs):
...
     db = DB()
-    ds = Dataset('ds_id', 'ds_name', 'input_path', datetime.now(), metadata, ds_config)
+    ds = Dataset(
+        id='ds_id',
+        name='ds_name',
+        input_path='input_path',
+        upload_dt=datetime.now(),
+        metadata=metadata,
+        config=ds_config,
+    )
```
LachlanStuart: No effective change here, it was just hard to tell if this needed to be changed for the new `Dataset` argument, so I switched to using named arguments

LachlanStuart:
@richardgoater Please review the webapp changes. Let me know if you want to walk through this together.

richardgoater:
@LachlanStuart would be good to go through together, I don't have very much context for the changes. I'm quite interested in the throttles and debounces, would be good to cover those.

PR #533, opened 2020-02-12
Fix & re-enable graphql tests in CircleCI

LachlanStuart:


metaspace/graphql/src/modules/webServer/storageServer.ts
```diff
@@ -141,7 +141,8 @@ function imageProviderFSBackend(storageRootDir: string) {
           cb(null, dest);
         }
         catch (e) {
-          logger.warn(e);
+          logger.error(e);
+          cb(e, '');
```
intsco: Just curious, why do we need to call this callback on exception?

LachlanStuart: Everything in the Express ecosystem uses callbacks instead of promises for async operations. If a function doesn't call the callback, the caller doesn't know it has finished. The caller generally won't even catch thrown exceptions - instead they bubble up and node gives a console error about the unhandled exception.

Because of this, not calling the callback will almost always result in Express not responding to the HTTP request, usually resulting in a browser-side timeout.

metaspace/graphql/yarn.lock
```diff
@@ -2859,15 +2859,10 @@ got@^6.7.1:
...
-  integrity sha512-IItsdsea19BoLC7ELy13q1iJFNmd7ofZH5+X/pJr90/nRoPEX0DJo1dHDbgtYWOhJhcCgMDTOw84RZ72q6lB+Q==
-
-graceful-fs@^4.1.2, graceful-fs@^4.1.6:
-  version "4.1.11"
-  resolved "https://registry.yarnpkg.com/graceful-fs/-/graceful-fs-4.1.11.tgz#0e8bdfe4d1ddb8854d64e04ea7c00e2a026e5658"
-  integrity sha1-Dovf5NHduIVNZOBOp8AOKgJuVlg=
+graceful-fs@^4.1.11, graceful-fs@^4.1.15, graceful-fs@^4.1.2, graceful-fs@^4.1.6:
+  version "4.2.3"
+  resolved "https://registry.yarnpkg.com/graceful-fs/-/graceful-fs-4.2.3.tgz#4a12ff1b60376ef09862c2093edd908328be8423"
+  integrity sha512-a30VEBm4PEdx1dRB7MFK7BejejvCvBronbLjht+sHuGYj8PHs7M/5Z+rt5lw551vZ7yfTCj4Vuyy3mSJytDWRQ==
```
LachlanStuart: Just for some context why this was an issue - graceful-fs actually patches the node `fs` module. It had code something like this:
```
var fs$WriteStream = fs.WriteStream; // reference to original node implementation
fs.WriteStream = function WriteStream() { // patched implementation
    ...
    return fs$WriteStream.apply(this, arguments);
}
```

The two different versions of graceful-fs were both loaded, and one of them got its `fs$WriteStream` pointing its own implementation, so when something called `fs.WriteStream` it got caught in infinite recursion until it had a JS error.

PR #549, opened 2020-02-24
CircleCI fix & other tweaks

LachlanStuart:
* Fix intermittent error on CircleCI by removing the auto-mocked resolver for graphql Subscriptions. The previous behavior was that the subscriptions would seem to receive some mock data at a random time point during the DatasetList tests, which sometimes caused the snapshots to fail depending on whether the data came before or after the snapshot rendered. With the resolvers removed they now don't get data at all. 
    * Some more work here will be needed if we want to test the subscription resolvers, but we don't currently test them so I just left it as-is.
* Change unit tests to use the same GraphQL client cache configuration as the production code. 
* Adds graphql to codecov so that we can track the stats more easily
* Bumps webapp's minimum coverage up to 50% (it's currently sitting on 50.17%), but increased the allowable amount of coverage drop to 0.5% so that codecov is less complainey.
* Change the development docker nginx config to use variables for `proxy_pass`. When explicit hosts are used, nginx caches connection failures and won't attempt reconnect. This can result in getting `502 Bad Gateway` errors against graphql, even after restarting & successfully compiling it, forcing you to restart nginx to clear the cache. Using variables prevents nginx from caching server status.

### DIFF
--- a/docker/nginx/config/sites-enabled/default
+++ b/docker/nginx/config/sites-enabled/default
@@ -45,36 +45,44 @@ server {
 #    }
 
     location /fs/iso_images/ {
-        proxy_pass http://sm-graphql:4201/fs/iso_images/;
+        set $graphql_storage sm-graphql:4201;
+        proxy_pass http://$graphql_storage$request_uri;
         limit_except GET {deny all;}
     }
     location /db/iso_images/ {
-        proxy_pass http://sm-graphql:4201/db/iso_images/;
+        set $graphql_storage sm-graphql:4201;
+        proxy_pass http://$graphql_storage$request_uri;
         limit_except GET {deny all;}
     }
     location /fs/optical_images/ {
-        proxy_pass http://sm-graphql:4201/fs/optical_images/;
+        set $graphql_storage sm-graphql:4201;
+        proxy_pass http://$graphql_storage$request_uri;
         limit_except GET {deny all;}
     }
     location /fs/raw_optical_images/ {
-        proxy_pass http://sm-graphql:4201/fs/raw_optical_images/;
+        set $graphql_storage sm-graphql:4201;
+        proxy_pass http://$graphql_storage$request_uri;
         client_max_body_size 50M;
     }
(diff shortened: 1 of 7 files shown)


docker/nginx/config/sites-enabled/default
```diff
@@ -45,36 +45,44 @@ server {
...
     location /graphql {
-        proxy_pass http://sm-graphql:3010/graphql;
+        set $graphql sm-graphql:3010;
+        proxy_pass http://$graphql$request_uri;
         include proxy-params.conf;
     }
     location /api_auth {
-        proxy_pass http://sm-graphql:3010/api_auth;
+        set $graphql sm-graphql:3010;
+        proxy_pass http://$graphql$request_uri;
```
intsco: @LachlanStuart I wonder if there is a difference between this approach and using nginx `upstream` directive.

LachlanStuart: `upstream` certainly looks more configurable. With the right settings it would probably achieve the same effect as with variables. Not sure if there's any advantage either way though...

One related issue I'm aware of is that `proxy_pass http://sm-graphql` will cause nginx to completely fail to start if it cannot resolve `sm-graphql`'s DNS name (e.g. because the container doesn't exist yet). I suspect `upstream` may have the same behavior. Using a variable to hold the host name delays DNS lookup until the URL is requested, so it doesn't cause nginx to fail to start.

richardgoater: Using variables to change behaviour in this way could be seen as a hack, but I see that others have blogged this as a solution.

Not for this PR but I would suggest that a smarter solution might be more useful overall, something like https://docs.traefik.io/. Integrates nicely with Docker too.

LachlanStuart: @richardgoater Almost everything I do with nginx feels like a hack, as if it's the wrong product for our use case. 

Traefik looks awesome! We just need an excuse to change...

PR #551, opened 2020-02-26
[sm-api] Molecular databases endpoint

intsco:
This PR adds endpoints to sm-api for CRUD operations with molecular databases.

As preliminary steps:
- Tests were refactored to use `conftest.py` for the fixtures.
- Transactions support was added to the `db` engine module. So that partial moldb import can be rolled back in case an exception is raised.
- The moldb model was connected with the group model. A scientific group owns molecular databases.
- To make sm-api more modular, the sm-api route handlers file was split into several.


metaspace/engine/tests/test_api_databases.py
```diff
@@ -0,0 +1,170 @@
...
+def fill_db(test_db):
+    db = DB()
+    db.insert(
+        'INSERT INTO graphql.group (id, name, short_name) VALUES (%s, %s, %s)',
+        [(GROUP_ID, 'test-group', 'test-group')],
+    )
+
+    yield
+
+    db.alter('TRUNCATE graphql.group CASCADE')
```
LachlanStuart: Doesn't a fresh DB get made for each test? There shouldn't be any need to clean up this data.

PR #563, opened 2020-03-11
[graphql] Moldb files upload to S3

intsco:
This PR adds both a frontend Vue.js component (using `uppy`) and a server side component (using `@uppy/companion`) for uploading custom database files directly to S3. Database files are uploaded into `config.upload.bucket` (renamed from `config.dataset_upload.bucket`) with the `moldbs` prefix.

Additionally, I had to update postgres docker-compose service to minimize the difference between the CI and local test environments.


metaspace/webapp/src/router.ts
```diff
@@ -2,6 +2,7 @@
 import Vue, { AsyncComponent } from 'vue'
 import VueRouter, { RawLocation } from 'vue-router'
 import AboutPage from './modules/App/AboutPage.vue'
+import UppyUploadPage from './modules/MolecularDatabases/UppyUploadPage.vue'
```
LachlanStuart: Please add `UppyUploadPage` back to `asyncPages` so that Uppy isn't included in the main bundle. Use `webpackChunkName: "UppyUploadPage"` so that it's only loaded when necessary.

PR #568, opened 2020-03-30
imzML download

### LINKED ISSUE #547: imzML Download
Depends on #544

Add the ability to download imzML files.
* feature flag in graphql - when disabled, only admins can download
* graphql API gives a link that's valid for ~30 minutes. In order to force the user to see the license information, send a serialized JSON object that includes the license, a link to EULA, and information about attribution
* python-client support for downloading datasets - always print the license / attribution information on each download

LachlanStuart:
Resolves #547 

Adds feature-flagged functionality to download the input files for datasets.

Details:
* When the feature flag is off, only admins can download. Otherwise, everyone who can see a dataset can download it. 
* Downloading is only implemented for files stored on S3.
* Currently the license is determined by the public/private status of the dataset. I'm not really happy with this - I think it needs further discussion. The UI explaining the license is a big ugly, but there's no sense working on it until we have a concrete plan about licensing.

## Download link
image

#### Download dialog
image


metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -497,65 +525,6 @@ def get_molecular_databases(self, names=None):
             return [d for d in mol_dbs if d['name'] in names]
 
 
-class MolDBClient:
```
LachlanStuart: Removed this code because these endpoints are no longer accessible

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -986,23 +995,6 @@ def get_annotations(self, fdr=0.1, db_name="HMDB-v4", datasetFilter={}):
...
-            }
-        )
-        annotations = pd.DataFrame.pivot_table(
-            pd.DataFrame.from_records([ion(r) for r in results.itertuples()]),
-            index=0,
-            columns=[2, 1],
-            values=3,
-        ).notnull()
-        return annotations
-
```
LachlanStuart: This block of code never runs, because there is an unconditional `return` statement above it. It has been this way since METASPACE v0.7. Based on the commit that changed this, I suspect it was just an oversight and this code wasn't intended to be kept.

metaspace/webapp/package.json
```diff
@@ -22,7 +22,7 @@
     "generate-local-graphql-schema": "cd ../graphql && yarn run gen-graphql-schema",
     "fetch-prod-graphql-schema": "apollo schema:download tests/utils/graphql-schema.json --endpoint https://metaspace2020.eu/graphql",
     "build": "yarn run deref-schema && cross-env NODE_ENV=production vue-cli-service build --modern",
-    "build-ci": "yarn run deref-schema && cross-env NODE_ENV=production vue-cli-service build --modern --no-progress --hide-modules",
+    "build-ci": "yarn run deref-schema && cross-env NODE_ENV=production vue-cli-service build --no-progress --hide-modules",
```
LachlanStuart: Unrelated change - `--modern` doubles the compilation time, but there's no benefit to doing that during CI.

LachlanStuart:
@intsco Please review the python-client and graphql changes.

@richardgoater Please review the webapp changes.

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -9,6 +10,33 @@
...
+
+class DatasetDownloadFile(TypedDict):
+    filename: str
+    link: str
+
+
+class DatasetDownload(TypedDict):
+    license: DatasetDownloadLicense
+    contributors: List[DatasetDownloadContributor]
+    files: List[DatasetDownloadFile]
```
intsco, line 38: This looks like dataclasses introduced in Python 3.7 https://www.python.org/dev/peps/pep-0557/#rationale

LachlanStuart: @intsco `TypedDict` definitions just give type hints for dictionary keys, e.g. `dataset_download['license']['link']` can be determined to be an `Optional[str]`, whereas data classes work like classes (i.e. using `dataset_download.license.link` syntax).

This case is just providing types for something deserialized from JSON, so it's expected to be a `dict` instance. Although I'd like to use data classes, last time I checked, there wasn't actually an easy way to deserialize JSON into a data class. Also, the need to upgrade Python is a bit of a deal breaker...

metaspace/webapp/package.json
richardgoater: We can probably trust the tools, but if we leave this in do we get more reassurance that we are testing the same code that we're deploying?

metaspace/webapp/src/modules/App/element-variables.scss
```diff
@@ -37,4 +37,11 @@ $--collapse-header-height: 40px;
   @include e(body) {
     word-break: normal;
   }
+
+  // Use `<el-dialog customClass="el-dialog-lean">` to remove excessive padding above/below body
+  &.el-dialog-lean {
```
richardgoater, line 42: I think there will be a future discussion point here, I have changed this file a lot as part of integrating the new Element UI theme.

metaspace/webapp/src/modules/Datasets/list/DatasetItem.vue
```diff
@@ -172,6 +172,17 @@ class="s-group ds-add-filter"
         </router-link>
       </div>
 
+      <div
```
richardgoater, line 175: I've been editing this section as well. This is nicely showing the lack of space that justifies a page for datasets :)

metaspace/webapp/package.json
LachlanStuart: No. This actually isn't used for anything more than checking "does it build?". All the tests are run on unbuilt code.

richardgoater: oh right, nice 👍

metaspace/webapp/src/modules/Datasets/list/DatasetItem.vue
LachlanStuart: Yeah, it's also alarmingly slow to render this page because of how much complexity is in each `DatasetItem`. It will be great to be able to trim this component down.

PR #585, opened 2020-05-07
White list annotations for small custom databases

### LINKED ISSUE #576: Add migration for data updates to existing databases
Convert hidden databases into custom databases that belong to the EMBL group.

Copy additional metadata (fullName, link, citation, description) for public databases from [DatabaseDescriptions.vue](https://github.com/metaspace2020/metaspace/blob/feat/custom-moldbs/metaspace/webapp/src/modules/MetadataEditor/inputs/DatabaseDescriptions.vue) into the database.

intsco:
The key idea of this PR is to add white listing of small custom databases (<1k formulas) so that for these databases annotations are kept independently of their FDR. But it turned out to be mostly about refactoring engine and graphql to address databases by id instead of name.

* `sm.engine.molecular_db` - separate model class and CRUD-functions
* replaced `ds_config.databases` with `ds_config.database_ids` in the database and ES
* graphql: deprecate and replace type fields
`Annotation.database` -> `Annotation.databaseDetails`
`Dataset.molDBs` -> `Dataset.databases`
* graphql: deprecate and replace filters and input fields
`Dataset{Create,Update}Input.molDBs` -> `Dataset{Create,Update}Input.databaseIds`
`AnnotationFilter.database` -> `AnnotationFilter.databaseId`

The rest of the changes in the code are related to the required refactoring or just updates done on the way.

Resolves #576


metaspace/graphql/src/modules/dataset/controller/Mutation.ts
```diff
@@ -212,48 +199,54 @@ type CreateDatasetArgs = {
...
-    // TODO: Many of the inputs are mistyped because of bugs in graphql-binding that should be reported and/or fixed
-    await molDBsExist(ctx.entityManager, input.molDBs as any || []);
-  }
 
-  const {submitterId, groupId, projectIds, principalInvestigator} = input;
-  const saveDSArgs = {
+  await setDatabaseIdsInInput(ctx.entityManager, input);
+
+  // Only admins can specify the submitterId
+  const submitterId = (ctx.isAdmin && input.submitterId) || (dataset && dataset.userId) || ctx.user.id;
```
intsco, line 233: I hope this line still behaves the same way.

richardgoater: looks ok to me, no complaints from TypeScript?

intsco: Nope, everything looks like it just should work but with JS/TS I'm never completely confident :)

metaspace/graphql/src/modules/annotation/queryFilters/colocalizationSamples.ts
```diff
@@ -5,10 +5,10 @@ import config from '../../../utils/config';
 import * as _ from 'lodash';
 import {setOrMerge} from '../../../utils/setOrMerge';
 
-const getColocSampleIons = async (context: Context, datasetId: string, fdrLevel: number, database: string | null,
+const getColocSampleIons = async (context: Context, datasetId: string, fdrLevel: number, databaseId: number | null,
                                   colocalizationAlgo: string) => {
   const result = await context.entityManager.findOne(ColocJob,
-    { datasetId, fdr: fdrLevel, molDb: database, algorithm: colocalizationAlgo },
+    { datasetId, fdr: fdrLevel, molDb: databaseId != null ? databaseId.toString() : null, algorithm: colocalizationAlgo },
```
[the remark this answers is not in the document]
intsco: Yeah, that's a good catch :) It's actually a temp solution before I add a migration for the `ColocJob.molDB` field. I'll do it in this PR.

metaspace/graphql/src/modules/annotation/queryFilters/queryFilters.spec.ts
```diff
@@ -61,7 +64,7 @@ describe('annotation/queryFilters applyQueryFilters (colocalization)', () => {
     const argsWithColocWith: QueryFilterArgs = {
       datasetFilter: { ids: job.datasetId },
       filter: {
-        database: job.molDb!,
+        databaseId: Number(job.molDb!),
```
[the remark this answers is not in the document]
intsco: The same reason, should be removed after the migration is ready.

metaspace/graphql/src/modules/dataset/controller/Dataset.ts
```diff
@@ -114,8 +116,16 @@ const DatasetResolvers: FieldResolversFor<Dataset, DatasetSource> = {
     return ds._source.ds_is_public;
   },
 
-  molDBs(ds) {
-    return ds._source.ds_mol_dbs;
+  async databases(ds, _, ctx): Promise<MolecularDB[]> {
+    return await Promise.all(
+      ds._source.ds_moldb_ids.map(async (databaseId) => (await getMolecularDbModel(ctx, databaseId)))
```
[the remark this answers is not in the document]
intsco: That's a fair point. According to @LachlanStuart, the use of [DataLoader](https://github.com/graphql/dataloader) in `getMolecularDbModel` should prevent a performance degradation.

metaspace/webapp/.gitignore
```diff
@@ -20,3 +20,5 @@ tests/e2e/screenshots
 
 # Build output
 /dist
+/src/assets/ims.json
+/src/assets/metadataMapping.js
```
LachlanStuart, line 24: These files should not exist in this location anymore. They should be generated into `src/lib/metadataSchemas`. Any idea what created them? I can't see anything in code that would re-create these files in `src/assets`.

metaspace/engine/sm/engine/msm_basic/msm_basic_search.py
```diff
@@ -70,32 +74,67 @@ def gen_ion_formulas(args):
...
+) -> Tuple[pd.DataFrame, pyspark.RDD]:
+    """Compute FDR for database annotations and filter them."""
 
-    formula_images_rdd = results_rdd.flatMap(lambda t: t[1].items())
-    return formula_metrics_df, formula_images_rdd
+    moldb_formula_map_df = ion_formula_map_df[ion_formula_map_df.moldb_id == moldb.id].drop(
+        'moldb_id', axis=1
+    )
+    moldb_metrics_fdr_df = compute_fdr(fdr, formula_metrics_df, moldb_formula_map_df)
+    max_fdr = 1.0 if moldb.targeted else 0.5
```
LachlanStuart, line 129: We may have covered this in a previous discussion, though I don't recall coming to a consensus on how to handle it.

Our FDR calculation can produce values higher than 1.0, and I think we should keep those annotations unless we have a good reason not to.

I realize that having FDR values like 250% doesn't make intuitive sense. As far as I can tell, we have implemented the FDR algorithm correctly. It's just that the FDR algorithm we've implemented doesn't take into account the statistical error of random sampling. I.e. A reported FDR 250% might be explained by having a margin of error of +/- 200%.

metaspace/graphql/src/modules/annotation/model.ts
```diff
@@ -15,8 +15,8 @@ export class ColocJob {
   @Column({ type: 'text', name: 'ds_id' })
   datasetId: string;
 
-  @Column({ type: 'text', nullable: true })
-  molDb: string | null;
+  @Column({ type: 'int', nullable: false })
```
LachlanStuart, line 18: `nullable: false` is the default - no need to specify it explicitly.

I can't see how this generated a foreign key in the migration. Was there a `@ManyToOne` annotation here that didn't get committed?

metaspace/engine/migrations/r1_8_20200513_moldb_id_as_identifier.py
```diff
@@ -0,0 +1,238 @@
...
+def migrate_moldbs():
+    update_public_database_descriptions()
+    update_non_public_databases()
+
+    moldb_name_id_map = build_moldb_map()
+    moldb_name_id_map_rev = {v: k for k, v in moldb_name_id_map.items()}
+
+    datasets = DB().select_with_fields('SELECT id, config FROM dataset')
+    failed_datasets = []
+    for ds_doc in datasets:
```
LachlanStuart, line 217: Is this script idempotent? I.e. if it fails in the middle of the migration, can we just restart it and still get the correct output?

I think the answer is yes, but I just wanted to double-check that you have the same view. Random errors happen fairly often with ES.

intsco: Yes, I designed it to be idempotent and locally it works fine but I'll need to test it on staging. It doesn't pick up where it left but I hope it should be fast enough.

metaspace/engine/sm/engine/msm_basic/msm_basic_search.py
intsco: I didn't find any annotations with `fdr > 1` on staging and I feel like this change needs to be introduced in a different PR.

metaspace/graphql/src/modules/annotation/queryFilters/queryFilters.spec.ts
```diff
@@ -70,21 +75,23 @@ describe('annotation/queryFilters applyQueryFilters (colocalization)', () => {
...
     expect(args).toMatchObject({
       offset: 0,
       limit: 1000,
       filter: {
+        databaseId: database.id,
         ion: ions.slice(1,4).map(ion => ion.ion),
       } as any
     });
+    args.filter!.databaseId = 1;
+    expect(args).toMatchSnapshot();
```
intsco, line 87: I didn't find a more elegant way to resolve the above issue.

metaspace/graphql/src/modules/annotation/queryFilters/queryFilters.spec.ts
```diff
@@ -37,16 +39,19 @@ describe('annotation/queryFilters applyQueryFilters (colocalization)', () => {
     })));
     await testEntityManager.insert(Ion, ions);
 
+    database = testEntityManager.create(MolecularDB, {id: 0, name: 'HMDB-v4', version: 'version'});
+    await testEntityManager.insert(MolecularDB, database);
```
intsco, line 43: @LachlanStuart TypeORM (or Postgres <10) doesn't support explicit assignment of an auto increment column. So after insert `database.id` contains a value > 0.

intsco:
@LachlanStuart I don't know why the test coverage step in graphql CI pipeline started failing. But the webapp one should be fixed by @richardgoater in the next PR.

metaspace/graphql/src/modules/dataset/controller/Mutation.ts
richardgoater: Nobody is 👍

PR #588, opened 2020-05-13
Upgrade TypeScript & other deps

LachlanStuart:
Webapp dependency updates:
* TypeScript 3.6.5 -> 3.9.2
* ElementUI 2.12.0 -> 2.13.1 (needed a TypeScript fix from 2.13.0)

Graphql dependency updates:
* TypeScript 3.6.2 -> 3.9.2

Graphql dependency updates to resolve security alerts:
* @sentry/node 5.4.0 -> 5.15.5
* knex 0.19.5 -> 0.20.15 (cannot go to 0.21 due to node 8 EOL :disappointed: )
    * Note that knex is now only used for tests and the unused DB versions of the image storage
* jest 24.9.0 -> 25.5.4 (cannot go to 26 due to node 8 EOL :disappointed: )
* ts-jest 23.1.3 -> 25.5.0
* nodemon 1.19.1 -> uninstalled, as we've been using ts-node-dev instead

Due to [changes in jest behavior](https://github.com/facebook/jest/pull/8143), `globalSetup`/`globalTeardown` had to be converted to TS to prevent a type-checking error from double-transpilation. [There's a Jest bug](https://github.com/kulshekhar/ts-jest/issues/1274#issuecomment-605099634) causing TypeORM migrations to not be transformed, so I added a workaround that imports everything as usual (via the ts-jest transform), but registers ts-node just before TypeORM runs the migrations, so that they're transformed via ts-node. 

Also, I wasn't able to find the root cause, but after these changes it was much more common for the first test to timeout during `setupTestFramework`'s `beforeAll` or `beforeEach`, causing entire test suites to randomly fail. I strongly suspect that `beforeAll` wasn't previously included in the timeouts, as it would regularly take 9-11s to run this function in CircleCI. I bumped the `testTimeout` from 5s to 15s to resolve this.

Noteworthy changelogs:
* TypeScript [3.7](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#37-breaking-changes) [3.8](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html) [3.9](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#breaking-changes)
    * Some small tweaks were needed in test files to fix compilation due to breaking changes
    * Exciting new feature: [`?.` optional chaining operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining)
* [ElementUI](https://github.com/ElemeFE/element/blob/dev/CHANGELOG.en-US.md)


metaspace/graphql/src/tests/globalSetup.ts
```diff
@@ -1,10 +1,9 @@
...
-const config = require('../utils/config').default;
-const {createConnection} = require('../utils/db');
-const {DbSchemaName} = require('../utils/typeOrmConfig');
+import * as Knex from 'knex';
+import config from '../utils/config';
+import {createConnection} from '../utils/db';
+import {DbSchemaName}  from '../utils/typeOrmConfig';
 
-module.exports = async () => {
+export = async () => {
```
richardgoater, line 6: this is new syntax for me, quickly looked up and found some old discussion here: https://github.com/microsoft/TypeScript/issues/7185

is this for commonJS compat?

LachlanStuart: Yes. `module.exports =` and `export =` are functionally equivalent. Both would be imported with `globalSetup = require(...)` / `import * as globalSetup from '...'`. In fact, TypeScript allows both forms of export, I just figured it would be better to match syntax with the rest of the file.

I thought it was an ES Modules thing, not a TypeScript thing, but I could be mistaken...

metaspace/webapp/package.json
```diff
@@ -42,7 +42,7 @@
     "d3": "^5.11.0",
     "date-fns": "^1.29.0",
     "dom-to-image-google-font-issue": "^2.6.2",
-    "element-ui": "^2.12.0",
+    "element-ui": "^2.13.1",
```
richardgoater, line 45: I'm still bearing the static CSS in mind, might need to change it sooner now. We might be missing a few fixes from this version, but it would take a lot of fixes to fix Element! 🙂

LachlanStuart: I'm actually more concerned that ElementUI's updates will include intentional style changes that will break our layout, than concerned that our old static CSS will stop being compatible.

Fortunately not much has changed: https://github.com/ElemeFE/element/compare/v2.12.0...v2.13.1

The only component with style changes that I'm concerned about is [`ElTable`](https://github.com/ElemeFE/element/compare/v2.12.0...v2.13.1#diff-62805ded38eb1a3547190f2d31771c12), as those tweaks could potentially change our AnnotationsTable layout a bit.

metaspace/graphql/src/tests/globalSetup.ts
LachlanStuart: MDN and Node's docs don't have any mention of it, so I guess it is a TypeScript thing after all.

PR #590, opened 2020-05-15
Allow project description to be edited throughout publishing process

richardgoater:
Small error in API validation

### DIFF
--- a/metaspace/graphql/src/modules/project/Publishing.ts
+++ b/metaspace/graphql/src/modules/project/Publishing.ts
@@ -16,25 +16,27 @@ export function validatePublishingRules(ctx: Context, project: ProjectSource, pr
     return
   }
 
-  if (project.publicationStatus == PSO.UNDER_REVIEW
-    && project.urlSlug != null
-    && projectDetails.urlSlug == null) {
-    throw new FormValidationErrors(
-      'urlSlug',
-      `Cannot remove short link as the project is under review`
-    )
+  if ('urlSlug' in projectDetails) {
+    if (project.publicationStatus === PSO.UNDER_REVIEW
+      && project.urlSlug !== null
+      && projectDetails.urlSlug === null) {
+      throw new FormValidationErrors(
+        'urlSlug',
+        `Cannot remove short link as the project is under review`
+      )
+    }
+
+    if (project.publicationStatus === PSO.PUBLISHED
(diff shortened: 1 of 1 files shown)


metaspace/graphql/src/modules/project/Publishing.ts
```diff
@@ -16,25 +16,27 @@ export function validatePublishingRules(ctx: Context, project: ProjectSource, pr
...
-    && projectDetails.urlSlug == null) {
-    throw new FormValidationErrors(
-      'urlSlug',
-      `Cannot remove short link as the project is under review`
-    )
-  }
+  if ('urlSlug' in projectDetails) {
+    if (project.publicationStatus === PSO.UNDER_REVIEW
+      && project.urlSlug !== null
+      && projectDetails.urlSlug === null) {
```
LachlanStuart, line 22: It's generally best to default to `==`/`!=` when comparing to `null`. IIRC input variables in graphql requests can be either `null` or `undefined` depending on whether the variable was omitted or explicitly set to null. Generally they should be treated the same.

`!==` / `===` is of course preferable when comparing against anything that isn't `null`.

richardgoater: In this case I at least wanted to check that `projectDetails.urlSlug` is definitely `null`, as I thought this is how the `urlSlug` is removed. I suppose the `in` check will remove the `undefined` case from the inner block. As mentioned on Slack, I usually default to never using coercion on equality checks for explicitness.

richardgoater: Struck out the above, I see that the key could be provided `undefined`.

richardgoater: ... or can it? `undefined` is not valid in JSON 🤷

LachlanStuart: Ah, sorry, my mistake. Friday afternoon brain. I didn't connect this to the Slack conversation. You're right that it would make more sense looking for explicit `null`s here.

You're right that `undefined` isn't valid JSON, but missing fields still evaluate to `undefined`. I've no idea whether Apollo preserves the original input JSON, or does some normalization to ensure all fields are present, causing `urlSlug` to be defined but have an `undefined` value...

PR #592, opened 2020-05-18
feat(browser): PoC implementation of dataset browser

intsco:
The PR has been refactored, multiple bugs were fixed, the prototype was extended and can be considered finished.

The main building blocks of the prototype:
* **Dataset preprocessing**: read _imzml/ibd_ files from S3, segment and sort all dataset peaks as (sp_idx, mz, intensity) by mz. The new peaks sorted by mz, mz index, and coordinates files are stored at the dataset's S3 path.
* **MZ image search**: read a dataset mz index, find and download needed chunk(-s) of data, extract mz image from the downloaded data. To speed up consequent mz searches within the same dataset, the last 128 dataset indices are cached in memory.
* **API** based on FastAPI + Uvicorn: start a web server that exposes two endpoints, _/preprocess_ and _/search_.

The _requirements.txt_ file contains a minimal set of dependencies. In order to read/write files in S3, user credentials should be provided to boto using one of the supported mechanisms, e.g. in `~/.aws/credentials`.

P.S. Unfortunately, the code severely lacks tests and documentation.

@LachlanStuart Please take a look at the whole PR once again or just the [recent changes](https://github.com/metaspace2020/metaspace/pull/592/files/3ee876ba5c86c7f4a84f04688438d01126399088..bd12cf8032ef23e140ef36e6a564ba90633a1769).


intsco:
@LachlanStuart please just take a look at the core concepts. No detailed review is required yet.

metaspace/browser/sm/browser/main.py
```diff
@@ -0,0 +1,147 @@
...
+        start = time.time()
+        log(start, "Building index:")
+
+        log(start, f"downloading dataset {self.dataset_s3_path} to {self.ds_local_path}")
+        utils.download_dataset(self.bucket, self.ds_s3_path, self.ds_local_path)
+
+        log(start, f"parsing imzml in {self.ds_local_path}")
+        ibd_stream = self.ibd_local_path.open("rb")
+        self.imzml_reader = split_sort.ImzMLReader(self.imzml_local_path)
+        self.imzml_reader.add_stream(ibd_stream)
```
LachlanStuart, line 59: Not sure what `add_stream` does, but I had to add an [`ibd_file` parameter to the `ImzMLReader` constructor](https://github.com/alexandrovteam/pyimzML/blob/master/pyimzml/ImzMLParser.py#L63) for CloudButton that probably does the same thing.

metaspace/browser/sm/browser/main.py
```diff
@@ -0,0 +1,147 @@
...
+
+        log(start, f"segmenting and sorting dataset by mz in {self.segments_path}")
+        ibd_size_mb = self.ibd_local_path.stat().st_size / 1024 ** 2
+        split_sort.segment_dataset(self.imzml_reader, ibd_size_mb, self.segments_path)
+
+        log(start, f"pickling imzml_reader to {self.imzml_reader_path}")
+        self.imzml_reader.remove_stream()
+        pickle.dump(self.imzml_reader, self.imzml_reader_path.open("wb"))
```
LachlanStuart, line 67: Any reason you're not using [`portable_spectrum_reader`](https://github.com/alexandrovteam/pyimzML/blob/master/pyimzml/ImzMLParser.py#L345) here? I'd expect it to be smaller and possibly easier to work with.

metaspace/browser/sm/browser/mz_search.py
```diff
@@ -0,0 +1,99 @@
...
+def create_mz_image(mz_peaks: np.ndarray, coordinates: np.ndarray) -> np.ndarray:
+    min_x, min_y = np.amin(coordinates, axis=0)
+    max_x, max_y = np.amax(coordinates, axis=0)
+    nrows, ncols = max_y - min_y + 1, max_x - min_x + 1
+    mz_image = np.zeros(shape=(nrows, ncols))
+
+    image_coords = coordinates[mz_peaks[:, 2].astype(int)]
+    xs = image_coords[:, 0] - min_x
+    ys = image_coords[:, 1] - min_y
+    mz_image[ys, xs] += mz_peaks[:, 1]
```
LachlanStuart, line 97: Unfortunately this doesn't behave as you'd expect. If multiple values exist for the same pixel, this type of assignment will cause the last one to be selected and the rest to be discarded. e.g.
```python
image = np.zeros((3,3))
xs, ys = np.array([0,0,0,0]), np.array([0,1,0,1])
ints = np.array([1,2,3,5])
image[xs, ys] += ints
image

> array([[3., 5., 0.],    # Would expect [4., 7,. 0.],
>        [0., 0., 0.],
>        [0., 0., 0.]])
```

I'd suggest using `coo_matrix.todense()` if possible. If it's too slow, a Numba njit function would let you efficiently loop through the whole array doing `+=` with each item.

metaspace/browser/sm/browser/main.py
intsco: I had to move it to it's own method so I can add a stream to an ImzmlReader object after unpickling it. While here a local binary stream is used, in `download_index` it's an object of S3File added to ImzmlReader after unpickling.

metaspace/browser/sm/browser/main.py
intsco: `ImzMLReader` stores a `PortableSpectrumReader` in the `spectrum_reader` field.

metaspace/browser/sm/browser/mz_search.py
intsco: Thanks, that's a good catch. I didn't test this extensively so it was pretty easy to overlook issues that don't have an immediate effect. Weird that numpy silently replaces `+=` operator with `=`.

metaspace/browser/sm/browser/main.py
LachlanStuart: I think you may have missed the point of `PortableSpectrumReader`. 

It is meant to be a light-weight class storing only the data needed to read the .ibd file. Once you have a `PortableSpectrumReader`, you throw the `ImzMLReader` away. If you've adapted `ImzMLReader` to be pickleable, there's no point in keeping a copy of the `PortableSpectrumReader`, as it contains only a subset of the data from `ImzMLReader`.

metaspace/browser/sm/browser/mz_search.py
LachlanStuart: It doesn't change it to `=`, it just does something weird so that a maximum of one source element is `+=`'d to each destination index. I suspect that it does something like this:
```
# Original code:
a[xs, ys] += b
# Actual operations
temp1 = np.zeros_like(a)
temp1[xs, ys] = b
a += temp1
```

metaspace/browser/sm/browser/main.py
LachlanStuart: Oh, sorry! I completely mistook `ImzMLReader` for `ImzMLParser`. I thought you had re-implemented this stuff. Ignore my previous comment.

metaspace/browser/sm/browser/mz_search.py
```diff
@@ -0,0 +1,99 @@
...
+            end = self.size - 1
+        range_header = f"bytes={start}-{end}"
+        return self.s3_object.get(Range=range_header)["Body"].read()
+
+
+def search_and_fetch_mz_peaks(
+    stream: BinaryFile, mz_index: np.ndarray, mz_lo: float, mz_hi: float
+) -> np.ndarray:
+    mz_lo_chunk_idx, mz_hi_chunk_idx = np.searchsorted(mz_index, [mz_lo, mz_hi])
+    mz_lo_chunk_idx -= 1  # previous chunk actually includes value
```
LachlanStuart, line 75: I'd suggest using the `side='left'`/`side='right'` logic that's is currently used for annotation just to make sure that this handles it correctly when `mz_lo`/`mz_hi` fall on a chunk boundary, or when multiple chunks have the same m/z value..

metaspace/browser/sm/browser/main.py
intsco: Sorry, I didn't clarify that `ImzMLReader` is just a local wrapper around `PortableSpectrumReader`.

metaspace/browser/sm/browser/mz_search.py
intsco: Yeah, I understand. I meant that effectively `image[ys, xs] += ints` produces the same result as `image[ys, xs] = ints`
```
In[4]: image = np.zeros((3,3))
  ...: xs, ys = np.array([0,0,0,0]), np.array([0,1,0,1])
  ...: ints = np.array([1,2,3,5])
  ...: image[xs, ys] = ints
  ...: image
Out[4]: 
array([[3., 5., 0.],
       [0., 0., 0.],
       [0., 0., 0.]])
```

LachlanStuart:
I haven't looked in depth into the creation of the index/peaks, but the data formats and everything inside `dataset_browser.search_mz` looks good.

metaspace/browser/sm/browser/mz_search.py
intsco: I think I had to add this `-= 1` hack because unlike in engine here the mz index stores only left bound of each "segment". But I'll double check this. I actually planned to add a suit of tests that would cover these cases.

LachlanStuart: Ah, that `-= 1` makes sense. I'm so used to the `hi` index being `+1`/`-1` that I didn't realize this was for the `lo` index.

I think if `hi` was calculated with `side='right'`, the ` + 1` in `bytes_to_read` wouldn't be needed.

metaspace/browser/sm/browser/mz_search.py
intsco:
```
np.add.at(mz_image, [ys, xs], mz_peaks[:, 1])
```
Seems to fix the mentioned issue.

metaspace/browser/sm/browser/mz_search.py
intsco: The current code seems to work correctly even though it might not be optimal. But I'd prefer not to refactor it before we have a set of tests for it.

intsco:
@LachlanStuart What are your thoughts about this PR, shall we finish and merge it into master?

metaspace/engine/sm/engine/__init__.py
```diff
@@ -0,0 +1 @@
+__version__ = '1.7.0'
```
LachlanStuart, line 1: `setup.py` still tries to import this from `engine/sm/__init__.py`

LachlanStuart (approved):
As discussed - code looks good to merge after the issue with the `__version__` is resolved.

PR #594, opened 2020-05-19
Migrate project descriptions

richardgoater:
Original code snippet from here: https://github.com/metaspace2020/metaspace/pull/550#discussion_r412376522


intsco:
@richardgoater I decided to leave this migration in graphql as I don't actually see much value in making it a manual one. Production is already deployed, it seems to work, so we are not rolling the latest changes back.

PR #603, opened 2020-06-04
Paginate dataset list

richardgoater:
Quickly figured this out, no UI changes but allows showing more datasets with `?page=2`.


intsco:
Looks good but I'm not sure those who initially raised the issue will be happy to edit URL parameters to get pagination working.

richardgoater:
No worries, it's just meant to be a quick hack before the UI is
implemented. At least it provides access to data which is currently not
accessible at all.

PR #605, opened 2020-06-05
Custom databases

intsco:
Changes in the configuration and setup:
- [ ] Test updated `docker-compose.yml`
- [ ] Dev installations: update postgres user
`docker-compose exec --privileged postgres psql postgres postgres -c "ALTER USER sm WITH SUPERUSER;"`
- [ ] Update configs
engine: `aws_region` -> `aws_default_region`
ansible/engine `data_path`: `/opt/data/sm_data` -> `/opt/data/metaspace`


intsco:
@LachlanStuart After your PR approval, I'm going to rebase this branch onto master.

LachlanStuart:
Happy for this to be merged once the remaining PRs are merged.

Just so we're clear on what has been deferred to the next iteration, this does not include:
* Annotations with MSM 0.000, or only 1 isotope image in targeted databases
* Reindexing datasets when DBs are renamed or deleted. Though this probably isn't necessary:
    * DB name is used in Datasets' `annotation_counts.db.name`, which is searchable through the free-text search, but I doubt people rely upon it
    * DB name is used in Annotations' `db_name` field, but that only seems to be used in `possibleCompounds` for selecting which template to use for the molecule URL. It's unlikely we'll change the core DB names.
    * Deleted DBs are automatically excluded from search results as they no longer are included in `getVisibleDatabaseIds`
* Anything else?

intsco:
@LachlanStuart 
>Annotations with MSM 0.000, or only 1 isotope image in targeted databases

Correct.

>Reindexing datasets when DBs are renamed or deleted. Though this probably isn't necessary

I don't remember discussing databases renaming functionality. `UpdateMolecularDBInput` doesn't have the `name` field.
Regarding database deletion, it's a good point. It should be available only to admins but we still need to take care of orphan annotations in Postgres and ES after their database was deleted. At least, we should add `ON DELETE CASCADE` to https://github.com/metaspace2020/metaspace/blob/3b4f0cdfca1c77e0853212764631d5ba5ed9a9e2/metaspace/engine/scripts/db_schema.sql#L289:L291.

>DB name is used in Datasets' annotation_counts.db.name, which is searchable through the free-text search, but I doubt people rely upon it
DB name is used in Annotations' db_name field, but that only seems to be used in possibleCompounds for selecting which template to use for the molecule URL. It's unlikely we'll change the core DB names.

I hoped that we'd be able to remove all references to database name and would be able to use only database id everywhere.

>Anything else?

* Database management UI.
* Database management in python-client.
* A bunch of minor updates like adding `assertCanUseDatabase` checks and enforcing unique names for public custom databases.

LachlanStuart:
> I don't remember discussing databases renaming functionality. UpdateMolecularDBInput doesn't have the name field.

Good point. I think we actually did discuss it and agreed to just not support renaming at all.

> I hoped that we'd be able to remove all references to database name and would be able to use only database id everywhere.

I support this, and anything that moves us away from using ElasticSearch for general data storage & retrieval.

intsco:
@LachlanStuart no idea why but after merging master and running `yarn install` in graphql directory `yarn.lock` changed again. Don't know if it's worth investigating.

LachlanStuart:
@intsco The `yarn.lock` changes in the latest commit look fine to me.

PR #606, opened 2020-06-05
[python-client] Custom databases: python-client deprecations

intsco:
Following the graphql API modifications in https://github.com/metaspace2020/metaspace/pull/585, I updated python-client so that it doesn't use the deprecated fields. However, I am not quite happy with the changes as they seem to add even more chaos to the code. There are also some potentially breaking changes in methods where I had to add one more argument.

### DIFF
--- a/metaspace/python-client/metaspace/sm_annotation_utils.py
+++ b/metaspace/python-client/metaspace/sm_annotation_utils.py
@@ -1,13 +1,17 @@
+import json
+import os
+import pprint
+from collections import defaultdict
 from concurrent.futures import ThreadPoolExecutor
-from pathlib import Path
-from shutil import copyfileobj
-from typing import Optional, List
-
-import pandas as pd
-import numpy as np
-import requests, json, os, pprint
 from copy import deepcopy
 from io import BytesIO
+from pathlib import Path
+from shutil import copyfileobj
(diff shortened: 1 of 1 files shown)


intsco:
@LachlanStuart 
>An alternative, less noisy approach would be to continue accepting a single databases parameter, and adding a shared "normalize_databases" function that converts names to IDs if necessary. Such a function would need to make a server call to look up the DBs if it was given string names, and it could provide a useful error message to the user if they gave an ambiguous name...

I like this approach more. We can actually completely avoid problems with ambiguous names if we go this way. The list of databases visible to a non-admin user will always have unique (name, version) pairs.

>If we changed python-client to only use the new API parameters, we could confidently remove the deprecated parameters from the API much sooner.

I didn't completely get this, you mean our life would be easier if we changed python-client to only use the new API parameters but we cannot because our users won't be happy about such an upgrade of python-client?

LachlanStuart:
> you mean our life would be easier if we changed python-client to only use the new API parameters

Yes.

> but we cannot because our users won't be happy about such an upgrade of python-client?

I mean that we can remove the parameters from the graphql API. The Python API could stay backwards-compatible, so users wouldn't notice.

intsco:
@LachlanStuart I've re-implemented it the way discussed above.

PR #615, opened 2020-06-17
Show all datasets when creating project

richardgoater:
I tested this locally by changing the default limit on the server and showing that I could see more if I passed a limit of `-1`, but it might want to be verified with an account that has more than 100 datasets on staging.

Also, I thought that a better default for this list is to start with no items selected rather than all. The "Transfer datasets" dialog should still work in the original way. Fine to talk more about this.

### DIFF
--- a/metaspace/webapp/src/api/dataset.ts
+++ b/metaspace/webapp/src/api/dataset.ts
@@ -125,7 +125,7 @@ export interface DatasetListItem {
 
 export const datasetListItemsQuery =
   gql`query GetDatasets($dFilter: DatasetFilter, $query: String) {
-    allDatasets(offset: 0, limit: -1, filter: $dFilter, simpleQuery: $query) {
+    allDatasets(offset: 0, limit: 10000, filter: $dFilter, simpleQuery: $query) {
       id
       name
       uploadDT
--- a/metaspace/webapp/src/api/dataset.ts
+++ b/metaspace/webapp/src/api/dataset.ts
@@ -125,7 +125,7 @@ export interface DatasetListItem {
 
 export const datasetListItemsQuery =
   gql`query GetDatasets($dFilter: DatasetFilter, $query: String) {
-    allDatasets(offset: 0, limit: 100, filter: $dFilter, simpleQuery: $query) {
+    allDatasets(offset: 0, limit: -1, filter: $dFilter, simpleQuery: $query) {
       id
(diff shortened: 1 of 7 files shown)


metaspace/webapp/src/api/dataset.ts
```diff
@@ -125,7 +125,7 @@ export interface DatasetListItem {
 
 export const datasetListItemsQuery =
   gql`query GetDatasets($dFilter: DatasetFilter, $query: String) {
-    allDatasets(offset: 0, limit: 100, filter: $dFilter, simpleQuery: $query) {
+    allDatasets(offset: 0, limit: -1, filter: $dFilter, simpleQuery: $query) {
```
LachlanStuart, line 128: `limit: -1` only seems to show me 10 results. Not sure where that value comes from. Might be an ElasticSearch thing.

I would suggest using `limit: 10000` instead.

metaspace/webapp/src/components/DatasetCheckboxList.vue
```diff
@@ -16,10 +16,18 @@
...
+          class="flex h-6 items-center"
+        >
+          <span
+            class="truncate"
+            :title="dataset.name"
+          >
+            {{ dataset.name }}
+          </span>
+          <span class="text-gray-700 text-xs tracking-wide pl-1">
+            <elapsed-time :date="dataset.uploadDT" />
```
LachlanStuart, line 30: Looks real nice!

metaspace/webapp/src/api/dataset.ts
richardgoater: oh right, tbf I just guessed it and it seemed to work for me. How many datasets does our most prolific user have?

metaspace/webapp/src/components/DatasetCheckboxList.vue
richardgoater: Thanks! 👍

intsco:
This PR is approved, should it be merged?

richardgoater:
This should have been part of the recent deployment, my fault.

intsco:
Maybe it's time to at least merge it into master so it will be included into the next deployment.

richardgoater:
I guess the next deployment will be some time, but it needs to be merged in any case. Can always rebase it onto release or something.

PR #616, opened 2020-06-17
Fix projectId data type in Python client

LachlanStuart:
Fixes a bug discovered by the MetaboLights team. 

IIRC `projectId`'s datatype was changed during PR in response to feedback. I must have forgotten to rerun the python-client tests after that change. 

Running the tests again, they failed when the datatype was `String!` and now they pass with `ID!`.

I've pushed this to PyPI already as someone was waiting on the fix.

### DIFF
--- a/metaspace/python-client/metaspace/__init__.py
+++ b/metaspace/python-client/metaspace/__init__.py
@@ -1,2 +1,2 @@
 name = 'metaspace2020'
-__version__ = '1.7.3'
+__version__ = '1.7.4'
--- a/metaspace/python-client/metaspace/projects_client.py
+++ b/metaspace/python-client/metaspace/projects_client.py
@@ -91,7 +91,7 @@ class ProjectsClient:
         """
 
         result = self._gql.query(
-            """mutation($projectId: String!, $provider: String!, $link: String!, 
+            """mutation($projectId: ID!, $provider: String!, $link: String!, 
                         $replaceExisting: Boolean!) {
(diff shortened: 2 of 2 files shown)


metaspace/python-client/metaspace/__init__.py
```diff
@@ -1,2 +1,2 @@
 name = 'metaspace2020'
-__version__ = '1.7.3'
+__version__ = '1.7.4'
```
intsco, line 2: Sorry for not catching it during the review. But it's a bit confusing that this version lags behind the git version tags, e.g. here is the 1.7.6 tag https://github.com/metaspace2020/metaspace/commit/bbadde5fc8367318b6bcfe022d33b65c84f16e0e

LachlanStuart: I agree, but I don't really want to manually keep these in sync. We rarely synchronize python-client releases and production releases. Unless we made new releases with no changes to fill the gaps, both sides spend most of their time out of sync.

I think last time we discussed this, we decided it would be best to keep the major/minor versions in sync, but not care about the patch version. Generally we choose to increment the minor version when there is a substantial API addition, and substantial API additions typically also come with python-client changes.

LachlanStuart: > Unless we made new releases with no changes to fill the gaps

To clarify, I mean that even if we had a policy of "skip unused version numbers to catch up to the other side when you do a release", they'd still be out of sync most of the time because they both get releases at different rates. Only through adding fake releases could we keep them somewhat in sync.

PR #617, opened 2020-06-17
[graphql] Custom databases access control in annotation filters

intsco:
The objective of this PR is to filter annotations against custom databases the user has access to.

To facilitate these changes, all databases access code, including `DataLoader`, was moved into a custom TypeORM repository, `MolecularDbRepository.ts`. In `esConnector.ts`, dataset and database authorization filters functions were separated. `ContextUser.getVisibleDatabaseIds` is used by `esConnector.ts` to create ES filters for annotations the user is allowed to see.


metaspace/graphql/src/getContext.ts
```diff
@@ -50,6 +51,12 @@ const getBaseContext = (userFromRequest: JwtUser | UserModel | null, entityManag
       .map(([id, role]) => id);
   };
 
+  const getVisibleDatabaseIds = async (): Promise<number[]> => {
+    const databases = await entityManager.getCustomRepository(MolecularDbRepository)
+      .findDatabases(user as any);
```
intsco, line 56: I'm not super happy about `user as any` hack that is used here to emulate not yet available context user (`MolecularDbRepository` methods accept `ContextUser` argument). But I haven't come up with a better way that won't require big code refactoring.

LachlanStuart: You could use `contextUser` here. Even though it's not declared yet, it's accessible because it's in an outer scope. Neither typescript nor eslint (with the config in the draft branch) seem to complain about it.

metaspace/graphql/src/modules/moldb/controller.ts
```diff
@@ -39,16 +31,18 @@ const assertUserBelongsToGroup = (ctx: Context, groupId: string) => {
...
+  const database = await ctx.entityManager.getCustomRepository(MolecularDbRepository)
+    .findDatabaseById(ctx, databaseId);
+  if (database.public) {
     throw new UserError('Cannot edit public database');
   }
-  if (moldb.groupId != null) {
-    assertUserBelongsToGroup(ctx, moldb.groupId);
+  if (database.groupId != null) {
+    assertUserBelongsToGroup(ctx, database.groupId);
   }
```
LachlanStuart, line 41: Is this logic correct? 

I wouldn't expect `moldb.public` to matter at all - "public" databases are just databases that other people can see, they're not irrevocably published.

Also, I think it should only allow admins to edit if `moldb.groupId == null`.

metaspace/graphql/src/modules/moldb/MolecularDbRepository.ts
```diff
@@ -0,0 +1,76 @@
...
+    if (whereClause) {
+      qb = qb.andWhere(whereClause, parameters);
+    }
+
+    // Avoid adding .where clauses to the returned queryBuilder, as it will overwrite the security filters
+    return qb;
+  }
+
+  private createDataLoader(ctx: Context, functionName: string) {
+    return ctx.contextCacheGet(functionName, [], () => {
```
LachlanStuart, line 48: This is just one DataLoader. No need for two different `functionName` values. The same instance can be safely shared between the two callers.

metaspace/graphql/src/modules/moldb/MolecularDbRepository.ts
```diff
@@ -0,0 +1,76 @@
...
+@EntityRepository()
+export class MolecularDbRepository {
+  constructor(private manager: EntityManager) {
+  }
+
+  private queryWhere(user: ContextUser | null, whereClause?: string | Brackets, parameters?: object) {
+    const columnMap = this.manager.connection
+      .getMetadata(MolecularDB)
+      .columns
+      .map(c => `"moldb"."${c.databasePath}" AS "${c.propertyName}"`);
```
LachlanStuart, line 19: Why the custom column mapping here, and the use of `getRawMany` below? These should only be necessary if you're adding extra computed columns...

metaspace/graphql/src/getContext.ts
intsco: Oh nice, `getVisibleDatabaseIds` uses `contextUser` that uses `getVisibleDatabaseIds` itself. It's a bit confusing but if it works, I'm fine with it.

metaspace/graphql/src/modules/moldb/controller.ts
intsco: >I wouldn't expect moldb.public to matter at all - "public" databases are just databases that other people can see, they're not irrevocably published.

I think I assumed that the public databases will be managed by python scripts but it's a good point, they don't have to.

I also used a wrong condition `database.public`, it meant to be `database.groupId == null`.

>Also, I think it should only allow admins to edit if moldb.groupId == null

This won't work for the current hidden databases as they are being moved into the EMBL group.

PR #622, opened 2020-06-19
Graphql dependency updates

LachlanStuart:
Manually updated packages, because code doesn't compile after the security updates in #607.

Updated packages:
* `apollo-server-express` 
* `subscriptions-transport-ws`
* `graphql-subscriptions` 
* `graphql@^14.5.3` (`@15` requires node 10+)

Several uses of `@deprecated` in .graphql files started to raise errors, as `@deprecated` isn't allowed on input types or arguments. These were commented out to stop the errors.

`graphql-server-core`'s `runQuery` utility function was refactored out. Fortunately there was an easy replacement option using the `graphql` library directly.

I also removed the old e2e tests in graphql, as they were unmaintained, not part of CI, and dependent on mocha.


LachlanStuart:
@richardgoater 
> Is that essentially a breaking change in apollo-server-core? 

Yes, but it's a bit weird. Previously `@deprecated` wasn't handled at all because it wasn't in the spec. It was treated like any other unknown directive and ignored. Once it was added to the spec, it started being validated against the specified allowed locations. There are currently several issues raised against the spec and the implementation to expand the allowed locations, but GraphQL development moves rather slowly.

> How does one deprecate input fields?

I think that directive is only used for generated documentation. Checking the server response of using a deprecated output field, it doesn't look like there's any automated warning or anything. Adding a comment to the field would achieve a similar effect.

PR #625, opened 2020-06-24
Last minute fixes

LachlanStuart:
Several last-minute fixes for custom DB edge cases:
* Filter output of `findDatabasesByIds` to not include `null`s. Previous behavior would return an array with nulls, and the `Dataset.databases` resolver would return them, causing a validation error. The alternate solution would be to change the `findDatabasesByIds` return type to include nulls, and handle the nulls at every caller. However, I felt most callers were built under the assumption that hidden databases just weren't returned, so I think filtering in `findDatabasesByIds` is the expected behavior.
* Filter `ds._source.annotation_counts` correctly (my previous `?? []` fix would have broken it - I made that thinking I was changing a `.map`, not a `.filter`). And exclude DBs not visible to the current user.
* Remove `logger.error` when a database can't be found in `ds._source.annotation_counts`. I don't think these deliver value as we don't normally monitor the logs. It was easier to remove it than factor in the possibility that no data was found due to hidden DBs (which is an edge cases that only happens when HMDB-v4 is somehow excluded/private)
* Fix `if (databaseId != null) {` codepath. If `databaseId` was null due to having no visible annotation count records, this would have returned an invalid object with null `databaseId`/`dbName` fields, failing schema validation. Now it returns a `null` value instead of an object with `null` fields.
* Unrelated fix: Make the `currentUser` prop in `DatasetItemActions.vue` optional, because Vue was logging console warnings as it's `null` when not logged in. My bad for not testing this case in my PR.

### DIFF
--- a/metaspace/engine/sm/engine/msm_basic/msm_basic_search.py
+++ b/metaspace/engine/sm/engine/msm_basic/msm_basic_search.py
@@ -126,8 +126,9 @@ def compute_fdr_and_filter_results(
         'moldb_id', axis=1
     )
     moldb_metrics_fdr_df = compute_fdr(fdr, formula_metrics_df, moldb_formula_map_df)
-    max_fdr = 1.0 if moldb.targeted else 0.5
-    moldb_metrics_fdr_df = moldb_metrics_fdr_df[moldb_metrics_fdr_df.fdr <= max_fdr]
+    if not moldb.targeted:
+        max_fdr = 0.5
+        moldb_metrics_fdr_df = moldb_metrics_fdr_df[moldb_metrics_fdr_df.fdr <= max_fdr]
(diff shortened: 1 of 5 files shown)


LachlanStuart:
Added two more fixes:
* Allow FDR values above 1.0. [Previous discussion](https://github.com/metaspace2020/metaspace/pull/585/files/d864a0e9c2ad3aeb899bc9012af9eaab9d6b27a9#r434490720). It looks like `analysis_version==1` defaults FDRs to 1.0 if they don't match any of the pre-defined levels. `analysis_version==2` doesn't discretize FDR values, so values above 1.0 are possible, though uncommon.
* Fixed an error when reprocessing datasets without changing the databases. In these cases, `doc['moldb_ids']` was not defined.

PR #627, opened 2020-06-25
[graphql] Check that only group members can use custom databases

intsco:
Public group databases can be used only by the group members but results against them can be viewed by everyone.

### DIFF
--- a/metaspace/graphql/src/getContext.spec.ts
+++ b/metaspace/graphql/src/getContext.spec.ts
@@ -31,6 +31,7 @@ describe('getContext', () => {
     userId = testUser.id;
   });
   afterEach(onAfterEach);
+
   describe('cachedGetEntityById', () => {
     const testIds = [1,2,3].map(n => `00000000-1234-0000-0000-00000000000${n}`);
     const [id1, id2, id3] = testIds;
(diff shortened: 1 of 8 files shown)


intsco:
@LachlanStuart I actually forgot to add "can use" check to the `molecularDatabases` query. I think returning databases that user cannot use doesn't make much sense. If we implement this behavior, both `canUse` and `hideUnusable` parameters won't be needed. Or am I missing something?

LachlanStuart:
@intsco A list of all visible databases, including unusable-but-public databases, must be available for the database filter on the annotations page. By default it should only show usable databases, but unusable-but-public databases can be selected by this filter in other ways, e.g. the "browse annotations" links on the Datasets page.

Per previous discussions, I believe @richardgoater was planning to do all the filtering on the client-side because the list of all visible databases is unlikely to be very large. This means the webapp needs to know whether a database is usable, so that it can decide whether to show it by default.

richardgoater:
Sorry to chime in late on this, I didn't want to do everything client-side, that was prescribed for me. I would prefer the API to return only the necessary data, and I was going to say that the growing list of flags is looking untidy and tricky to use. I would suggest creating an enum for the levels of access we need, happy to discuss.

intsco:
I'm merging this PR and adding the agreed upon changes to the next PR.

PR #630, opened 2020-06-30
[graphql] New MolecularDB fields

intsco:
* `created_at` and `molecule_link_template` MolecularDB model fields.
* Data migration of the template links for the metaspace databases.
* Remove deprecated database name field from a few graphql types.
* Replace `mapToMolecularDB` with proper `MolecularDbResolvers` object.
* Rename `MolecularDB.public` to `isPublic`.
* Refactor the engine tests, extract molecular database creation into _tests.utils_ module.
* Add `MolecularDB.group` and `group.molecularDatabases` fields.
* Replace `hideArchived` argument of `molecularDatabases` query with `onlyUsable` as it has broader meaning.

Most of the engine changes can be safely skipped as they are mostly tests refactorings.


intsco:
Most of the engine changes can be safely skipped as they are mostly tests refactorings.

richardgoater:
Something interesting happening to the webapp tests

intsco:
@richardgoater I have implemented the strategy that we discussed today in this commit. Please review it
https://github.com/metaspace2020/metaspace/pull/630/commits/5fb8f348a230f480d881b444a827a8752c8ac100

richardgoater:
> @richardgoater I have implemented the strategy that we discussed today in this commit. Please review it
> [5fb8f34](https://github.com/metaspace2020/metaspace/commit/5fb8f348a230f480d881b444a827a8752c8ac100)

Thanks @intsco. Naming things is hard but I think `hideUnusable` is sort of a double-negative and that makes it harder to understand. I would suggest `usableOnly` or `onlyUsable`.

Just realised a cleaner future approach could be something like `$filter = { usable: true }`, allowing extendable options.

intsco:
>what is targeted?

@richardgoater We call small databases (<1k formulas) _targeted_ because with such small databases we perform a targeted search and show users the results for every single from the database. Our default behavior, _untargeted_ search, is to search through a large database, like HMDB, and pick only the highest scoring molecules.

At the moment, this field is used inside engine but it also might be useful in webapp, e.g. to customize the annotation view.

richardgoater:
Thanks! I was worried it was needed on the management UI and I had missed
it :)

LachlanStuart:
The webapp tests are failing because of the two removed fields from the .graphql files. The reason that the error messages are so messy is that it seems apollo only logs a warning that the query has failed, allowing the test to continue running and fail for other reasons when it doesn't get any data back from the mock graphql datasource. Fixing the .graphql files will fix the tests.

PR #635, opened 2020-07-09
feat(engine): Save zero MSM annotations

intsco:
For targeted databases, save annotations even with zero scores if at least one image is not empty.

_Edit:_
* Add `isPublic` field to molecular database mutations.
* `MolecularDB.createdDT` init with the local time zone and not UTC. I double checked behavior of this graphql field, it's consistent with `Project.createdDT`.

### DIFF
--- a/metaspace/engine/sm/engine/molecular_db.py
+++ b/metaspace/engine/sm/engine/molecular_db.py
@@ -117,7 +117,7 @@ def create(
                 (
                     name,
                     version,
-                    datetime.utcnow(),
+                    datetime.now(),
                     group_id,
                     is_public,
                     description,
--- a/metaspace/engine/sm/engine/molecular_db.py
+++ b/metaspace/engine/sm/engine/molecular_db.py
@@ -136,6 +136,7 @@ def delete(moldb_id: int):
     DB().alter('DELETE FROM molecular_db WHERE id = %s', params=(moldb_id,))
 
 
+# pylint: disable=unused-argument
 def update(
     moldb_id: int,
     archived: bool = None,
--- a/metaspace/engine/sm/engine/molecular_db.py
+++ b/metaspace/engine/sm/engine/molecular_db.py
@@ -145,16 +145,15 @@ def update(
     link: str = None,
     citation: str = None,
 ) -> MolecularDB:
-    assert archived is not None or description or full_name or link or citation
-
     kwargs = {k: v for k, v in locals().items() if v is not None}
     kwargs.pop('moldb_id')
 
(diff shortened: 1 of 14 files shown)


metaspace/engine/sm/engine/msm_basic/formula_validator.py
```diff
@@ -83,39 +84,51 @@ def compute_metrics(iso_images_sparse, formula_ints):
     return compute_metrics
```
LachlanStuart, line 84: GitHub won't let me leave a comment on the exact line - approx 5 lines above, but `doc['total_iso_ints']`, `doc['min_iso_ints']` and `doc['max_iso_ints']` are only set for annotations with non-zero metrics. This causes a big downstream problem - webapp shows all the images as black because the intensities are all zero after scaling to the unset `min`/`max_iso_int` values.

metaspace/engine/sm/engine/molecular_db.py
```diff
@@ -117,7 +117,7 @@ def create(
                 (
                     name,
                     version,
-                    datetime.utcnow(),
+                    datetime.now(),
```
intsco, line 120: This actually doesn't change anything. The time is saved correctly in UTC in any case.

LachlanStuart: If your local setup is all in docker, then UTC and local time are probably the same thing. Docker containers run in UTC unless you explicitly override their timezones.

I tested from IPython outside docker and can confirm that there is a difference in behavior between saving `datetime.now()` and `datetime.utcnow()`:

```python
from sm.engine.db import ConnectionPool, DB
from sm.engine.util import SMConfig
from datetime import datetime
c = ConnectionPool(SMConfig.get_conf()['db'])
db = DB()
db.alter('CREATE TABLE foo (ts TIMESTAMP);')
db.alter('INSERT INTO foo VALUES (%s)', (datetime.now(), ))
db.alter('INSERT INTO foo VALUES (%s)', (datetime.utcnow(), ))
db.select('SELECT * FROM foo')
```
> ```python
> [(datetime.datetime(2020, 7, 13, 19, 12, 27, 540638),),
> (datetime.datetime(2020, 7, 13, 17, 12, 31, 659750),)]
> ```

metaspace/graphql/schemas/dataset.graphql
```diff
@@ -59,6 +59,7 @@ type Dataset {
 type FdrCounts {
   databaseId: Int!
   dbName: String!
+  dbVersion: String!
```
intsco, line 62: @richardgoater Here is an extra field you can use in the databases pop-up on the datasets page.

PR #636, opened 2020-07-15
feat(graphql): Сheck that public database name doesn't exist

intsco:
`MolecularDB` create/update mutations: check that a public database with the same name doesn't exist.

### DIFF
--- a/metaspace/graphql/schemas/moldb.graphql
+++ b/metaspace/graphql/schemas/moldb.graphql
@@ -48,6 +48,7 @@ type Query {
 }
 
 type Mutation {
+  # throws exceptions of types: ["wrong_parameters", "public_already_exists", "already_exists", "malformed_csv"]
   createMolecularDB(databaseDetails: CreateMolecularDBInput!): MolecularDB!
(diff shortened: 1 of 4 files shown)


LachlanStuart:
Doesn't this prevent having multiple versions of a public database?

Is this change actually necessary? Ideally we'd show which group a database comes from in the UI, so there shouldn't be any problems caused by using the same DB name in different groups...

richardgoater:
> Ideally we'd show which group a database comes from in the UI

Reckon this is an impeding issue, don't think the Element UI select component is adequate for this. I'd prefer to build one that orders/separates the list of databases by group.

intsco:
@LachlanStuart 
>Doesn't this prevent having multiple versions of a public database?

I think I wrote this even before we agreed on name-version splitting of the existing public database names.

>Is this change actually necessary? Ideally we'd show which group a database comes from in the UI, so there shouldn't be any problems caused by using the same DB name in different groups...

Okay, I wasn't completely sure whether we are going to implement this. If that's the plan, this PR isn't necessary I think.

Issue #552, opened 2020-02-26
Update node version on production


LachlanStuart:
We're currently running node 8, which is now unsupported. We should update to the most recent version supported by our dependencies.

LachlanStuart:
Easiest path forward seems to be installing node through nvm. This would allow us to specify an exact version, upgrade easily, and even run different versions for different projects.

PR #637, opened 2020-07-15
Upgrade to Node 14

LachlanStuart:
This change adds nvm for managing node versions, which would allow webapp/graphql to have their own independent versions if necessary. Unfortunately nvm did not integrate well with ansible / supervisor, so the commands to launch node/yarn are a bit hacky in places.

The main issue is that nvm isn't a script - it's a shell function that gets installed when you `source` the ` ~/.nvm/nvm.sh` script, which usually is done automatically via `.bashrc`. To get past this, I had to ensure `bash` was running whenever nvm was needed, and had to prefix commands with `. ~/.nvm/nvm.sh` to install the shell function and `nvm use` to switch to the working directory's specified version of node. Neither Ansible nor Supervisord have any non-hacky ways to source `.profile`/`.bashrc`/etc. before running their commands.

Node 14 is [not LTS yet](https://nodejs.org/en/about/releases/) - it's due to become LTS in 3 months. I like new shiny things though, and Node 14 has native ES Module support. I pinned it to 14.5.* because I've had dev environment issues in the past when my local copy of node has had a different minor version than my docker copy.

@richardgoater Please review the JS changes

@intsco Please review the ansible & docker changes

Related: #552 

## Code changes

* Remove hacks needed for node 8 (asyncIterator polyfill, incorrect date format in ElapsedTime.spec.ts)
* Updated FilterPanel snapshot. It seems the reason it was reordered was because Array.prototype.sort started using a stable sorting algorithm in Node 11, and `sortFilterKeys` in store/mutations.js sorts against the same value for most items.

Some dependencies needed to be updated for Node 14 support, or to prevent console warnings:
* Update webapp `babel` (via manual lockfile deletion & reinstall) due to babel/babel#11216
* Update webapp `node-sass` for Node 14 binary support
* Update graphql `bcrypt` to fix the rebuild-on-node-version-change behavior, as it would previously keep the node 8 binaries even after a `yarn install` with node 14
* Update graphql `knex` and `pg` due to knex/knex#3836
* Update graphql `winston` due to winstonjs/winston#1797
* Update graphql `ts-node-dev` due to whitecolor/ts-node-dev#143

I also snuck in a couple unrelated changes:
* Update `element-ui` due to ElemeFE/element#19389 (this benign bug has happened >200 times on Sentry)
* Pin version of `graphql-binding` because I remember newer versions not being compatible

## Ansible changes

* I found my previous changes that map `python` to `python3` were overwritten when provisioning from scratch, so I had to make several fixes to prevent python2 from overwriting the symlink.
* I fixed a warning about the use of jinja `{{ var }}` templates in the `when` clause in the ElasticSearch task
* I re-enabled `autorestart` for the supervisor services.
* I found an issue in sm-web - Mol DBs can't be installed before sm-graphql has created the database schema. Too hard to fix in this PR, so I left a TODO.

## Docker changes

* Switched to alpine images
* Removed code for pre-caching yarn packages, as doesn't help for development mode

## Testing

For ansible, I provisioned & deployed to a new dev server to test that the ansible changes all worked. 

For webapp, I'd expect any issues would surface during build/test. There are no new errors/warnings as far as I can see.

For graphql:
* I did the basic CRUD operations for datasets and projects to test that interop with Postgres, ElasticSearch, sm-api and the graphql image/storage server all still work.
* I tested a dataset status update and a system maintenance message to check that RabbitMQ, and websocket subscriptions still work.

## Deployment

I don't plan to merge this until after the current batch of changes to prod are deployed, as the pending deployment already has a lot of manual steps.

Before deployment the following should be run to install nvm:
```bash
ansible-playbook -i env/ENVIRONMENT provision/web.yml -t nodejs
```


docker/sm-graphql/Dockerfile
```diff
@@ -1,14 +1,15 @@
-FROM node:8.16.2
+FROM node:14.5.0-alpine3.12
```
richardgoater, line 1: :+1: for Alpine

ansible/roles/sm_cluster_autostart_app/templates/sm-cluster-autostart-daemon.supervisor.j2
```diff
@@ -8,5 +8,4 @@ redirect_stderr = true
 stdout_logfile = {{ sm_ansible_home }}/aws/logs/{{ sm_cluster_autostart_app_name }}.log
 numprocs = 1
 numprocs_start = 7000
-autorestart = false
```
intsco, line 11: I wonder if this can become a problem if daemon get stuck in an indefinite loop of starting and failing. It has built-in retry logic for some of the error-prone steps. I'm also not sure if the daemon can successfully pick up where it left after restarting.

ansible/roles/sm_graphql/tasks/main.yml
```diff
@@ -1,38 +1,26 @@
...
-    name: graphql
-
-# TODO: Move into sm_web tasks after production has been updated, as this is a provisioning step
-- name: Import uuid-ossp extension into sm postgres database
-  become: yes
-  become_user: postgres
-  postgresql_ext: db=sm name=uuid-ossp
+- name: Install sm-graphql's version of Node.js (if needed)
+  # This only installs a new version if there isn't a matching existing version
+  shell: . ~/.nvm/nvm.sh && nvm use || nvm install
```
intsco, line 5: To reduce some duplication, you could use a role or global variable:
```
nvm_use: . ~/.nvm/nvm.sh && nvm use
```
```
shell: {{ nvm_use }} || nvm install
```

ansible/roles/sm_web/tasks/main.yml
```diff
@@ -52,6 +65,8 @@
     executable: /bin/bash
   when: sm_create_es_index
 
+
+# TODO: The DB schema for this doesn't exist until graphql has run migrations. Run graphql migrations before this step.
```
intsco, line 69: Why to run graphql migrations, shouldn't running `db_schema.sql` be enough?

LachlanStuart: Running `db_schema.sql` unfortunately doesn't populate the migrations table, which means graphql would later try to run the migrations and fail due to the tables already existing.

There's no reason why we couldn't make `db_schema.sql` populate the migrations table though... It'd just be more code in the `dump_sql_schema.js` script that generates `db_schema.sql`. I'll update the comment to put that as a suggested solution.

ansible/roles/sm_graphql/tasks/main.yml
LachlanStuart: Unless I misunderstand roles, I don't think they would work. Each `shell` task is executed with a clean environment, so it's not possible to run an environment-affecting command in one task and use those environment vars in another task.

I see there's already a `sm_activate_venv` variable, so I'll follow that pattern.

What's frustrating is that Ansible are aware of the problem, but their [suggested solution](https://docs.ansible.com/ansible/latest/user_guide/playbooks_environment.html#working-with-language-specific-version-managers) is to just figure out conda/nvm/rbenv's internal directory structure and hard-code changes to the `$PATH` variable. I didn't want to use that in this case, because I don't want to add even more places where the node version is explicitly specified, and I don't want to specify the node version down to the patch level if it's not necessary.

ansible/roles/sm_cluster_autostart_app/templates/sm-cluster-autostart-daemon.supervisor.j2
LachlanStuart: In my experience it generally has been pretty good at recovering when restarted. You make a fair point though - the daemon is a risky component for autorestart. If it was misconfigured, it could potentially start many copies of the cluster, etc.

I'll revert this file

ansible/roles/sm_graphql/tasks/main.yml
intsco: >What's frustrating is that Ansible are aware of the problem, but their suggested solution is to just figure out conda/nvm/rbenv's internal directory structure and hard-code changes to the $PATH variable. I didn't want to use that in this case, because I don't want to add even more places where the node version is explicitly specified, and I don't want to specify the node version down to the patch level if it's not necessary.

It might actually be a good idea to explicitly specify the version in the configuration management system, e.g. with a global variable, since it's important to know what exactly and where is installed. Not sure about other places though.

PR #641, opened 2020-07-17
Misc back-end fixes

LachlanStuart:
* Increase Spark memory. I decided that it was better to possibly set it too high than too low. If Java OOMs we get confusing exception messages. If Python OOMs because Java has already eaten too much memory, we get nicer exception messages.
* Change all the `README.md` files to link to the main `README.md`'s acknowledgements/funding/license sections, as the main one is nicer and more up-to-date.
* Removed the link to Gitter from `README.md`. The link has already been removed from webapp.
* Retry downloading images in python-client if there's an error. VS found that she would often get errors on her home internet connection.
* Retry downloading/uploading/deleting images if there's an error in `ImageStoreServiceWrapper`. This probably won't be an issue once the instance is upgraded, but the prod logs include several failures in image upload and image download(for colocalization). The requests library only retries on certain types of exception, so it seemed safer to handle the retry logic externally.
* Reduced pool size and added `pool_block=True` to `ImageStoreServiceWrapper` to try to reduce the frequency of exceptions. Normally the number of connections is actually allowed to exceed `pool_maxsize`, it just won't actually keep them in the pool above that number. `pool_block=True` changes this so that it will block execution instead of allocate non-pooled connections.
* Change ESExporter's retry decorator to retry on any ES-specific exception. I found a few cases of `ConnectionTimeout`s in the logs, and I found that if `update_ds` hypothetically manages to have a conflict (this generally doesn't happen anymore due the DB-based dataset locker), it would often actually throw as a `NotFoundError` because of conflicts creating/deleting the pipeline. I felt it safer to just retry on any error.

### DIFF
--- a/metaspace/engine/sm/engine/es_export.py
+++ b/metaspace/engine/sm/engine/es_export.py
@@ -1,12 +1,15 @@
 import logging
-import random
 from collections import MutableMapping, defaultdict
-from functools import wraps
-from time import sleep
 
 import numpy as np
 import pandas as pd
-from elasticsearch import ConflictError, Elasticsearch, ElasticsearchException, NotFoundError
+from elasticsearch import (
+    ConflictError,
+    Elasticsearch,
+    ElasticsearchException,
+    NotFoundError,
+    ConnectionTimeout,
+)
 from elasticsearch.client import IndicesClient, IngestClient
 from elasticsearch.helpers import parallel_bulk
 
@@ -17,7 +20,7 @@ from sm.engine.formula_parser import format_ion_formula
 from sm.engine.isocalc_wrapper import IsocalcWrapper
 from sm.engine import molecular_db
 from sm.engine.molecular_db import MolecularDB
-from sm.engine.util import SMConfig
+from sm.engine.util import SMConfig, retry_on_exception
 
 logger = logging.getLogger('engine')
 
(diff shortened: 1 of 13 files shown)


metaspace/engine/sm/engine/es_export.py
```diff
@@ -305,7 +285,7 @@ def _remove_mol_db_from_dataset(self, ds_id, moldb):
     def _select_ds_by_id(self, ds_id):
         return self._db.select_with_fields(DATASET_SEL, params=(ds_id,))[0]
 
-    @retry_on_conflict()
+    @retry_on_exception(TransportError)
```
intsco, line 288: In the PR description, it says that you decided to catch any ES exceptions. Does `TransportError` cover all concerned cases?

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -868,7 +868,10 @@ def fetchImage(url):
             if not url:
                 return None
             url = self._baseurl + url
-            im = mpimg.imread(BytesIO(self._session.get(url).content))
+            try:
+                im = mpimg.imread(BytesIO(self._session.get(url).content))
+            except:
+                im = mpimg.imread(BytesIO(self._session.get(url).content))
```
intsco, line 874: Is it just a custom retry once logic?

intsco:
Do we plan to make it a part of the current production deployment?

LachlanStuart:
> Do we plan to make it a part of the current production deployment?

No. All the testing was done prior to this PR, so it's too late to include it.

I will probably do a python-client release later this week and deploy Node 14 and this PR next week.

metaspace/python-client/metaspace/sm_annotation_utils.py
LachlanStuart: "custom" -> "lazy" :wink: 

Doing this the right way would have been a lot more verbose. I didn't feel it was worth the extra code bloat.

metaspace/engine/sm/engine/es_export.py
LachlanStuart: `TransportError` covers all server-generated HTTP errors (including `NotFoundError` and `ConflictError`), as well as connection errors (e.g. `ConnectionTimeout`). However, it doesn't include general code errors inside the elasticsearch library such as `ImproperlyConfigured` and `SerializationError`.

PR #647, opened 2020-07-24
More python client fixes

LachlanStuart:
This fixes two issues that would make a python-client release potentially problematic.

I moved the default database filter out of `GraphQLClient.getAnnotations` into `SMDataset.results` and `SMDataset.annotations`. Several reasons:
* Due to `database` being converted to `databaseId` in the `SMDataset` layer, `database` was being re-added in the `GraphQLClient` layer. GraphQL gives priority to `databaseId`, but it caused warnings to be printed, and made it impossible to remove the database filter entirely.
* Due to weird indentation that probably resulted from a merge conflict, the default filters were only applied in `GraphQLClient` when some value was supplied for `annotationFilter`. This meant that anyone using `GraphQLClient` directly could get very unexpected behavior when adding/removing filters.
* Having the defaults in the low-level methods has caused a lot of issues in the past, as it's really easy to change the code in the high-level layer in a way that makes it impossible to remove the filter by overriding the default value with an explicit `None`.

For forwards compatibility with the discussed change, I made `map_database_to_id` handle both integer and stringified integer IDs, and added some tests, and cleaned up a few places where passing no database at all may have caused issues.

Also, I changed how `json_normalize` is used so that it doesn't log `DeprecationWarning`s with Pandas >=1.0.0

### DIFF
--- a/metaspace/python-client/metaspace/sm_annotation_utils.py
+++ b/metaspace/python-client/metaspace/sm_annotation_utils.py
@@ -22,6 +22,11 @@ try:
 except ImportError:
     TypedDict = dict
 
+try:
+    from pandas import json_normalize  # Only in Pandas 1.0.0+
+except ImportError:
+    from pandas.io.json import json_normalize  # Logs DeprecationWarning if used in Pandas 1.0.0+
+
 
 DEFAULT_DATABASE = 'HMDB-v4'
 
@@ -787,7 +792,7 @@ class SMDataset(object):
         if not records:
             return pd.DataFrame()
 
-        df = pd.io.json.json_normalize(records)
+        df = json_normalize(records)
         return (
             df.assign(
                 moleculeNames=df.possibleCompounds.apply(
@@ -1129,7 +1134,7 @@ class SMInstance(object):
         Pandas dataframe for a subset of datasets
         where rows are flattened metadata JSON objects
         """
-        df = pd.io.json.json_normalize([d.metadata.json for d in datasets])
+        df = json_normalize([d.metadata.json for d in datasets])
         df.index = [d.name for d in datasets]
         return df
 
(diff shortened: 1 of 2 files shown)


metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -487,7 +510,7 @@ def create_dataset(
                 'name': ds_name,
                 'inputPath': data_path,
                 'isPublic': is_public,
-                'databaseIds': [self.map_database_to_id(db) for db in databases],
+                'databaseIds': databases and [self.map_database_to_id(db) for db in databases],
```
intsco, line 513: Why to have this extra "databases is not empty" check here?
```
a = []
[x for x in a]
Out[62]: []
```

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -21,6 +22,14 @@
 except ImportError:
     TypedDict = dict
 
+try:
+    from pandas import json_normalize  # Only in Pandas 1.0.0+
+except ImportError:
+    from pandas.io.json import json_normalize  # Logs DeprecationWarning if used in Pandas 1.0.0+
+
+
+DEFAULT_DATABASE = 'HMDB-v4'
```
intsco, line 31: Reminder to myself, not to forget to update this in `feat/database-management-ui` (https://github.com/metaspace2020/metaspace/blob/feat/database-management-ui/metaspace/graphql/src/migrations/1594744305589-MolecularDbNameVersionSplit.ts#L14)

metaspace/python-client/metaspace/sm_annotation_utils.py
LachlanStuart: `databases` is an optional argument with a default value of `None`, not empty array.

PR #649, opened 2020-07-24
feat(graphql): Set custom database annotations FDR to null

intsco:
* Store null FDR values in ES as -1
* Modify the FDR filter, use one-sides range: `fdr < fdr_level  + 0.001`

Should wait for https://github.com/metaspace2020/metaspace/pull/638


richardgoater:
@intsco what should happen on the UI when FDR is null?

intsco:
@richardgoater I think we should indicate the user that this FDR value isn't relevant, e.g. by placing '-' or other symbol instead of the FDR. I'd also change the color of the cell from green to something more neutral.

richardgoater (approved):
Thanks dude, I will follow on with the front-end change separately, don't think I handled it in the big PR.

LachlanStuart (approved):

PR #654, opened 2020-08-04
Security fixes & small tweaks

LachlanStuart:
* Removed the DB-based image storage. This code had a dependency on an insecure version of the jpeg-js package. Although it was probably trivial to fix, we don't use this storage method, and probably won't ever use it in the future, so I thought it better to just remove the code and the dependency.
    * I've left the fields in place to allow different storage types, as I don't know what the S3/COS storage will look like yet.
* Upgrade Pillow 6.2.0 -> 7.2.0 to address the security advisory. I tested annotation and optical image upload - both worked with no code changes
* Fixed a couple issues in the JS dockerfiles:
    * Yarn doesn't detect when packages need to be rebuilt for the Alpine images' musl libc. I was getting frequent issues caused by missing musl libraries when I `yarn install`ed outside the container.
    * The nodemon-triggered `gen-binding` script often wouldn't finish compilation before TypeScript would try to load it. The solution was to explicitly run gen-binding synchronously before starting the TypeScript process.
* Suppress a spammy warning generated by spectral clustering

### DIFF
--- a/docker/sm-graphql/docker-entrypoint.sh
+++ b/docker/sm-graphql/docker-entrypoint.sh
@@ -18,7 +18,9 @@ if [ "$SM_DOCKER_ENV" = "development" ]; then
   export NODE_ENV=development
   cd /opt/dev/metaspace/metaspace/graphql
   yarn install
+  npm rebuild bcrypt # Ensure the musl version is installed
   yarn run deref-schema
+  yarn run gen-binding
   nodemon -e graphql -q --exec "yarn run gen-binding" &
   # exec nodemon -e js,json,ts --require ts-node/register server.js
   exec yarn exec ts-node-dev -- --respawn server.js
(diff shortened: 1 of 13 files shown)


docker/sm-graphql/docker-entrypoint.sh
```diff
@@ -18,7 +18,9 @@ if [ "$SM_DOCKER_ENV" = "development" ]; then
   export NODE_ENV=development
   cd /opt/dev/metaspace/metaspace/graphql
   yarn install
+  npm rebuild bcrypt # Ensure the musl version is installed
   yarn run deref-schema
+  yarn run gen-binding
   nodemon -e graphql -q --exec "yarn run gen-binding" &
```
intsco, line 24: Is this line still needed?

LachlanStuart: Yes. It sets up nodemon to continuously watch the `.graphql` files and rerun `yarn run gen-binding` if there are any changes (e.g. if you change branch).

PR #655, opened 2020-08-06
feat(python-client): CRUD operations with custom molecular databases

intsco:
The PR provides classes and methods in python-client for managing and querying molecular databases.

* Default database config value is replaced with a database table field.
* Split M\  database names into the name and version. A migration script provided. For backwards compatibility, handle use of the old database names in graphql and python-client.
* New python-client data class `MolecularDB`.
* Use better naming for the custom databases graphql queries.
* Allow users to delete databases in their groups.
* Allow database management mutations when authenticated via an API key.
* Add `multipart_upload` function for handling S3 multipart uploads from python-client.
* Add python-client tests for molecular database actions.
* Python-client GraphQL queries better formatting.


metaspace/graphql/schemas/moldb.graphql
```diff
@@ -44,7 +44,10 @@ input UpdateMolecularDBInput {
 }
 
 type Query {
-  molecularDatabases(onlyUsable: Boolean=false): [MolecularDB!]
+  visibleMolecularDBs: [MolecularDB!]
+  usableMolecularDBs: [MolecularDB!]
+  metaspaceMolecularDBs: [MolecularDB!]
```
LachlanStuart, line 49: I don't know whether to comment to @intsco or @richardgoater ... this is in both PRs, but presumably this one will be merged first, so it should be fixed here? I have already written out an explanation tailored for @richardgoater , so here's the version for him:

> We try to keep backwards compatibility in the graphql schema. Old webapp code can stay active for anywhere between 10 minutes (due to HTTP caching) to 24 hours (due to people leaving browser tabs open). Also, people rarely update the Python client, and it's not obvious that they need to update it when there's a schema-related error. We also have people who have services that are in continual operation that depend on python-client (MetaboLights and SpaceM), and it would be a dick move to break a previous version without informing them and giving them a chance to upgrade.
>
> Generally we avoid making significant structural changes to existing GraphQL methods for this reason. If you want to get rid of the `molecularDatabases` query completely, the only safeish way to do so is deprecate it in one release and make sure no code is still using it (including either coordinating with the MetaboLights & SpaceM teams, or waiting a few months and hoping they upgrade), then remove it in a separate release.
>
> IMO, this change isn't worth the time cost of change. It certainly is possible to re-add the `molecularDatabases` query but deprecate it in favor of the new queries, but my recommendation in this case would be to go back to the original pattern of adding parameters with default values to `molecularDatabases` as it would be less work.
>
> In GraphQL it's also generally better practice to have a small number of highly configurable queries than to have tailor-made queries for all the different use cases. It's easier to maintain fewer queries, and there's a greater likelihood that new use cases can be served by an existing queries. `allAnnotations` is a great example of this - it's used for a variety of different views on the Annotations page (the main list, the isomers/isobars/other adducts/colocalized molecules lists, the isobars dropdown in the diagnostics section, CSV export, etc.) and has required very little new code to support each additional use case.

metaspace/graphql/schemas/moldb.graphql
```diff
@@ -44,7 +44,10 @@ input UpdateMolecularDBInput {
...
+  getMolecularDB(databaseId: Int!): MolecularDB!
```
LachlanStuart, line 50: To be consistent with other "get by ID" Query names like `dataset` and `annotation`, this query should be called `molecularDB`.

metaspace/graphql/schemas/moldb.graphql
```diff
@@ -25,8 +25,8 @@ type MolecularDB {
 input CreateMolecularDBInput {
   name: String!
   version: String!
-  filePath: String!
-  isPublic: Boolean! = False
+  filePath: String!  # s3://[BUCKET]/path/to-file.tsv
+  isPublic: Boolean = False
```
LachlanStuart, line 29: Why is this now nullable?

intsco: Don't they actually call them **optional** vs **required** fields? Not sure if **nullable** is the same as optional (a general Python developer confusion). From https://graphql.org/learn/schema/:
"Arguments can be either required or optional. When an argument is optional, we can define a default value..."

If I'm not missing something, having a required argument that have a default value doesn't make any sense.

metaspace/graphql/schemas/moldb.graphql
intsco: This is the code that I blindly cherry-picked from the `feat/database-management-ui` branch. I agree that this a breaking change and we'd better stick to extending `molecularDatabases` query. Although, with each extra argument it gets more difficult to reason about its expected result.

richardgoater: Think I mentioned a potential solution before: an enum of named views/filters. This way they are not intended to be combinatorial, as this API change tried to achieve.

I'm also thinking that we could move visible/usable to the `currentUser` query as we could consider these conditions to be a relation. Could even group them together like this:
```
currentUser {
    ...
    molecularDatabases {
        usable { ... }
        visible { ... }
    }
}
```
This leaves the `metaspace` filter to be applied to the top-level `molecularDatabases` query.

metaspace/graphql/schemas/moldb.graphql
LachlanStuart: I remember finding that some of graphql's behavior with optional parameters was completely non-sensical and poorly documented, so if the answer is "GraphQL wouldn't accept it", then that's fine. I thought the situation I faced was a bit different to this situation though, and we have had several significant updates to the graphql library since then...

> Arguments can be either required or optional.

By "Arguments", they only mean the top-level arguments supplied to a query/mutation, i.e. in `mutation { createMolecularDB(databaseDetails: CreateMolecularDBInput!) }`, `databaseDetails` is an argument, which means `!` means "required" and also implies non-nullable. However, `CreateMolecularDBInput.isPublic` is an input *field*, which should mean that `!` means "non-nullable".

The non-sensical situation I encountered is that arguments couldn't be both optional and non-nullable. Even if the schema had a non-null default, the caller could still override that by explicitly supplying `null`. I did not think this applied to fields though, but if you say it didn't work, I'll believe you.

metaspace/graphql/schemas/moldb.graphql
LachlanStuart: @richardgoater I have a strong preference for us to follow the existing patterns used by `allDatasets`/`dataset` and `allAnnotations`/`annotation`.

Consider this: Some day we may want to add text search and pagination. We may also want switches to show/hide archived databases, or add some additional access levels through feature flags (e.g. for beta access to METASPACE MS/MS). If we have 3 queries instead of 1, that means almost 3x as many code changes to update the API.

As I said in the Slack call - the 3 current filters represent 3 independently applicable `WHERE` clauses that could be added to an SQL query. It doesn't matter that there are some non-sensical combinations (e.g. querying usable but not visible databases). It's better to have a single configurable query than have a pile of pre-configured queries.

They definitely should not be nested under `currentUser`. Under that logic, a significant chunk of the schema would be moved...

One more suggestion regarding GraphQL design: I would suggest throwing out the old definition's `onlyUsable` parameter and moving fully to the pattern used by the `DatasetFilter` and `AnnotationFilter` structure. In particular:
* Having all the filters in a separate graphql input type gives more room for documentation
* Making unapplied filters be `null`/unspecified, instead of having some default value allows tri-state true/false/unfiltered logic, which means we don't have to use an awkward `only` prefix that requires mental gymnastics to reason about
* Filters that are directly applicable to a field should be named after that field (I can see us querying for `default: true` or `archived: false` in the not-too-distant future)

richardgoater:
```
It's better to have a single configurable query than have a pile of pre-configured queries.
```
citation needed? 🙂 

For `currentUser`, how does it negatively affect the schema? It should be another relation, similar to e.g. the user's groups.

`onlyUsable` was an improvement on something like `notUnusable`, but yes, I do agree that boolean flags like this are  hacky and didn't really like it being used here. I agree that the filter object is the most flexible option, and maybe flexibility is better when backwards compatibility is a strong constraint.

I'm actually wondering if the `usable` filter is needed by the front-end. I think this is now achieved by querying the user's groups.

LachlanStuart: > citation needed? slightly_smiling_face

I've justified this twice already in this thread:

> It's easier to maintain fewer queries, and there's a greater likelihood that new use cases can be served by an existing queries. allAnnotations is a great example of this - it's used for a variety of different views on the Annotations page (the main list, the isomers/isobars/other adducts/colocalized molecules lists, the isobars dropdown in the diagnostics section, CSV export, etc.) and has required very little new code to support each additional use case.

> Consider this: Some day we may want to add text search and pagination. We may also want switches to show/hide archived databases, or add some additional access levels through feature flags (e.g. for beta access to METASPACE MS/MS). If we have 3 queries instead of 1, that means almost 3x as many code changes to update the API.

If you're asking me to hunt down some blog post of some guy who agrees with me, then... no. The existence of opinions on the internet does not prove anything. There are opinions on the internet to both support and refute virtually every argument. We already have the benefit of experience with `allAnnotations` working well. We should not try to reinvent the wheel in a part of the system that is expensive to change when we already have a successful pattern that fills the need.
 
> For currentUser, how does it negatively affect the schema? It should be another relation, similar to e.g. the user's groups.

Most queries are somehow filtered based on who the current user is, does that mean that we should nest all of these queries into `currentUser` as well? Should we have `currentUser.datasets.visible`? Or should it be `currentUser.groups.visible.datasets.visible`, because some datasets are in groups?

I can't see any advantage to this line of thinking, but it adds complexity by adding more cross-module code dependencies. The codebase has been structured to minimze interdependencies between the "modules" so that refactoring stays easy and you don't have to jump between too many places when you want to fully understand one thing. If we nest `molecularDatabases` inside `currentUser`, we lose some of this, because the `user` module now has dependency to the `moldb` module.

##

If I sound frustrated, it's because you have not proposed a solution that solves the issue I raised (backwards compatibility) and have made several proposals to dramatically diverge from our existing patterns without giving adequate justification. It's not evident why you want that style of API, but to me it looks like it will cost us significantly more effort down the track when we want to add new features on top of it.

> I'm actually wondering if the usable filter is needed by the front-end. I think this is now achieved by querying the user's groups.

Seems a bit awkward, but I have no fundamental objections to doing that filtering in the front-end.

metaspace/graphql/schemas/moldb.graphql
```diff
@@ -43,8 +43,14 @@ input UpdateMolecularDBInput {
...
 }
 
+input MolecularDBFilter {
+  usable: Boolean
+  metaspace: Boolean
+}
+
 type Query {
-  molecularDatabases(onlyUsable: Boolean=false): [MolecularDB!]
+  allMolecularDBs(filter: MolecularDBFilter): [MolecularDB!]
```
LachlanStuart, line 52: There still needs to be a `molecularDatabases` for backwards compatibility. It can be deprecated and simply redirect to `allMolecularDBs`, but old code that depended on it shouldn't just suddenly break.

metaspace/graphql/schemas/moldb.graphql
richardgoater: Sorry to troll a little bit, it just looked like stating opinion as fact, that's all. That's something that annoys me, but I appreciate that text convos are easy to misread. I trust your opinion! But I feel our opinions should come with reasons to keep things fair. I see that you did this further down.

Most of this is just me being a GraphQL noob and asking questions, and I hope it shows that I'm at least getting into it. The advantage of `molecularDatabases` under `currentUser` would be that it provides a shortcut to this info, so e.g. if the usability or visibility criteria changes, the client doesn't need to change. We could probably get similar from a `molecularDatabases` filter, so I suppose it's just an ergonomic/aesthetic choice.

With the groups, there's no client-side filtering. Querying `currentUser`-> groups -> dbs` implicitly applies the filter, and as a bonus it's pre-grouped for the select component. The implicit filter has the same downside as above 👍

metaspace/graphql/schemas/moldb.graphql
intsco: @LachlanStuart @richardgoater Before I continue with changes in this PR, I wanted to have a mutual agreement that [these](https://github.com/metaspace2020/metaspace/pull/655/commits/69f338c8c906c833269df47381584052fed9b4f5) API changes is what we want.

I'll return `molecularDatabases` query for backwards compatibility. Any other comments?

richardgoater: Fine for me :+1:

intsco:
@richardgoater I'm merging this PR before yours (https://github.com/metaspace2020/metaspace/pull/638). It'll likely cause merge conflicts later when you'll be merging master into your branch. Let me know and I'll help you with that.

PR #662, opened 2020-08-24
Null FDR UI changes

richardgoater:
* blue row colour for annotation table
* hide FDR in related molecules/annotations


LachlanStuart (approved):
Web changes look good. Did you also check the CSV export handles null FDR values?

richardgoater:
@LachlanStuart Thanks, just checked and it should appear as an empty cell if the value is null, is that ok?
image
can't upload a csv to Github 🤷

LachlanStuart:
@richardgoater 👍

PR #666, opened 2020-08-31
Databases feedback

richardgoater:
* Show usable databases for upload
* Refresh filter list when user signs in/out
* Use radio buttons for annotation access controls
* Tweak isPublic column heading
* Fix DB name in filter for public annotations, though it reveals group in list
* Hide archived DBs until user searches filter list


metaspace/webapp/src/modules/MetadataEditor/MetadataEditor.vue
```diff
@@ -320,9 +316,7 @@ export default {
           metaspaceOptions.databaseIds = molecularDatabases.filter(d => d.default).map(_ => _.id)
         } else {
           for (const db of selectedDbs) {
-            if (this.molDBsByGroup.every(group =>
-              group.molecularDatabases.find(_ => _.id === db.id) === null,
-            )) {
+            if (molecularDatabases.find(_ => _.id === db.id) === null) {
```
LachlanStuart, line 319: I'm having trouble seeing how this works, both in the old and new code. `Array.prototype.find` returns `undefined` if nothing was found. I can't see any value that could be in `molecularDatabases` that wouldn't raise an error when `_.id` was evaluated, and would result in `.find` returning `null` to satisfy this condition. Even the falsy object `document.all`, which `== null`, doesn't `=== null`.

LachlanStuart (approved):
Looks good. I put a comment against one weird bit of code, but it's not actually a regression so I don't consider it a blocker.

metaspace/webapp/src/modules/MetadataEditor/MetadataEditor.vue
richardgoater: good spot, it shouldn't be trying to equal `null`. Should be easy to fix it.

PR #668, opened 2020-09-01
Fix python-client publish GH workflow

intsco:
The previous PR didn't solve the problem completely so here is a new one.


LachlanStuart (approved):

LachlanStuart:
It's OK to skip code review for testing these tweaks.

PR #671, opened 2020-09-04
Inline color scales and remove plotly.js dependency

LachlanStuart:
We only used Plotly.js for its colormaps, but as you can see by the `yarn.lock` diff, it was a huge dependency. It also is the reason we've had a security advisory for static-eval since GitHub launched the security advisories feature... That security advisory was recently re-opened. Fortunately Plotly.js is MIT licensed, just like METASPACE, so we're free to copy and modify it as long as we preserve the copyright text.

I've previously looked at alternatives but never found one that covered all of our existing options. e.g. D3.js (which we already depend on, and actually use) has many color scales built in, but not Portland or Hot.


metaspace/webapp/src/lib/createColormap.ts
```diff
@@ -12,7 +12,7 @@ const OPACITY_MAPPINGS: Record<OpacityMode, (x:number) => number> = {
 export default function createColormap(name: string, opacityMode: OpacityMode = 'constant',
   opacityScale: number = 1): number[][] {
   const { domain, range } = getColorScale(name)
-  const sclFun = d3.scaleLinear<d3.RGBColor>().domain(domain).range(range).clamp(true)
+  const sclFun = d3.scaleLinear<string>().domain(domain).range(range).clamp(true)
```
LachlanStuart, line 15: This was a bit weird. We've been passing `string`s instead of `d3.RGBColor`s this whole time, D3 just internally autodetected that they were color strings and made the interpolation magically work. I changed it to `string` to match reality.

metaspace/webapp/src/lib/getColorScale.ts
```diff
@@ -1,24 +1,180 @@
...
+  }
   if (name[0] !== '-') {
-    return extractScale(scales[name], 0, 1) // normal
+    return scales[name]
   } else {
     // inverted - reverse both arrays so that the domain is always in ascending order
-    const { domain, range } = extractScale(scales[name.slice(1)], 1, 0)
-    return { domain: reverse(domain), range: reverse(range) }
+    const { domain, range } = scales[name.slice(1)]
+    return { domain: reverse(domain.map(v => 1 - v)), range: reverse(range) }
```
richardgoater, line 178: Not sure that I understand this line, why wasn't it in the previous code?

richardgoater (approved):
This is a nice idea, will help with the new image viewer. Just added one question but it's not a blocker.

metaspace/webapp/src/lib/getColorScale.ts
richardgoater: ok, maybe I should have more explicitly that it looks like a double reverse with the new `map` call and `reverse` together. I get something like this if I try it in the console:
```
a = [0, 1]
(2) [0, 1]
b = a.map(v => 1 - v)
(2) [1, 0]
b.reverse()
(2) [0, 1]
```
or am I reading it wrong?

LachlanStuart: @richardgoater The goal is not to double-reverse, but to invert (so that the colors are mirrored) and reverse (to preserve increasing order). I.e. to turn `[0, 0.1, 1]` into `[0, 0.9, 1]`, not `[0, 0.1, 1]`. I'm not sure if it makes a difference with the limited source data. Not really worth the time to check, so I'm just preserving the old behavior.

PR #674, opened 2020-09-08
feat(python-client): Add custom databases management notebook

intsco:
The new notebook [link](https://github.com/metaspace2020/metaspace/blob/feat/custom-dbs-notebook/metaspace/python-client/example/create-custom-molecular-database.ipynb)


metaspace/python-client/example/api_examples.ipynb
```diff
@@ -101,25 +99,56 @@
...
+   ],
+   "source": [
+    "ds.database_details"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 10,
+   "metadata": {},
+   "outputs": [
```
LachlanStuart, line 141: Weird that the results have changed here... the number of annotations seems to have increased by ~50% according to the change in this plot.

Old plot:
image

New plot:
image

Maybe the old plot was generated first, then `.cumsum()` was added to the code but the plot not regenerated... Or maybe the old plot was generated with HMDB-2.5. IDK. The new plot looks valid compared to the website, so it's not an issue.

*HOWEVER,* now that there are 6 FDR==10% annotations, the `.value_counts()` produces values in the wrong order, causing the plot to loop backwards from 0.10 to 0.05 then back up to 0.20. I think this needs to be changed to something like `results.fdr.value_counts().sort_index().cumsum().plot()`...

LachlanStuart (changes requested):
Found one issue with the updated FDR plot. As this is currently our primary tutorial example notebook, I think it should be fixed.

All other changes look good

LachlanStuart (approved):

PR #675, opened 2020-09-09
Several back-end bug fixes

LachlanStuart:
* (Per slack discussion) use `PIL.Image(BytesIO(resp.content))` instead of `PIL(resp.raw)` to see if it fixes the intermittent bug while downloading images for colocalization
* Remove the redundant `@retry_on_exception()` in `get_ion_images_for_analysis`, because I found in the logs that failing downloads were actually attempted 9 times due to nesting of `@retry_on_exception()` functions
* Add a semaphore to smApi calls so that no more than 4 calls are made simultaneously. Also slightly refactor smApi calls so that they all go through one function
* Add a parameter `async_es_update` to the update dataset sm-api endpoint, and supply this parameter for all bulk DS updates. I found that this provides a huge speedup in the API call (no speedup to the queued index update though) - renaming a project that contains 170 datasets goes from 91s to 7.3s. This is a big deal, because the front-end actually waits for this to finish.
    * Initially I was thinking of applying this to all DS update calls, but this could introduce a race condition for direct DS updates (e.g. renames & private/public updates) - the UI could potentially show the user old data when they're returned to the datasets list. My gut tells me we'd have to debug this in the future if the direct updates were made async.
    * When `asyncEsUpdate` is true, the dataset only gets a partial update-by-query (via `ESExporter.update_ds`) instead of a full update (via `ESExporter.sync_dataset`). This means fewer fields are sync'd... This needs a sanity check - do you think this is ok?

### DIFF
--- a/metaspace/engine/sm/engine/png_generator.py
+++ b/metaspace/engine/sm/engine/png_generator.py
@@ -136,7 +136,6 @@ class ImageStoreServiceWrapper:
             h, w = mask.shape
             value = np.empty((len(img_ids), h * w), dtype=np.float32)
 
-        @retry_on_exception()
         def process_img(img_id, idx, do_setup=False):
             img = self.get_image_by_id(storage_type, 'iso_image', img_id)
             if do_setup:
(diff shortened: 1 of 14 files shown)


intsco (approved):
Looks good to me. Regarding your question about updating dataset documents vs indexing them. As far as I understand, we are talking about difference between `self._es.update` that updates a dataset by id and `self._es.update_by_query` that updates a dataset by query. I think both approaches should have the same result.

PR #684, opened 2020-09-26
feat(engine): Script for enabling versioning of S3 buckets

intsco:
@LachlanStuart do you think it's the right place for this script?


LachlanStuart (approved):
@intsco Yeah, I think this directory is fine. We don't really have enough codebase-independent maintenance scripts to warrant making a separate place for them.

PR #685, opened 2020-09-29
IndexError in all_annotations_images when all images are None

aeisenbarth:
I got an error for a specific dataset. The python client returns ions including "C20H28O2+K", but no ion images for that ion, causing an index error on the `non_empty_images` list. 

The following code reproduces it (fill in login details, set a breakpoint in sm_annotation_utils.py:1054):

```python
from metaspace.sm_annotation_utils import SMInstance
sm = SMInstance()
sm.login(email="XXX", api_key="YYY")
dataset = sm.dataset(id="2020-09-22_14h53m35s")
dataset.all_annotation_images(fdr=0.5, database=("HMDB-endogenous", "v4"), only_first_isotope=True)
```

Traceback:
```
  File "…/metaspace/sm_annotation_utils.py", line 1103, in get_annotation_images
    return self.isotope_images(
  File "…/metaspace/sm_annotation_utils.py", line 1054, in isotope_images
    shape = non_empty_images[0].shape
IndexError: list index out of range
```

It seems the case that the `images` list contains None is a valid situation (then no underlying bug?), but the case that None is returned for all is not considered. 

A fix could be to skip the code if the filtered `non_empty_images` list is empty. However I'm not sure about the intended return value. Originally with `scale_intensity = True`, None would be replaced by zeroed arrays (for which shape must be known), but for `scale_intensity = False` the return value would still contain None.

### DIFF
--- a/metaspace/python-client/metaspace/__init__.py
+++ b/metaspace/python-client/metaspace/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '1.8.1'
+__version__ = '1.8.2'
 
 from metaspace.sm_annotation_utils import (
     SMInstance,
--- a/metaspace/python-client/metaspace/sm_annotation_utils.py
+++ b/metaspace/python-client/metaspace/sm_annotation_utils.py
@@ -1051,14 +1051,15 @@ class SMDataset(object):
 
         if scale_intensity:
             non_empty_images = [i for i in images if i is not None]
-            shape = non_empty_images[0].shape
-            for i in range(len(images)):
-                if images[i] is None:
-                    images[i] = np.zeros(shape, dtype=non_empty_images[0].dtype)
-                else:
-                    lo = float(image_metadata[i]['minIntensity'])
-                    hi = float(image_metadata[i]['maxIntensity'])
-                    images[i] = lo + images[i] * (hi - lo)
+            if non_empty_images:
+                shape = non_empty_images[0].shape
+                for i in range(len(images)):
+                    if images[i] is None:
+                        images[i] = np.zeros(shape, dtype=non_empty_images[0].dtype)
+                    else:
+                        lo = float(image_metadata[i]['minIntensity'])
+                        hi = float(image_metadata[i]['maxIntensity'])
+                        images[i] = lo + images[i] * (hi - lo)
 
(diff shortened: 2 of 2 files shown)


LachlanStuart:
> It seems the case that the images list contains None is a valid situation (then no underlying bug?)

It is a valid situation, though only the recent custom database changes made this edge case possible. Previously the first isotopic image would always be non-None (so there would always be at least one non-None image, even when `only_first_isotope` is set), but now with `targeted` databases, it's allowed for the first image to be None. 

HMDB-endogenous is not a `targeted` database, so I'm not sure why it generated this annotation. This exception still needs to be fixed to support targeted databases.

> However I'm not sure about the intended return value. 

It's a bit of a mess. I would prefer it if None was always returned when there was no image, but backwards compatibility is always a concern. Does SpaceM rely on images being non-None, or can it handle both None and zeroed images?

LachlanStuart (approved):
~@aeisenbarth Do you rely on the ~
No 2nd comment intended here - I didn't realize GitHub had saved a draft comment when I clicked Approve

LachlanStuart:
@aeisenbarth Can you please also update the version number in `metaspace/python-client/metaspace/__init__.py`?

aeisenbarth:
> Does SpaceM rely on images being non-None, or can it handle both None and zeroed images?

No, it does not matter for SpaceM.

intsco:
>@aeisenbarth Can you please also update the version number in metaspace/python-client/metaspace/__init__.py?

@LachlanStuart If we want a Python package to be published after this PR is merged, we need to merge it into the release branch. Only then the GH actions workflow that performs publishing will be triggered.

LachlanStuart:
@intsco That's a bit of an issue, as most python-client releases occur out-of-sync with production deployments.

GitHub actions now [supports manual triggers for actions](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/). Are you happy for me to add that as a trigger, so that I can release python-client from master?

intsco:
@LachlanStuart Sure that would be useful, we can also release python-client from master automatically.

LachlanStuart:
@aeisenbarth This change has now been published to PyPI

PR #689, opened 2020-10-11
Fix "inconsistent annotations" error

### LINKED ISSUE #653: Bug: Inconsistent isobars/isobaricWith query results
These bug reports have been popping up a lot more often than expected - sentry reports 960 occurrences across >20 users. These errors aren't visible to the user - the inconsistent data is just hidden from them.

E.g.
```
Inconsistent annotations between Annotation.isobars and isobaricWith query results. Annotation 2020-07-19_11h28m19s_25087699 C45H76NO7P+Na+: C42H80NO8PK != C45H76NO7PNa,C42H80NO8PK,C42H79O10PNa
```

Occurs on this page: https://metaspace2020.eu/annotations?db_id=26&ds=2020-07-19_11h28m19s&fdr=0.5&q=25087699&sections=5 (dataset only visible to EMBL users)

Every time I've checked on these reports, it has happened with non-HMDB-v4 annotations. The diagnostics isobar selection list usually shows duplicate entries. I suspect the cause is that one of `Annotation.isobars` or `isobaricWith` is handled in a DB-specific way, and the other is handled with no DB filter.

LachlanStuart:
Fixes #653

Context: Two annotations are isobaric if they have overlapping m/z ranges for one or more of their ion images. This is a reflexive relationship - if A is isobaric with B, B is isobaric with A. The `Diagnostics` component relies on this reflexivity - some data comes from the list of isobars in the original annotation, and other data comes from a "all annotations that list (original annotation) as an isobar" query. Because it needs data from both queries for each isobar, it reports an error when the two lists don't match up.

The isobars query was missing a parameter for the database ID. This parameter is important because isobars are always calculated within the set of annotations for a single database. When annotation A was isobaric with annotation B from a different database, A wouldn't list B as an isobar, but the isobars query would find B anyway because it had no database filter. This lead to the error report that the isobars were inconsistent. Based on Sentry, this has happened >1200 times.

Additionally, Sentry has had trouble consolidating these errors because they use a different error message every time. This has gotten really spammy lately, because it keeps reporting them as new errors. I added a new `extraData` parameter to `reportError` so that the message would stay the same, but the data would still be sent. When `extraData` is supplied, the data appears near the bottom of the page on Sentry:

image

### DIFF
--- a/metaspace/webapp/src/api/annotation.ts
+++ b/metaspace/webapp/src/api/annotation.ts
@@ -188,10 +188,10 @@ export const relatedMoleculesQuery =
   }`
 
 export const isobarsQuery =
-  gql`query IsobarsQuery($datasetId: String!, $ionFormula: String!, $databaseId: String!) {
+  gql`query IsobarsQuery($datasetId: String!, $filter: AnnotationFilter!) {
     allAnnotations(
       datasetFilter: { ids: $datasetId },
-      filter: { isobaricWith: $ionFormula, databaseId: $databaseId },
+      filter: $filter,
       orderBy: ORDER_BY_FDR_MSM,
       sortingOrder: ASCENDING
     ) {
--- a/metaspace/webapp/src/api/annotation.ts
+++ b/metaspace/webapp/src/api/annotation.ts
@@ -36,6 +36,9 @@ gql`query GetAnnotations($orderBy: AnnotationOrderBy, $sortingOrder: SortingOrde
           metadataJson
           isPublic
         }
+        databaseDetails {
+          id
+        }
         isotopeImages {
           mz
           url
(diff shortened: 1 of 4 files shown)


LachlanStuart:
Sorry for the rebase after opening the PR. I found why CodeCov sometimes reports ridiculous results in PRs, at least. It seems that it only compares coverage against `master` if the latest commit from `master` is in the branch, otherwise it does a commit-by-commit diff and complains when e.g. the latest commit adds 3 lines but only 1 is covered by a test. It's a little bit frustrating, but good to finally have an explanation for why it seems to fail so often when you have a few small bugfix commits at the end of a PR.

richardgoater (approved):

PR #690, opened 2020-10-11
Reorganize msm_basic module and refactor job CRUD

LachlanStuart:
As part of the IBM Cloud change, I'm trying to keep the Spark code intact so that we can feature flag it and do side-by-side testing. I'm consolidating a bunch of shared code in the process. To facilitate this, I've split the `msm_basic` module into an `annotation` module to hold the non-spark-specific code, and `annotation_spark` to hold the spark-specific code.

Does this seem like a reasonable approach, and are you happy with the reorganization? 

In addition, I've refactored the CRUD code for `job`s a bit, to reduce the amount of code in `annotation_spark.annotation_job`. This also fixes a bug: `AnnotationJob._remove_annotation_jobs` didn't clean up images, so when removing a database from an existing dataset without `del_first` specified, the images would be left on disk.


ansible/aws/PRODUCTION_UPGRADE.md
```diff
@@ -315,4 +315,12 @@ once you're done.
...
\ No newline at end of file
+```
+
+## Deployment history
+
+[The Git tags](https://github.com/metaspace2020/metaspace/tags) can be useful for seeing what has been deployed. 
+
+The Git reflog on the server can also be used to check when specific commits were deployed: `git reflog --date=iso`
+
+Note that the cluster will always be deployed with the latest code from its configured branch.
```
LachlanStuart, line 326: Unrelated change. This docs update didn't seem big enough to warrant its own PR.

metaspace/graphql/src/modules/dataset/controller/Subscription.ts
```diff
@@ -49,7 +49,7 @@ async function waitForChangeAndPublish(payload: DatasetStatusPayload) {
 
   try {
     for (let attempt = 1; attempt <= maxAttempts; attempt++) {
-      logger.debug(JSON.stringify({attempt, action}));
+      logger.debug(JSON.stringify({attempt, ...payload}));
```
LachlanStuart: Unrelated changes: The "dataset updated" and "dataset deleted" graphql subscriptions seem to be a bit flakey - I often see a lot of log messages reporting all the retries it does while waiting for ES to update. I've made it log the whole payload here, because previously it was missing `ds_id` and `stage`, which made debugging a lot harder.

Also, I've increased the number of retries so that it gives up after ~33s instead of ~10s, in case it just wasn't waiting long enough

metaspace/graphql/src/modules/dataset/controller/Subscription.ts
```diff
@@ -41,15 +41,15 @@ async function waitForChangeAndPublish(payload: DatasetStatusPayload) {
...
+  const maxAttempts = 12;
 
   if (!KNOWN_ACTIONS.includes(action)) {
     return;
   }
 
   try {
     for (let attempt = 1; attempt <= maxAttempts; attempt++) {
-      logger.debug(JSON.stringify({attempt, action}));
+      logger.debug('Pushing status update to subscribers: ' + JSON.stringify({ attempt, ...payload }));
```
LachlanStuart, line 52: Unrelated changes: The "dataset updated" and "dataset deleted" graphql subscriptions seem to be a bit flakey - I often see a lot of log messages reporting all the retries it does while waiting for ES to update. I've made it log the whole payload here, because previously it was missing `ds_id` and `stage`, which made debugging a lot harder.

Also, I've increased the number of retries so that it gives up after ~33s instead of ~10s, in case it just wasn't waiting long enough

PR #695, opened 2020-10-14
[Custom Databases] UI changes for error messages

### LINKED ISSUE #694: Custom Databases error amendments
Writing up on request of @intsco.

The recent formula error is returned with the error type `malformed_csv`. We could consider that the file is a valid TSV, it just has some bad content. I'd like to suggest a new error type `bad_formulae` for this, I can use this key to display a more appropriate message.

However I think the `hint` field is a larger issue. It contains both plain text and nested JSON with incorrect quotes, so it is not completely human-readable or machine-parseable. I'd like to suggest moving the JSON section to a `details` field. I should be able to safely use this in the formatted messages on the front-end.

**Examples**
_newlines added for readability_

Current format:
```json
{
  "type":"malformed_csv",
  "hint":"Failed to parse some formulas\n{'line': 2, 'formula': 'C21H21O11+', 'error': 'ParseError(\\'Line 1: expected [0-9], \"(\", [A-Z]\\\\nC21H21O11+\\\\n          ^\\',)'}"
}
```
Suggested format:
```json
{
  "type": "bad_formulae",
  "hint": "Failed to parse some formulas",
  "details": [{"line": 2, "formula": "C21H21O11+", "error": "..."}]
}
```

richardgoater:
Resolves #694

### DIFF
--- a/metaspace/engine/sm/engine/molecular_db.py
+++ b/metaspace/engine/sm/engine/molecular_db.py
@@ -13,9 +13,16 @@ logger = logging.getLogger('engine')
 
 
 class MalformedCSV(Exception):
+    def __init__(self, message):
+        self.message = message
+        super().__init__()
+
+
+class BadData(Exception):
     def __init__(self, message, *errors):
-        full_message = '\n'.join([str(m) for m in (message,) + errors])
-        super().__init__(full_message)
+        super().__init__()
+        self.message = message
+        self.errors = errors
 
 
 class MolecularDB:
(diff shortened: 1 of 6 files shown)


metaspace/graphql/src/utils/smApi/databases.ts
```diff
@@ -18,7 +18,8 @@ export const smApiDatabaseRequest = async (uri: string, args?: any) => {
     if (valid_error_statuses.includes(content.status)) {
       throw new UserError(JSON.stringify({
         type: content.status,
-        hint: content.errors,
+        error: content.error,
+        details: content.details,
```
intsco, line 22: @richardgoater I hope this change doesn't break anything, this naming is better IMO.

metaspace/graphql/src/utils/smApi/databases.ts
```diff
@@ -5,7 +5,7 @@ import { snakeCase } from 'typeorm/util/StringUtils';
 import { smApiJsonPost } from './smApiCall'
 
 
-const valid_error_statuses = ['wrong_parameters', 'already_exists', 'malformed_csv'];
+const valid_error_statuses = ['wrong_parameters', 'already_exists', 'malformed_csv', 'bad_data'];
```
intsco, line 8: @richardgoater I added a new error type as `bad_data` because invalid can be not only formulas but values in other fields as well.

intsco (commented):
@richardgoater I addressed the problems you described in #694 with some modifications. Please review.

metaspace/engine/sm/engine/molecular_db.py
```diff
@@ -50,19 +57,28 @@ def to_dict(self):
...
+                errors.append({'line': line_n, 'row': row.values.tolist(), 'error': 'Empty value'})
+
         try:
             if '.' in row.formula:
                 raise InvalidFormulaError('"." symbol not supported')
             parseSumFormula(row.formula)
         except Exception as e:
-            csv_file_line_n = row.Index + 2
-            errors.append({'line': csv_file_line_n, 'formula': row.formula, 'error': repr(e)})
+            errors.append({'line': line_n, 'row': row.values.tolist(), 'error': repr(e)})
```
intsco, line 71: @richardgoater Mind the changes of the keys of error documents.

metaspace/graphql/src/utils/smApi/databases.ts
richardgoater: sorry dude, I think this might still be an issue. I'm not sure that I can safely turn these into formatted messages:
image

This is an example from the formula error:
image

I don't think it's possible to provide a message for this specific case from this info, but it might be enough to highlight the line and the user can try to figure it out. What do you think?

richardgoater: Example generic message:
image

LachlanStuart (approved):

PR #701, opened 2020-11-03
Prevent datasets from being un-deleted

### LINKED ISSUE #700: Prevent datasets from being un-deleted when deleted while still processing
I found 25 datasets in a broken state, with a row in the `public.dataset` table, but no matching row in the `graphql.dataset` table. They also have no `job`s or `annotation`s. They're semi-broken in the METASPACE UI, showing no submitter, raising an error on the Edit Metadata page and logging a warning in graphql.

I checked the logs for one of the datasets (2020-10-06_21h37m52s) - it appears that it was deleted after processing, but before off-sample classification finished. The likely cause seems to be the upsert logic in `Dataset.save` that can re-insert a deleted dataset if it's saved after deletion due to some ongoing job.

As it appears these datasets were un-deleted after deletion, I've manually deleted the following datasets to comply with the submitters' intentions. Here are the IDs in case there's any need to review the logs:
```
 2019-05-27_11h04m07s
 2019-05-27_11h05m01s
 2019-06-05_04h52m19s
 2019-06-05_05h59m51s
 2019-06-17_10h07m46s
 2019-08-04_20h55m26s
 2019-08-04_20h57m18s
 2019-08-04_20h58m13s
 2019-08-15_18h07m16s
 2019-08-26_21h49m20s
 2019-10-09_10h44m32s
 2019-10-17_15h32m14s
 2019-11-12_14h36m46s
 2020-02-24_23h19m00s
 2020-03-12_15h24m00s
 2020-04-24_16h56m26s
 2020-05-11_23h26m09s
 2020-06-02_10h23m59s
 2020-07-03_13h19m02s
 2020-09-04_00h03m18s
 2020-09-07_13h04m41s
 2020-09-19_23h59m52s
 2020-09-22_05h14m47s
 2020-10-01_19h22m06s
 2020-10-06_21h37m52s
```

The underlying bug needs a fix.

LachlanStuart:
Fixes #700 by preventing everywhere in sm-engine (except the tests) from actually inserting `public.dataset` rows. Normally these rows are only created in graphql.

In making this change, I found a lot of repeated code for creating datasets in tests, so I consolidated it into `create_test_ds`. This required moving the default `metadata` and `ds_config` values into constants, as it wasn't possible to inject fixtures that deep.

### DIFF
--- a/metaspace/engine/tests/test_sm_daemons.py
+++ b/metaspace/engine/tests/test_sm_daemons.py
@@ -4,9 +4,9 @@ import sys
 import logging
 from collections import OrderedDict
 from functools import partial
+from pathlib import Path
 from unittest.mock import patch
 import time
-from datetime import datetime
 
 import pytest
 from PIL import Image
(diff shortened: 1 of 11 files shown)


metaspace/engine/tests/test_sm_daemons.py
```diff
@@ -190,85 +187,70 @@ def test_sm_daemons(
     db = DB()
     es = ESExporter(db, local_sm_config)
 
-    try:
```
LachlanStuart, line 193: Turn off whitespace diffs in this file. I removed the try/finally, because the `rm -rf` isn't actually needed - `data_dir_path`/`/tmp/abc` isn't created as a result of these tests (although it is created in the segment_centroids tests).

Also, further down the file there was an unused cleanup of `annotate_daemon`/`update_daemon`, which weren't ever actually assigned. Daemon setup/teardown is now handled completely in `run_daemons`, so I think this was just leftover code from a previous refactor.

intsco (approved):
Cool, thanks for cleaning up the mess. Eventually, it turned out to be a bad idea to have one `Dataset.save` method instead of two separate `.insert` and `.update`.

PR #703, opened 2020-11-05
Update engine deps in docker startup & fix link in email

LachlanStuart:
Two small fixes:

* Make the sm-api/sm-update-daemon/sm-annotate-daemon reinstall pip dependencies on startup so that they stay in sync with `requirements.txt`. Bizarrely, the ubuntu docker image doesn't include `sudo`, so I had to update the Dockerfile to include it. The missing `sudo` doesn't actually break startup, but if you want the reinstall feature or to stop getting a warning on startup, you can rebuild your docker containers with these commands after this is merged (cc @richardgoater ):
```
docker-compose rm -fs sm-api sm-update-daemon sm-annotate-daemon
docker-compose build sm-api sm-update-daemon sm-annotate-daemon
docker-compose up -d sm-api sm-update-daemon sm-annotate-daemon
```

* Fix the link in the project publishing reminder email. I received an email advising me to go to https://metaspace2020.eu/project/null because `urlSlug` is nullable.

### DIFF
--- a/docker/sm-engine/Dockerfile
+++ b/docker/sm-engine/Dockerfile
@@ -22,6 +22,7 @@ RUN APT_INSTALL="apt-get install -y --no-install-recommends" && \
         git \
         nano \
         netcat \
+        sudo \
         && \
 # ----- Spark -----
     DEBIAN_FRONTEND=noninteractive $APT_INSTALL \
--- a/docker/sm-engine/start-common.sh
+++ b/docker/sm-engine/start-common.sh
@@ -18,7 +18,9 @@ else
   cd /opt/metaspace/metaspace/engine
 fi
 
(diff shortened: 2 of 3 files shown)


intsco (approved):

intsco (approved):
>Bizarrely, the ubuntu docker image doesn't include `sudo`
Maybe because it's considered insecure to run commands as root inside containers?

LachlanStuart:
> Maybe because it's considered insecure to run commands as root inside containers?

If so, it's fairly weak protection. Dockerd generally runs as root, and it lets containers run as root-owned processes and have root-level access to any shared volumes. I only have to use sudo because I explicitly made the container run as a user via `docker-compose.uid.yml` to prevent it from making annoying undeletable root-owned files in the shared volumes.

PR #704, opened 2020-11-10
Fix DS creation

LachlanStuart:
I had specifically tested whether `allow_insert=True` was needed on this line when preparing #701 , but I forgot that sm-engine doesn't auto-reload on code changes, so I was testing against old code when I checked it :facepalm:. 

Fortunately it appears no users have encountered this bug yet.


LachlanStuart:
Merging as it's a major bug and I need to get this to prod before any users encounter this. I'd still appreciate a review though.

intsco:
Looks good, too bad I missed it when reviewing #701

PR #705, opened 2020-11-10
Image viewer improvements

richardgoater:


richardgoater:
@LachlanStuart I'd like to submit this one for review and continue with saved states in a new PR. Guess it's better to break it up a bit and get these improvements deployed under flags. Also, I think I broke most of the focus states on the site in the last deployment, would be good to fix that.

LachlanStuart:
The behavior of the ion intensity sliders in this PR is very different to my expectations, based on the workflow and MSiReader UI that Måns demonstrated to us. I'm not sure what is by design and what is a bug, so I'll just explain everything that surprised me:

* The numbers at the left/right ends of the scale show the min/max selectable values, not the currently selected values. I would expect these numbers to show the values of the dimmest/brightest represented pixels (i.e. they'd change as you move the sliders), because that would help understand the colors being rendered in the image.
* Clicking the padlock locks the min/max selectable values, not the slider position or the selected values used for rendering. E.g. If you select an annotation with a `maxIntensity` of 1e8, drag the slider down to 1e7, turn on the lock, then click on some other annotation (e.g. one with intensity 1e6), then the slider's maximum selectable value will stay at 1e8, but slider will move all the way back to the default position, changing the actual selected value to 1e8 again. I'd expect that clicking the lock would preserve the selected values, not the range of selectable values.
* Clicking the min/max intensity numbers and typing in a new value only changes the range of the sliders, not the actual selected values. This means that if you moved the right slider to the half-way point and entered 4e8, the brightest pixel would be 2e8. I would expect that if I enter a value, that value would become the brightest shown pixel value, resetting the slider position if needed.
* Clicking the lock buttons while multiple layers are selected only syncs the min/max selectable values. I'd expect it to sync the actual min/max values used for colorization.

It seems really counter-intuitive to me to have a UX that's centered around changing the minimum/maximum values of sliders, rather than changing the actual values of the sliders. Have you run all this by Måns?

richardgoater:
Thanks for the feedback, Måns agrees with the points and I don't mind implementing it here. Resetting the sliders would be much easier with an event-sourcing API, let's see how it goes otherwise.

LachlanStuart (approved):
Code changes look good. As it's feature flagged and stable, it's fine to merge even if this isn't the final functionality. It's up to you whether you merge now and start a new PR or you update this one.

richardgoater:
Thanks, think I will merge 👍

PR #713, opened 2020-11-30
IBM Cloud Annotation

LachlanStuart:
## Overview
* Adds Lithops implementations for annotation, colocalization and ion thumbnails. It's off by default, but selectable by using the `lithops` feature flag in webapp.
    Note: there have been some significant changes since the `pywren-annotation-pipeline` repository - mostly cleaning up & refactoring:
    * `annotation.io` contains a suite of utilities for saving/loading variables from cloud objects
    * The pipeline cache has been completely re-done. It was only useful for development purposes, so it's turned off by default.
    * I implemented a rudimentary centroids cache - it caches the calculated peaks and pickled FDR objects for an entire job. Cache items are keyed by a hash based on the `ds_config` JSON, so it's unlikely that the cache will be reused between different users as every user has their own set of DBs. Calculating peaks is much faster in Lithops, so hopefully this won't be a big problem.
* Updates Python to 3.8
* Updates Spark to 3.0.1
* A guide for setting up an IBM Cloud account, and debugging strategies when using Lithops is in `docs/Lithops.md`
* Some stuff isn't yet complete:
    * Sci test isn't working in CircleCI (runs fine locally though)
    * There's an issue with Lithops jobs stalling (still troubleshooting with Lithops)
    * Some jobs don't correctly get marked as failed
    * I haven't proven large datasets work yet (waiting on some Lithops fixes)

## Testing

Automated testing is unfortunately lacking - There's an integration test for the centroids calculation, but the rest is only covered by the sci-test.

Manually, I've run it against [46 datasets on Staging](https://staging.metaspace2020.eu/datasets?prj=ed1e8514-3321-11eb-ac6a-072233edf08d&q=-Spark). They were picked with a variety of file sizes, including the top 5 biggest .ibd files that were on Staging. Results pending.

## Deployment notes:
* Reprovision with `ansible-playbook -i env/staging provision/web.yml -t miniconda,ibm_cloud`
* The IBM Cloud VPC instance needs to be manually created and have an SSH key from the config added to its `authorized_keys` - we don't have our final account yet, so I haven't done this yet
* Python 3.8/Spark 3.0.1 required new AMIs - make sure I don't forget to update the configs before deploying prod

## Notable changes to the general project architecture:
* `sm-lithops-daemon` - I think I previously mentioned just adding it to the update daemon. I realized that we don't really know the long-term stability of Lithops, especially considering that it does a lot with multiprocessing and has used `process.exit(1)` instead of exceptions in the past, so I thought it best to isolate it in its own process.
* `utils.perf_profile` - classes & a `contextmanager` for collecting performance information, logs and arbitrary extra information (e.g. number of peaks, number of segments, etc. for later debugging), saved to the database. Additionally, a `SubtaskProfiler` for combining the performance data from many parallel processes into one entry
* `Executor` - a wrapper around the Lithops executor class(es) to make it less complicated to use. Automatically switches between Cloud Functions and the VM depending on memory requirements, retries with more memory if there's an OOM, etc. I'm hoping to get some of these changes upstreamed into Lithops, but that's a discussion for when I have more time.
* `utils.log_capture.capture_logs` - a `contextmanager` that allows code to collect log messages into a List. Used by `PerfProfiler`.


ansible/aws/templates/all/vars.yml.template
```diff
@@ -44,7 +44,7 @@ miniconda_env:
   channels:
     - conda-forge
   dependencies:
-    - python=3.6
+    - python=3.8
```
LachlanStuart, line 47: Sadly it looks like we're stuck at 3.8. PyArrow's native libraries don't compile with 3.9 yet. However, the real killer is that `elasticsearch-dsl` 5.x.x will probably never be updated, and it has the following issue:

```
  /usr/local/lib/python3.8/dist-packages/elasticsearch_dsl/search.py:97: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
    elif isinstance(doc_type, collections.Mapping):
```

ansible/roles/miniconda/tasks/environment.yml
```diff
@@ -11,14 +11,14 @@
   args:
     creates: '{{ miniconda_prefix }}/envs/{{ miniconda_env.name }}'
   register: miniconda_env_create
-  when: "miniconda_env != '' and conda_env_list_out.stdout.find('{{ miniconda_env.name }}') == -1"
+  when: "miniconda_env != '' and conda_env_list_out.stdout.find(miniconda_env.name) == -1"
```
LachlanStuart, line 14: Just fixing an Ansible warning here - Jinja template syntax shouldn't be used inside `when` expressions

ansible/roles/sm_lithops_daemon/defaults/main.yml
```diff
@@ -0,0 +1 @@
+sm_lithops_daemon_app_name: sm-lithops-daemon
```
LachlanStuart, line 1: I think I previously mentioned just adding it to the update daemon. I realized that we don't really know the long-term stability of Lithops, especially considering that it does a lot with multiprocessing and has used `process.exit(1)` instead of exceptions in the past, so I thought it best to isolate it in its own process.

docker/sm-engine/Dockerfile
```diff
@@ -27,30 +28,31 @@ RUN APT_INSTALL="apt-get install -y --no-install-recommends" && \
...
     $PIP_INSTALL \
         setuptools && \
     apt-get clean -y && \
     apt-get autoremove -y && \
     rm -rf /var/lib/apt/lists/* /tmp/* ~/.cache/*
 
 # --- Engine dependencies ---
-COPY requirements.txt /requirements.txt
+COPY metaspace/engine/requirements.txt /requirements.txt
+COPY docker/sm-engine/ci-requirements.txt /ci-requirements.txt
```
LachlanStuart, line 55: I changed the Docker build context to the root of the repository, so that this could grab the real `requirements.txt` file, to make sure they don't go out of sync.

The downside of having the build context include the whole repository is that Docker has to actually copy the entire repository into a container when it rebuilds the image. This adds ~10s on my computer. It's only for building though, not starting existing containers

metaspace/engine/.pylintrc
```diff
@@ -145,7 +145,11 @@ disable=print-statement,
         too-few-public-methods,
         broad-except,
         bad-continuation,
-        fixme
+        fixme,
+        # no-member is a nightmare with Pandas due to this bug that prevents the generated-members setting from
+        # allowing you to exclude Pandas classes: https://github.com/PyCQA/pylint/issues/2498
+        no-member
```
LachlanStuart, line 151: IDK why, but pylint seemed much more aggressive than normal in this PR. I had never seen this lint error before, but it started appearing in existing code. Maybe it's a result of the type annotations...

metaspace/engine/conf/config.json.template
```diff
@@ -24,10 +24,15 @@
   },
   "services": {
     "img_service_url": "{{ sm_img_service_url }}",
+    "img_service_public_url": "{{ web_public_url }}",
```
LachlanStuart, line 27: This public URL is needed so that functions can download images from IBM Cloud

ansible/aws/README.md
```diff
@@ -6,7 +6,7 @@ Software requirements for the developer machine:
 * boto3>=1.5
 * boto>=2.48
 * pyyaml
-* [Ansible>=2.4](http://docs.ansible.com/ansible/intro_installation.html#latest-releases-via-pip)
+* [Ansible==2.8.6](http://docs.ansible.com/ansible/intro_installation.html#latest-releases-via-pip)
```
LachlanStuart, line 9: This was the version installed on Prod in the `sm` conda environment, and I think these scripts will need changes to work with anything else...

Ansible 2.6.2 (installed by apt outside of the `sm` conda environment) didn't work under Python 3.8 because it has some weird behavior in Jinja2 scripts that cause it to interpret Python objects as their string representation

Ansible 2.10.3 (latest) seemed to have different behavior with the `ec2_metadata_facts` action. I gave up trying to get it to work as it was taking too long.

metaspace/engine/docs/Lithops.md
```diff
@@ -0,0 +1,132 @@
+# Environment setup
```
LachlanStuart, line 1: Can you think of a better place to put docs? I'm considering making a top-level `/docs` directory to assemble general project documentation that isn't suitable for project `README.md` files. `/ansible/PRODUCTION_UPGRADE.md` also feels like it's in the wrong place.

metaspace/engine/sm/engine/dataset_locker.py
```diff
@@ -1,24 +0,0 @@
-from contextlib import contextmanager
-from zlib import adler32
-
-from psycopg2 import connect
-
-LOCK_KEY = 894951
-
-
-class DatasetLocker:
```
LachlanStuart, line 9: Not removed, just renamed to `DbMutex` so that I can re-use it for other purposes.

metaspace/engine/sm/engine/image_store.py
```diff
@@ -0,0 +1,174 @@
...
+from PIL import Image
+from requests.adapters import HTTPAdapter
+from scipy.ndimage import zoom
+
+from sm.engine.util import SMConfig, retry_on_exception
+
+logger = logging.getLogger('engine')
+
+
+class ImageStoreServiceWrapper:
```
LachlanStuart, line 17: Not significantly changed. I just pulled this out of `png_generator.py` to reduce dependency inclusion so that e.g. the colocalization code (which never creates PNGs) can run without pypng.

metaspace/engine/sm/engine/ion_thumbnail.py
```diff
@@ -22,19 +26,19 @@
...
-        *args, use_distance_from_centroid=False
+    'image-medoid': lambda i, m, h, w: _thumb_from_image_clusters(i, m, h, w, use_centroids=False),
+    'image-centroid': lambda i, m, h, w: _thumb_from_image_clusters(i, m, h, w, use_centroids=True),
+    'pixel-mean-intensity': lambda i, m, h, w: _thumb_from_pixel_clusters(
+        i, m, h, w, use_distance_from_centroid=False
     ),
-    'pixel-distance': lambda *args: _make_thumbnail_from_pixel_clusters(
-        *args, use_distance_from_centroid=True
+    'pixel-distance': lambda i, m, h, w: _thumb_from_pixel_clusters(
+        i, m, h, w, use_distance_from_centroid=True
```
LachlanStuart, line 41: No functional change here, mypy just didn't like the way `*args` was used

metaspace/engine/sm/engine/off_sample_wrapper.py
```diff
@@ -36,29 +34,6 @@ def base64_images_to_doc(images):
     return images_doc
 
 
-def retry_on_error(num_retries=3):
```
LachlanStuart, line 39: Pylint didn't like this duplicated code (it's almost identical to `util.retry_on_exception`). No idea why it didn't complain in the old codebase, as I hadn't touched either function. Mmaybe the Python 3.6->3.8 update changed pylint's behavior...

The only difference was the retry timings - this implementation had much higher delays. I added a configurable wait time to `util.retry_with_exception` so that it can be extra patient for the below `call_api` function.

metaspace/engine/sm/engine/sm_daemons.py
```diff
@@ -131,6 +167,9 @@ def delete(self, ds):
         self._db.alter('DELETE FROM dataset WHERE id=%s', params=(ds.id,))
 
     def _send_email(self, email, subj, body):
+        if not self._sm_config['services'].get('send_email', True):
```
LachlanStuart, line 170: Fixing an old bug - it hasn't been possible to disable emails via config for several years.

metaspace/engine/sm/engine/util.py
```diff
@@ -5,13 +5,11 @@
...
 import os
 from copy import deepcopy
-from datetime import datetime
 from pathlib import Path
 import random
 from time import sleep
+from typing import Dict
 
-from sm.engine.db import ConnectionPool
-from sm.engine import molecular_db
```
LachlanStuart, line 14: I had to refactor a bit here so that no Lithops modules imported anything that imported psycopg2, just so that we don't need to include that dependency in the Docker image. It's not strictly necessary, it's just that docker image size affects function startup speed, so it's good to try to keep it small.

I made a `utils` directory - in future we should try to put small util functions as separate files in that directory to keep dependencies isolated to the functions that actually use them.

metaspace/engine/sm/engine/util.py
```diff
@@ -103,47 +101,33 @@ def get_ms_file_handler(cls, ms_file_path):
         )
 
 
-def create_ds_from_files(ds_id, ds_name, ds_input_path, config_path=None, meta_path=None):
```
LachlanStuart, line 106: `create_ds_from_files` was moved into its own file in the `utils` directory

metaspace/graphql/src/modules/dataset/controller/Mutation.ts
```diff
@@ -228,14 +229,18 @@ const setDatabaseIdsInInput = async (
 };
 
 const createDataset = async (args: CreateDatasetArgs, ctx: Context) => {
-  const { input, priority, force, delFirst, skipValidation } = args,
-    datasetId = args.datasetId || newDatasetId(),
-    datasetIdWasSpecified = args.datasetId != null;
+  const { input, priority, force, delFirst, skipValidation, useLithops } = args;
+  const datasetId = args.id || newDatasetId();
+  const datasetIdWasSpecified = args.id != null;
```
LachlanStuart, line 234: These changes were to fix the ability for admins to specify an ID while creating a dataset

metaspace/graphql/src/migrations/1605218230747-PerfProfile.ts
```diff
@@ -0,0 +1,27 @@
+import {MigrationInterface, QueryRunner} from "typeorm";
+
+export class PerfProfile1605218230747 implements MigrationInterface {
+
+    public async up(queryRunner: QueryRunner): Promise<any> {
+        // The schemas were out of sync with the migrations, and these two changes were missing.
+        // Fortunately they're idempotent and hopefully shouldn't cause any issues on computers that already have them
+        await queryRunner.query(`ALTER TABLE "public"."molecular_db" ALTER COLUMN "created_dt" SET NOT NULL`);
```
LachlanStuart, line 8: FYI, this schema change is why there are small changes in some of the graphql tests

metaspace/engine/sm/engine/annotation/formula_validator.py
```diff
@@ -2,27 +2,60 @@
 Classes and functions for isotope image validation
```
LachlanStuart, line 2: This file was refactored so that it could be reused in the Lithops code. There were 2 big structural changes needed:

* Spark code iterates centroid peaks in m/z order, collecting a buffer of images for each `formula_i`, whereas the Lithops code iterates centroid peaks in `formula_i` order so that it doesn't need to keep a buffer. I extracted `iter_images_in_sets` to do the buffer management for Spark.
* Lithops code needs to be able to flush images to storage to minimize memory usage, so it uses the iterator-based functions directly. Spark code (including a lot of tests) uses the `formula_image_metrics()` wrapper function, which returns the full results all at once.

metaspace/engine/sm/engine/annotation_lithops/annotation_job.py
```diff
@@ -0,0 +1,340 @@
...
+
+class LocalAnnotationJob:
+    """
+    Runs an annotation job from local files and saves the results to the filesystem.
+
+    As a developer convienence, if a list of integers is given for `moldb_files`,
+    then it will try to connect to the configured postgres database and dump the formulas
+    for the specified databases.
+    Otherwise, this can be used to run the pipeline without any external dependencies.
+    """
```
LachlanStuart, line 164: I added this filesystem-only wrapper so that we could easily set IBM up to use this repository if they continue experimenting.

LachlanStuart:
Here's an update on testing that the results are equivalent. Just to reiterate what I've said in person - I don't think you need to do extensive testing yourself, but this would certainly benefit if you reviewed the testing I've done. I'm happy to do further checks if you want.

Out of the 50 datasets that I reprocessed with Lithops, 38 successfully annotated:

#### 6 failed due to missing input data

Either they were in temporary buckets, or they had been cleaned up.

#### 6 failed due to reasons related to their size. 
All of these were 30GB+. There are several changes out of my hands that I'm waiting on, which I expect will resolve the issues with these datasets:
* Lithops reworking how they handle IBM VMs to improve disk performance (currently limited to 50MB/s)
* Lithops limiting parallelism on VMs to prevent OOMs when many VM tasks are run at once
* Another change coming late January that I can't talk publicly about

#### 8 had annotations with MSM discrepancies higher than 0.001
This was 49 out of ~92000 annotations, 0.05%. The worst MSM discrepancy was 0.115. I found 2 bugs in the Spark implementation contributing to this:
* Under some conditions, `compute_metrics(iso_images_sparse, formula_ints)` seems to get a `0`s in the `formula_ints` when there's also an empty image. It was hard to get these values out of the 4th peak. The spatial & spectral metrics normalize this array, which caused the weights for each image to change slightly when computing those metrics. This seemed to happen consistently even after reprocessing.
    * Normally I'd try to fix Lithops to match Spark so that annotated results don't change. However, I couldn't figure out the cause. 
* Under some conditions the 4th isotopic image would be missing some pixels. This happened inconsistently, affecting different annotations if reprocessing. 
    * I suspect this is due to the way DS segments are handled - if the 4th peak crosses a DS segment boundary, part of the image from the first segment will be yielded, triggering metrics to be calculated without waiting for the part from the second segment. However, I've previously suspected this and tried to make a reliable reproduction, but couldn't, even with very granular dataset segments (IIRC ~10kB/segment). I guess it's not worth investigating further at this point.
    * Just to show you what makes me suspect this, here are the images:
        * [First annotation with spark](https://user-images.githubusercontent.com/26366936/102530007-030c9280-40a1-11eb-87af-cd751e5fbc8a.png)
        * [Reprocessed with Spark](https://user-images.githubusercontent.com/26366936/102530059-16b7f900-40a1-11eb-9ab8-753bc39d34d5.png)
        * [Annotated with Lithops](https://user-images.githubusercontent.com/26366936/102530140-32bb9a80-40a1-11eb-8e25-87ca199cb53d.png)
    * In case you're still curious, I think this is so rare and hard to reproduce because most datasets have 32-bit m/z values (often stored as 64-bit, but not using the extra bits). I couldn't reproduce this issue on Untreated, which I know only uses 32 out of the 64 bits. However, these MSM discrepancies are happening on a dataset that uses the full 64-bits. With 32-bit m/z values, they're much more clumped together, so DS segment boundaries would be much less likely to slice through the 4th isotopic image in a way that substantially affects the MSM.

I don't know if these cover all the issues - I only dug deeply into 5 of the 49 annotations. However, I think I've seen enough to determine that the number of MSM issues in Lithops is lower than our existing tolerance for MSM issues.

#### 16 had MSM discrepancies resolved by reprocessing the spark versions

I mentioned this in Slack. Not sure why MSM calculation has changed. This affected roughly half of the datasets, no visible pattern. Some datasets had >1000 discrepancies. 

I re-deployed the `master` code and cleared the centroids cache so I could check whether the issue came from this branch. The results were the same as the results from within this branch.

#### 0.07% of annotations changed by more than 1 FDR level

67 annotations spread across 14 of the 38 successfully annotated datasets. FDR has some randomness, so IMO this is acceptable.

#### Lithops produced slightly fewer annotations

* 91103 annotations appeared in both Lithops and Spark datasets
* 434 appeared only in the Lithops datasets
* 626 appeared only in the Spark datasets

There doesn't seem to be any systematic bias - some datasets have slightly more with Spark, others have slightly more with Lithops. My gut feeling is this is just due to random chance with the FDR cutoff values.

intsco (approved):
In general, the code looks quite similar to the one we had in the prototype with addition of extra utilities and abstractions for COS, Lithops, performance measurement and debugging. So the main advantages and unfortunately some disadvantages remained the same. I assume most of limitations originate in the general distributed nature of computations. If something should be computed in distributed manner, the code for it will get longer and harder to read.

Just to name a few concerns, although none of them can be considered major:
* The same logic implemented in code became longer and got more abstractions.
* While reading the code one has to jump around the file quite a bit to follow the flow.
* Due to Cloud Functions being naturally memory constrained, lots of computations have to be chunked into really small bits.
* As one of the consequences of the above, the code contains many hard coded size limits and custom calculations based on sizes of inputs.

On the positive side, the new computation backend comes with lots of potential, scalability for a really big jobs (large databases), running multiple jobs in parallel, easier debugging, and less setup and operations work.

metaspace/engine/conf/config.json.template
```diff
@@ -53,6 +57,57 @@
...
+    "sm_storage": {
+{#
+Each entry in this block is an array of [Bucket name, Prefix]
+e.g. "imzml": ["my-bucket","files/go/here"] means ImzML files will be stored under cos://my-bucket/store/things/in/here/
+The Prefix is necessary and should not have a leading or trailing slash
+#}
+      "imzml": ["{{ sm_lithops_cos_bucket_imzml }}", "imzml"],
+      "moldb": ["{{ sm_lithops_cos_bucket_data }}", "moldb"],
+      "centroids": ["{{ sm_lithops_cos_bucket_data }}", "centroids"],
+      "pipeline_cache": ["{{ sm_lithops_cos_bucket_temp }}", "pipeline_cache"]
```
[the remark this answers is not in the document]
intsco: Ok, I read the code below and understood why having two buckets (persistent and temp) would be useful but I'm not sure about a separate bucket for imzml files.

metaspace/engine/docs/Lithops.md
LachlanStuart: I'll do that in a separate PR. I've made #730 so I don't forget.

ansible/aws/templates/all/vars.yml.template
```diff
@@ -200,8 +215,8 @@ sm_activate_venv: source {{ miniconda_prefix }}/bin/activate {{ miniconda_env.na
 sm_activate_nvm: source {{ ansible_env.HOME }}/.nvm/nvm.sh && nvm use
 
 spark_key_file: ~/.ssh/ubuntu_spark
-spark_version: 2.3.3-bin-hadoop2.7
-spark_mirror: "{{ vault_spark_mirror }}"
+spark_version: "3.0.1-bin-hadoop2.7"
+spark_mirror: https://archive.apache.org/dist/spark/spark-3.0.1/
```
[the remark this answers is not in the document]
LachlanStuart: I think we discussed this IRL since then, but to recap: I successfully reran annotation with the Spark pipeline on staging without having to customize Spark. I suspect we're just not using S3 in Spark anymore. IIRC the last place we used it was `formula_centroids.py` for the cache, but that has since been changed to use boto directly.

metaspace/engine/conf/config.json.template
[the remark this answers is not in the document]
LachlanStuart: Just realized I hadn't responded to this (GitHub hid it in a collapsed section as there are so many comments).

The reasoning for having imzml files in a separate bucket is basically for billing. COS only allows selection of storage class on a bucket level (vs. S3 which allows per-object storage class selection). 

With the Smart tier, we pay significantly less if the bucket is considered to be infrequently accessed (`monthly # API calls + MB retrievals < MB stored`). The imzml bucket will probably be considered to be infrequently accessed. Moldb and centroids probably won't be considered as infrequently accessed, but who knows.

metaspace/engine/conf/config.json.template
```diff
@@ -99,35 +154,42 @@
...
         },
         "engine.db": {
-            "handlers": ["console_debug"],
             "level": "INFO"
         },
+        "engine.lithops-wrapper": {
+            "level": "DEBUG"
+        },
+        "lithops": {
+            "level": "DEBUG"
```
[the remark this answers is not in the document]
LachlanStuart: I agree, but I see it as necessary until we're confident in Lithops' stability. Lithops-daemon logs to a separate file, so hopefully it won't interfere much with investigating issues that arise in other daemons.

metaspace/engine/docker/lithops_ibm_cf/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+RUN grep -E "^(c?pyMSpec|c?pyImagingMSpec|numpy|scipy|pandas|msgpack|pyimzML|pyarrow|scikit-learn|Pillow)" /tmp/requirements.txt > /tmp/requirements-lithops.txt
+RUN pip install --no-cache-dir -r /tmp/requirements-lithops.txt
+
+# create action working directory
+RUN mkdir -p /action \
+    && mkdir -p /actionProxy \
+    && mkdir -p /pythonAction
+
+ADD https://raw.githubusercontent.com/apache/openwhisk-runtime-docker/8b2e205c39d84ed5ede6b1b08cccf314a2b13105/core/actionProxy/actionproxy.py /actionProxy/actionproxy.py
+ADD https://raw.githubusercontent.com/apache/openwhisk-runtime-python/3%401.0.3/core/pythonAction/pythonrunner.py /pythonAction/pythonrunner.py
```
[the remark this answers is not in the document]
LachlanStuart: This file will be able to be simplified at some point by inheriting from a [pre-made Lithops image](https://github.com/lithops-cloud/lithops/blob/master/runtime/ibm_cf/Dockerfile.python38). However, those images only very recently started being [published to Docker Hub](https://hub.docker.com/r/lithopscloud/ibmcf-python-v38/tags?page=1&ordering=last_updated). Probably a task for a separate PR.

metaspace/engine/sm/engine/annotation_lithops/annotation_job.py
```diff
@@ -0,0 +1,340 @@
...
+        sm_config = sm_config or SMConfig.get_conf()
+        self.storage = Storage(config=sm_config['lithops'])
+        sm_storage = sm_config['lithops']['sm_storage']
+
+        self.imzml_cobj = _upload_if_needed(imzml_file, self.storage, sm_storage, 'imzml')
+        self.ibd_cobj = _upload_if_needed(ibd_file, self.storage, sm_storage, 'imzml')
+        if isinstance(moldb_files[0], int):
+            self.moldb_defs = _upload_moldbs_from_db(moldb_files, self.storage, sm_storage)
+        else:
+            self.moldb_defs = _upload_moldbs_from_db(moldb_files, self.storage, sm_storage)
```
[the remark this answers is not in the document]
LachlanStuart: Good catch. Fixed.

PR #716, opened 2020-12-07
Hotfix to prevent quick search losing focus while typing

LachlanStuart:
This change prevents the annotation table from stealing focus when focus is on an input element.

I've manually tested these cases:
* Typing in the quick search bar - doesn't steal focus
* Typing a molecule name into the molecule filter - doesn't steal focus
* Changing FDR - doesn't steal focus
* Removing a filter - table takes focus
* Loading the page - table takes focus
* Changing between pages of annotation via keyboard or mouse - table re-takes focus after re-appearing
* IE11 - the patch seems to work, thanks to polyfill.io including an `Element.closest` polyfill


LachlanStuart:
I'll deploy this to prod now. A review would still be appreciated when you have time.

PR #717, opened 2020-12-09
Fix MSM filter

### LINKED ISSUE #714: MSM filter is not working

richardgoater:
Closes #714

### DIFF
--- a/metaspace/webapp/src/modules/Annotations/AnnotationTable.vue
+++ b/metaspace/webapp/src/modules/Annotations/AnnotationTable.vue
@@ -515,7 +515,7 @@ export default Vue.extend({
     },
 
     setCurrentRow(rowIndex, rows = this.annotations) {
-      if (this.$refs.table) {
+      if (this.$refs.table && rows.length) {
         this.$refs.table.setCurrentRow(null)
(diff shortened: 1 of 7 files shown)


richardgoater:
Decided to upgrade the UI component for it as well. This unfortunately leaves no good candidate for testing the `InputFilter` component. I tried to get `compoundName` working but debouncing is breaking it - messing with jest timers didn't seem to do anything.

LachlanStuart (approved):
LGTM.

> messing with jest timers didn't seem to do anything.

Yeah, I've never had any luck with `jest.useFakeTimers()`. Every time I've tried to use it, it seemed like the code was still using the actual `setTimeout` instead of the mocks. 

You can directly mock `setTimeout`. I've done it successfully a few times, IIRC by just overwriting `global.setTimeout` with a hand-written function. It's always questionable whether it's worth adding such complexity for a test though.

richardgoater:
Thanks, tried a couple of things but still didn't work. Ideally `debounce` would be mocked, but can't find how to do it with `lodash-es`. Switching to other copies e.g. `lodash/debounce` or `lodash.debounce` seemed to break other stuff. Will skip it for now.

PR #720, opened 2020-12-11
New feature popups

richardgoater:


richardgoater:
image
This is logging from the previous method when clicking on annotation table rows. It is checking for popups and hitting local storage all the time. This PR should save some battery at least.

LachlanStuart:
Please remember to do screenshots!

image image image

The Ion Image Channels one seems to be clipped by the title bar. Happens to me both on Chrome & Firefox.

## Performance

If you care about performance, always use the profiler to confirm your suspicions. It's super easy to get distracted by things that look slow, but aren't actually slow, and this is definitely one of those cases. 

Since you mentioned clicking through the annotation table, let's look at that. I did a profile of 10 clicks and zoomed in on one.
image

The biggest source of both CPU usage (/battery drain) and jank is that 60ms click event. There are small bursts of non-janky CPU usage/battery drain afterwards for animation, but let's assume we're not planning to reduce non-janky animation for the sake of battery consumption and focus on that 60ms.

Finding the performance impact of something easy. Ctrl+F and type "`NewFeaturePopup`" and it highlights it in the timeline:

image

1 occurrence, totaling 0.052ms out of 60ms, i.e. a 0.087% contribution to the jank. If you search the whole timeline, there's an average of 1 more `NewFeaturePopup` callback per in between each "click" event, but these are usually ~0.02ms, presumably because the callback is throttled, so the battery consumption contribution might be as high as 0.12%.

To put that in perspective, these are several other things that are very quickly obvious when you look at that 60ms click event:
* ~6ms (~10%) is spent rendering `DatasetInfo`'s `ElTreeNode` children, which are in the `Metadata` section that was closed the whole time.
* ~6ms (~10%) is used by `renderScaleBar`, mainly because it's internally using `toDataURL`.
* ~4ms (~10%) is spent re-rendering `CandidateMoleculesPopover` elements, even though they haven't have changed. It looks like it's because Vue treats `slot`s like an always-changing prop, and the `CandidateMoleculesPopover` gets some `slot` content from `AnnotationTable`, which re-renders due to the click.

Fixing any one of the above would have taken the same amount of time or less, but would have had ~100x more impact.

## Conclusion

It's frankly quite frustrating that you were so mortified by my code that you seemingly spent ~3 days (based on commit times) rewriting it and on Monday proclaimed "it's much better now". There was never a problem to begin with. You could have spent that time more effectively. I hope I've shown you enough to convince you to profile before you optimize.

Anyway, the code in this PR looks fine to merge once you fix the bug with the Ion Image Channels popup.

LachlanStuart (approved):

richardgoater:
There have been screenshots on Slack for the general look and feel, and I have voiced my disagreement on providing screenshots in PRs. In this PR they would not have shown that the popups display when scrolling into view, and you would not have found the bug that you did (thanks for spotting it).

I honestly didn't know that you wrote the original implementation. My comments were about the code, not the coder. I apologise that I did not make this clear and that my words caused you offence.

I'm really happy with the new implementation, it's nicely declarative and removes some error-prone things:

* Hard-coding the route
* Using arbitrary selectors
* Using string templates for content

I've spoken about predictability recently - that was the goal here. The project now has a reference implementation for a global persisted queue, proof that DOM nodes can be shared far across the component tree, and a nice IntersectionObserver hook. I don't think it was a waste of time.

PR #722, opened 2021-01-03
Update docker-compose configs

LachlanStuart:
As discussed last year, hopefully this helps to reduce fragmentation between dev setups and make it easier to keep everything in a working state.

Changes:
* merges compose files
* moves elasticsearch
* removes sm- prefix
* removes "standalone mode"
* removes unused logstash, kibana and adminer containers
* removes several unused test config files
* uses directory volumes when possible (for macOS/Windows support)
* moves example config files to live next to the real files so they're easier to keep up to date

Notes:
* postgres still runs as a different user and `chown`s the pgdata directory
* `install-dbs.sh` currently doesn't work - there seem to be multiple issues with the DB files: HMDB has formulas with `.` and `(C5H8)n` that fail to parse, and other databases seem to have empty fields. For new staff I suggest we just set them up with DB dumps that have datasets/annotations/users removed.
* [There's something weird with sci-test and CircleCI](https://app.circleci.com/pipelines/github/metaspace2020/metaspace/1032/workflows/5ed0e2cb-f5e0-42a9-a599-32a282087037/jobs/12936) - I SSH'd in and the checked out copy had the 2 new files in `engine/conf` in a "deleted" state in git. No idea why this is happening - probably a CircleCI caching bug. I'll have to continue investigating. I manually restored those files and sci-test ran fine though.

Migration might be different for everyone. Here are the main points:
* Using `docker-compose.custom.yml` to maintain a stable dev environment is still expected, but now it should be easier to diff against the reference version.
* Add `PGDATA` and `ESDATA` variables to your `.env` file
* Diff your engine/graphql config files against the `.docker.js(on)` files to see if anything needs to be changed
* The file owner for all files in `DATA_ROOT` should be your user, except for `pgdata` (which postgres automatically fixes on startup). Use `sudo chown -R <your username>:<your username> <your DATA_ROOT>` to fix it if needed.
* ElasticSearch's data directory is now outside the container, so you'll need to reindex to see datasets after updating
* Isotope storage has been moved to a subdirectory in the engine config - it should regenerate by itself, but you may want to clean up the old dirs.

**If this doesn't run out of the box for you after updating your `.env` and syncing your config files with the `.docker.js(on)` files, LMK.**


docker/.env.example
```diff
@@ -23,4 +24,4 @@ WWW_PORT=8999
 # You can find your user ID and group ID with the `id` command in Linux.
```
richardgoater, line 24: Is this advice outdated with the removal of `docker-compose.uid.yml`?

docker/docker-compose.yml
```diff
@@ -1,39 +1,53 @@
-version: '2'
+version: '3.5'
 
 services:
 
   elasticsearch:
-    image: elasticsearch:5.5.2
+    build: ./elasticsearch
```
richardgoater, line 6: This doesn't work when starting from scratch, there's no `Dockerfile` in the dir?

LachlanStuart: My bad, I didn't notice they weren't added to git. Should be in there now.

richardgoater:
One general comment, did you consider removing the `sm-` prefix from the subfolders in `./docker`? I guess it might mess up the diffs.

docker/.env.example
LachlanStuart: `docker-compose.uid.yml` is merged into `docker-compose.yml`. It's on-by-default. It can be commented out in a `.custom.yml` file if needed, but if you do need to comment it out to get things working (even after the `chown` step in the PR description), please speak up so I'm aware of any cross-platform issues.

LachlanStuart:
> One general comment, did you consider removing the sm- prefix from the subfolders in ./docker? I guess it might mess up the diffs.

There's both pros and cons to having the prefix, but ultimately these directories are so infrequently used I didn't think it worth the opportunity cost to dwell on it.

docker/.env.example
richardgoater: Sorry, GH wouldn't let me target the specific line, `docker-compose.uid.yml` is still mentioned in the same comment.

LachlanStuart: Ah, good catch. I've updated the comment

LachlanStuart:
@richardgoater I've added the `.dockerignore` file as discussed. Seems to work well locally - the ~20s delay I had when rebuilding containers is now ~0s. Docker doesn't scan parent directories for the `.dockerignore`, and only the engine services need the repository as a context, so I've only whitelisted the `requirements.txt` files and `/docker/sm-engine`.

I've also changed all instances of `${SOME_VAR}/:/some/mount/point` to not have the `/` after the variable. The issue I mentioned on Slack was caused by me not having `PGDATA` defined, causing Docker to mount `/` and the postgres container to try to fix permissions on my root directory...

richardgoater (approved):
Think I'm up and running now, although I haven't tested uploads.

docker/setup-dev-env.sh
```diff
@@ -2,12 +2,18 @@
...
+
 
 # set up molecular DB
-docker-compose run --rm sm-api /install-dbs.sh
-./fetch-mol-images.sh
 
-docker-compose up -d
\ No newline at end of file
+# This is currently broken as the databases no longer pass validation
+# docker-compose run --rm api /sm-engine/install-dbs.sh
```
intsco, line 16: If this worked, it would be the easiest way to have a complete setup. Otherwise, it requires an additional step of loading a database dump which isn't described anywhere.

Do you want me to fix the database import at least for HMDB v4?

docker/README.md
```diff
@@ -93,3 +93,13 @@ Watching application logs:
 Rebuilding the Elasticsearch index:
 
 * `docker-compose run --rm sm-api /rebuild-es-index.sh`
+
```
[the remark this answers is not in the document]
intsco: Adminer and Kibana mentioned in the Development tools section aren't available anymore.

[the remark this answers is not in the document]
LachlanStuart: I've updated the commands, fixed a few errors that were logged in `rebuild-es-index.sh`, and replaced the Adminer/Kibana URLs with instructions on how to access Postgres/ElasticSearch with external tools.

metaspace/engine/conf/config.docker.json
```diff
@@ -127,5 +135,8 @@
       "+": ["+H", "+Na", "+K"],
       "-": ["-H", "+Cl" ]
     }
+  },
+  "colocalization": {
+    "enabled": true
```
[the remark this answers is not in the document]
LachlanStuart: That change is still in the IBM Cloud PR. I'll update this file when I merge master into that PR.

IIRC technically this config setting isn't actually used in master, but that is a bug.

docker/setup-dev-env.sh
LachlanStuart: I think it would be best to fix the CSV exports of HMDB to make it valid with the new code. I had a brief look at changing the code to relax the parsing so that HMDB, etc. was still accepted, but it felt like it was making the code worse at every step...

Loading a database dump is only really an option for staff, so I don't think it should be in the public instructions. However, it is a good way to get a large amount of example data for realistic testing in a dev environment, so I think we should do it internally.

PR #725, opened 2021-01-07
Copy ion and m/z to clipboard

richardgoater:
Screenshot 2021-01-07 at 01 40 27

Unfortunately the tooltips do not work perfectly in Safari, let me know if this is an issue.


LachlanStuart (approved):

metaspace/webapp/src/components/CopyButton.vue
```diff
@@ -0,0 +1,74 @@
...
+      <duplicate-icon class="h-full w-full text-gray-400 hover:text-gray-500 sm-copy-icon" />
+    </button>
+    <p class="leading-none m-0">
+      <span
+        v-if="copied"
+        key="copied"
+      >
+        Copied!
+      </span>
+      <slot v-else>
```
LachlanStuart, line 23: I think this should be a `span`... It doesn't make sense to have a nested `slot` AFAICT

PR #726, opened 2021-01-08
Make clipping popups less annoying

richardgoater:
Moved out the way of the controls for channels:
image

Same style and logic applied to single images:
image

Additionally fixes intensity editing.


metaspace/webapp/src/modules/ImageViewer/IonIntensity.vue
```diff
@@ -14,44 +14,22 @@
...
+        :class="{
+          'font-medium text-red-700': status === 'CLIPPED',
+          'font-medium text-blue-700': status === 'LOCKED',
+          'cursor-default': tooltipDisabled
+        }"
+        @mouseover="showPopover"
+        @mouseleave="hidePopover"
+        @focus="showPopover"
+        @blur="hidePopover"
+        @click.stop="editing = true"
```
LachlanStuart, line 29: If you click to edit the intensity and leave the mouse in place while this button fades out, the popover stays visible even after the intensity is updated and the user clicks away

I think this `click` handler should call `hidePopover` to prevent it from getting stuck open when the button isn't around to receive `blur`/`mouseleave` events

LachlanStuart (approved):
There was one tiny issue, but it otherwise looks good. Sorry for the delayed review.

metaspace/webapp/src/modules/ImageViewer/IonIntensity.vue
richardgoater: Good spot, thanks. Yet another state machine issue, I should have known 👍

richardgoater: p.s. this was behaving fine in Chrome, and in Safari the popover would disappear if the mouse is moved. Pretty annoying 😕

PR #728, opened 2021-01-15
Upload custom mol DB files in binary mode

LachlanStuart:
Alyona encountered this error:
```
GraphQLException: {"type":"malformed_csv","error":"Malformed CSV: 'utf-8' codec can't decode byte 0xb1 in position 1: invalid start byte"}
```

The file was valid on the local filesystem, but when I re-downloaded it from S3 it had an invalid `B1` byte instead of a `±` UTF-8 character (represented in UTF-8 with 2 bytes: `C2 B1`). Switching to reading the input file as bytes seems to have fixed it.

### DIFF
--- a/metaspace/python-client/metaspace/__init__.py
+++ b/metaspace/python-client/metaspace/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '1.8.2'
+__version__ = '1.8.3'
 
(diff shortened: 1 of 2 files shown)


intsco (approved):
Thanks for fixing it!

PR #731, opened 2021-01-18
Uppy dataset upload

richardgoater:
Caveat with this PR is the project now depends entirely on S3. Optional branch `feat/dev-local-s3` contains a demo of a local S3 server, but it needs some work on the configuration: https://github.com/metaspace2020/metaspace/compare/feat/uppy-dataset-upload...feat/dev-local-s3


metaspace/webapp/src/components/UppyUploader/UppyUploader.vue
```diff
@@ -0,0 +1,188 @@
...
+        return 'UPLOADING'
+      }
+      return 'PENDING'
+    }
+
+    const files = computed(() => {
+      const files = uppy.getFiles()
+      if (props.requiredFileTypes) {
+        return props.requiredFileTypes.map(ext => {
+          const matchingFile = files.find(f => f.extension === ext)
```
LachlanStuart, line 126: Occasionally we get `.imzml`, `.IMZML` and `.IBD` files uploaded. The file selection dialog and Uppy seem to happily accept them. However, this case-sensitive comparison seems to cause the UI element for that file to stay in the empty state. It doesn't seem to prevent upload/submission, but it's confusing when it happens.

metaspace/webapp/src/modules/MetadataEditor/UploadPage.vue
```diff
@@ -223,36 +295,23 @@ export default {
...
+
+      const [file] = result.successful
+      const { name, extension, uploadURL } = file
+
+      this.inputPath = createInputPath(uploadURL, this.uuid)
+
+      const dsName = name.slice(0, name.lastIndexOf('.'))
       Vue.nextTick(() => {
         this.$refs.editor.fillDatasetName(dsName)
       })
```
LachlanStuart, line 312: The new behavior waits for upload to finish before populating the dataset name field. If the upload is slow enough, the user may manually enter this field, but their input would be overwritten once the upload finished. Is it possible to return to the old behavior of populating this field immediately once upload starts?

metaspace/webapp/src/modules/MetadataEditor/UploadPage.vue
```diff
@@ -167,54 +205,88 @@ export default {
...
-      return this.uploadedUuid != null && !this.isSubmitting
+    uuid() {
+      return this.storageKey.uuid
     },
 
-    fineUploaderDataTypeConfig() {
-      const activeDataType = this.$store.getters.filter.metadataType
-      return (activeDataType in DataTypeConfig) ? DataTypeConfig[activeDataType] : DataTypeConfig.default
+    enableSubmit() {
+      return this.uuid != null && !this.isSubmitting
```
LachlanStuart, line 237: It seems to be possible to submit (and cause a server error) while the upload is still running

LachlanStuart (approved):
The new UI looks and feels really nice! Great job!

I hit some minor issues, which I've listed in the comments. Aside from those, this PR looks good.

metaspace/webapp/src/components/UppyUploader/UppyUploader.vue
richardgoater: thanks for the extra info 👍

metaspace/webapp/src/modules/MetadataEditor/UploadPage.vue
richardgoater: good suggestion, I'll try to sort it

metaspace/webapp/src/modules/MetadataEditor/UploadPage.vue
richardgoater: thanks for spotting this, sorry I missed it.

richardgoater:
Think I'm missing the config to test the full flow locally, I'd to test it on staging before merging.

richardgoater:
@LachlanStuart I think the code has changed quite significantly, so might be best to review the update if you don't mind. It's also deployed on Staging 👍

LachlanStuart (approved):

Issue #579, opened 2020-04-28
Review of Icons


richardgoater:
Some icons are quite low resolution and could be cleaned up, and it would be better to use fewer icon sets in general.

A way to import SVG as a component is being added (https://github.com/metaspace2020/metaspace/commit/7b510f0a229b22bc891d25e4e2ef173d48c2b3ea), this will help a lot.

Additionally, a component abstraction for icon buttons in the Annotations view could improve code quality. Would remove the unscoped class `av-icon-button` and bring Tailwind into play.

LachlanStuart:
@lmacielvieira Please triage this - move it to "Ideas" if you want to do it at some point, or close it if you're not interested.

PR #735, opened 2021-01-21
Webapp UX tweaks and fixes

LachlanStuart:
# Annotations page

* Fix: Fixed broken isomers/isobars alert icon & popup on Molecules section
* Fix: Fixed broken Diagnostics tour step (tour text wouldn't appear until you scrolled down, appeared in wrong position)
* Fix: Fixed a console error from `ClippingNotice` getting a null `intensity` prop when loading a saved URL
* Optimization: Moved the `peakChartData` query into Diagnostics component(s) so that it would only be run if that section was actually open. Also enabled client-side caching on this query.
* Optimization: Made the Metadata section not render if it's not open
* Optimization: Moved the molecular formula in the table into its own component to prevent expensive re-renders when selecting annotations.
* UX: Moved the filter button for the molecular formula to the right end of the cell be consistent with other columns. This also shrinks the hoverable area 
image
* UX: Switched the filter icons to SVG. I tried coloring them when not clickable, but wasn't too happy with how noisy the grid became. WDYT?
image
    Though this is a non-change, I left them as SVG as they fit better with the `refactoring-ui` icon pattern.

# Elsewhere

* Fix: run `deref-schema` on Docker container startup to generate `metadataRegistry.ts`, etc. so that a fresh git checkout can run the containers without a manual step
* Optimization: Added HTTP caching settings so that ion/optical/thumbnail images can be cached by the browser
* UX: Added loading spinners to the Dataset Summary page as it's quite slow on prod
* UX: Fixed the issue that caused the Edit Metadata page to automatically append `?mdtype=Imaging MS` to the URL, which created an unwanted history entry.
* UX: Changed the Edit Metadata page to use `$router.back()` instead of `$router.push('/datasets')` so that the previous filter & scroll position are preserved.

Each change is in its own commit, in case you prefer to review it change-by-change.


metaspace/webapp/src/lib/useFilter.ts
```diff
@@ -0,0 +1,29 @@
+import { computed, ref, watch } from '@vue/composition-api'
+import { FilterKey } from '../modules/Filters/filterSpecs'
+
+/**
+ * Returns a property for getting/setting a specific filter from $store.getters.filter.
+ * @param $store
+ * @param filterName
+ */
+export default ($store: any, filterName: FilterKey) => {
+  // For some reason, whenever $store.getters.filter is recalculated, ALL downstream watchers are triggered,
```
richardgoater, line 10: sounds like our discussion about Vue "magic" 👍

metaspace/webapp/src/modules/Annotations/AnnotationTable.vue
```diff
@@ -285,9 +260,9 @@
 </template>
 
 <script>
-import { renderMolFormulaHtml } from '../../lib/util'
 import ProgressButton from './ProgressButton.vue'
-import CandidateMoleculesPopover from './annotation-widgets/CandidateMoleculesPopover.vue'
+import AnnotationTableMolName from './AnnotationTableMolName.vue'
+import FilterIcon from '../../assets/inline/refactoring-ui/filter.svg'
```
richardgoater, line 265: I support the change to SVG, it looks much sharper 👍 

Just to note that the `refactoring-ui` folder is intended for icons that come from the Refactoring UI icon set. I plan to add the whole set to the folder as part of the UI documentation. I will push a commit that moves the new icon to `inline` folder, where it should still work as a Vue component.

Fun fact is I tried to make a filter icon that looks like it comes from Refactoring UI, but I'm not 100% sold on it:
image

metaspace/webapp/src/modules/MetadataEditor/MetadataEditPage.vue
```diff
@@ -99,7 +99,7 @@ export default {
 
   methods: {
     handleCancel() {
-      this.$router.push('/datasets')
+      this.$router.back()
```
richardgoater, line 102: Thanks for changing this. I'm just thinking of the corner case that it will break if the `/edit` url is opened in a new tab. We could consider this rare enough not to worry about though 👍

metaspace/webapp/src/modules/ImageViewer/useIonImages.ts
```diff
@@ -144,7 +144,7 @@ function createComputedImageData(props: Props, layer: IonImageLayer) {
         ),
       }
     }
-    return null
+    return { min: 0, max: 0 }
```
richardgoater, line 147: Sorry I'm not sure what this does, is this the `ClippingNotice` change? I think it is supposed to return `null` if the image is not ready, I'm concerned about knock-on effects from changing the API.

richardgoater (commented):
Thanks a lot for doing these 👏 

I think there are a few discussion points so I will save approval for now.

I support the idea of consistency in the annotation table, but the change does mean that in many cases the value cannot be filtered while checking the candidate molecules. I suspect this is on purpose, so we might want to check it with Theo and other users.

richardgoater:
p.s. regarding "coloring [filter icons] when not clickable`, I agree that it is noisy. It might work if the icon clipped the content of the table cell, but maybe we don't want to go that far.

metaspace/webapp/src/modules/ImageViewer/useIonImages.ts
richardgoater: Ah, yes it is. I will see if there's an alternative solution, feels like the entire notice should not show if this is `null`.

richardgoater: Ah, yes it is. I will see if there's an alternative solution, feels like the entire component should not show if this is `null`.

metaspace/webapp/src/modules/MetadataEditor/MetadataEditPage.vue
LachlanStuart: That actually was my motivation for changing it from `$router.back()` to `$router.push('/datasets')` years ago. `history.back()` has several unmitigable edge-cases, and can often cause bad UX if there are multiple ways to access a page. However, in this case it'd take significantly more code to make the datasets page restore the previous filters & scroll to the just-edited dataset, which didn't seem worth it.

metaspace/webapp/src/modules/Annotations/AnnotationTableMolName.vue
```diff
@@ -0,0 +1,74 @@
...
+    <candidate-molecules-popover
+      class="cell-span"
+      placement="right"
+      :possible-compounds="annotation.possibleCompounds"
+      :limit="10"
+      :isomers="annotation.isomers"
+      :isobars="annotation.isobars"
+    >
+      <span
+        v-html="molFormulaHtml"
```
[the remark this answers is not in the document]
LachlanStuart: I had actually planned to, but it slipped my mind. Fixed.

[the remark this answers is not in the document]
richardgoater: No worries, thanks 👍

LachlanStuart:
> I support the idea of consistency in the annotation table, but the change does mean that in many cases the value cannot be filtered while checking the candidate molecules. I suspect this is on purpose, so we might want to check it with Theo and other users.

This was actually in response to feedback from our team, it has just been on my todo list for a long time. IIRC it was Don (Måns' predecessor) who complained that it was too easy to accidentally click them when selecting rows, and when I proposed moving them to a consistent position he said that would help. Someone complained more recently in GM about the popovers appearing too often, and I think reducing the size of the triggers by excluding the invisible filter buttons should help with that.

metaspace/webapp/src/modules/Annotations/AnnotationTable.vue
LachlanStuart: I think your version looks a bit better, but by comparison it makes the settings icon stand out as an eyesore. If there's a matching replacement for the cog icon, I'd say it'd be worth doing.

richardgoater: Might look something like this (plus mono alternative):
image

I created this issue a while ago: #579 - we could make a separate PR to look at all the icons on this page.

metaspace/webapp/src/modules/ImageViewer/useIonImages.ts
richardgoater: This case used to be handled by the `loading` API in the menu item component. The popover has to wrap the menu item, leaving the clipping notice unprotected. Adding an extra `loading` condition sorts it out 👍

LachlanStuart: Cool, thanks for the fix.

richardgoater (approved):

PR #737, opened 2021-01-26
fix: Moldb import for docker setup

intsco:


LachlanStuart (approved):
LGTM, but please also update the URLs in the [ansible vars.yml template](https://github.com/metaspace2020/metaspace/blob/master/ansible/aws/templates/all/vars.yml.template#L122) for the updated databases. IMO there's no need to update the other listed databases (HMDB 2.5, PAM DB, etc) - if anything we should remove them from the template as we don't really actively maintain them.

PR #739, opened 2021-01-28
Reorganize sm-engine files

LachlanStuart:
I'm not aware of any big changes in sm-engine at the moment, so this seemed like an opportune time to do some refactoring. If I'm wrong about this, let me know. These changes are mostly IDE-based refactoring and took barely any time, so I'm happy to wait until other projects are finished and re-do this.

Some of these organizational changes were started in the IBM Cloud PR, but I didn't want to add any unnecessary changes in that PR as it was already huge. This PR updates existing code to match the new patterns introduced in that PR.

Feel free to critique the changes or suggest more. I haven't really discussed these changes yet, so this is as much of an RFC as it is a PR.

**Changes:**
* Consolidate daemon-related code into `daemons/`
  * `sm/engine/daemons/lithops_daemon.py` -> `sm/engine/daemons/lithops.py`
  * `sm/engine/sm_daemons.py` - split into several files:
    * `SMAnnotateDaemon` -> `sm/engine/daemons/annotate.py`
    * `SMIndexUpdateDaemon` -> `sm/engine/daemons/update.py` (and renamed to just `SMUedateDaemon`)
    * `DatasetManager` -> `sm/engine/daemons/dataset.py`
  * `sm/engine/daemon_action.py` -> `sm/engine/daemon/actions.py` 
* Continue consolidating code directly related to the annotation pipeline into `annotation/`
  * `sm/engine/fdr.py` -> `sm/engine/annotation/fdr.py`
  * `sm/engine/isocalc_wrapper.py` -> `sm/engine/annotation/isocalc_wrapper.py`
  * `sm/engine/png_generator.py` -> `sm/engine/annotation/png_generator.py`
* Move pipeline steps that happen after annotation into `postprocessing/`
  * `sm/engine/colocalization.py` -> `sm/engine/annotation/colocalization.py`
  * `sm/engine/ion_thumbnails.py` -> `sm/engine/annotation/ion_thumbnails.py`
  * `sm/engine/off_sample_wrapper.py` -> `sm/engine/annotation/off_sample_wrapper.py`
* Continue consolidating utils and splitting `util.py` into separate files in `utils/`
  * For context: the problem with having an all-inclusive `util.py` file is that it makes it hard to import e.g. `SMConfig` without also causing `DB`, `psycopg` and many other libraries to be imported. This is a big issue for Lithops as we've excluded unnecessary libraries like psycopg from the runtime.
  * `SMConfig`, `init_loggers()` from `util.py`  -> `sm/engine/config.py` (config seemed more like core functionality than a utility, so I didn't put it in the utils dir)
  * `retry_on_exception()` from `util.py`  -> `sm/engine/utils/retry_on_exception.py`
  * `sm/engine/db_mutex.py` -> `sm/engine/utils/db_mutex.py`
* **Unrelated bugfixes**
  * Re-add `sm/__init__.py` so that Lithops can correctly serialize the `sm` module
  * Change `test_config.json` to use `localhost` mode instead of `serverless` mode so that tests can run locally. This was an oversight in the Lithops PR

It shouldn't be necessary to review individual code changes, as almost everything was done through IDE-based refactoring.  The only exceptions are the 2 unrelated bugfixes, and the string-based references to modules in `@patch()` calls in the tests, which had to be manually updated.


intsco (approved):
Looks much better now! The project evolved and this structure matches the current state of the project.
Last time I tried to run the tests after the lithops changed had been merged and had some configuration issues locally. So I fetched these changes and run the tests locally. All good.

Since we are doing refactoring here, maybe it's time to get rid of `metaspace/engine/utils/metabolights.py` file.

sergii-mamedov (approved):
Look perfect for me.
Just one question, why is the first character in log strings is always a space?

LachlanStuart:
@sergii-mamedov 
>  Just one question, why is the first character in log strings is always a space?

I'm not actually sure. I've never followed this convention and also have never heard it discussed. 

It appears a leading space is only used for log messages that also begin with `" [x]"`, `" [v]"`, and `" [*]"`, which seem to represent failure, success and neutral log messages. It also only seems to be used in the daemon/queue-related code. As far as I'm aware nobody relies on this pattern, and log levels convey the same information, so I think it's safe to ignore that pattern.

PR #742, opened 2021-02-01
Local S3 for Development

richardgoater:
Key idea is the `aws` section of any config is optional and should only be present when using AWS.


richardgoater:
the issue in the previous commit name was because of old config, it shouldn't affect new setups 🤞

metaspace/graphql/s3Client.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export const getS3Config = () => {
+  if (config.aws) {
+    return {
+      credentials: getCredentials(config.aws),
+      region: config.aws.region,
+    }
+  }
+  const { endpoint, httpOptions, s3ForcePathStyle } = config.s3 || {}
+  return {
```
richardgoater, line 21: I thought config was being spread here, I could fix it you like.

docker/docker-compose.yml
```diff
@@ -198,6 +198,16 @@ services:
     #  - "api:172.17.0.1"
     #  - "graphql:172.17.0.1"
 
+  storage:
+    image: minio/minio:RELEASE.2021-01-16T02-19-44Z
+    command: server /data
```
intsco, line 203: As the expected bucket name isn't created on startup, the very first upload fails with `NoSuchBucket: The specified bucket does not exist`.
One way of handling it here would be
```
    entrypoint: sh
    command: -c 'mkdir -p /data/sm-engine-dev && /usr/bin/minio server /data'
```

intsco (approved):
LGTM. I am not quite confident in my review of the JS code though.

metaspace/graphql/config/development.docker.js
```diff
@@ -12,9 +12,31 @@ config.defaults = {
 }
 
 config.upload = {
-  destination: '/opt/data/metaspace/uploads',
+  bucket: 'sm-engine-dev',
+  moldb_prefix: 'databases',
 }
 
+config.s3 = {
+  endpoint: 'http://localhost:9000',
```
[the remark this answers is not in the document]
richardgoater: In my testing, `endpoint` needed to be the external address, one that works from the browser like `localhost`. The `httpOptions.proxy` property provides the reference to the internal `storage` host name 👍

docker/docker-compose.yml
richardgoater: Good spot and nice idea, I'll add it.

metaspace/graphql/s3Client.ts
LachlanStuart: It's fine either way

metaspace/engine/sm/engine/molecular_db.py
```diff
@@ -72,7 +74,9 @@ def _validate_moldb_df(df):
 
 def read_moldb_file(file_path):
     try:
-        moldb_df = pd.read_csv(file_path, sep='\t', dtype=object, na_filter=False)
+        bucket_name, key = split_s3_path(file_path)
+        buffer = get_s3_bucket(bucket_name).Object(key).get()['Body']
+        moldb_df = pd.read_csv(buffer, sep='\t', dtype=object, na_filter=False)
```
LachlanStuart, line 79: @intsco This no longer supports local paths, which breaks `scripts/import_molecular_db.py`.

LachlanStuart (approved):
I gave it a try and it works great. 

There's one issue with `read_moldb_file` not supporting local files anymore, which should be fixed. However, it's fine if that comes in a follow-up PR.

intsco:
@richardgoater When do you plan to merge this PR? I want to start working on my next task based on the changes in this PR.

richardgoater:
@intsco I can merge it now

PR #745, opened 2021-02-02
Add linting to GraphQL

### LINKED ISSUE #227: Add linting to enforce code style in sm-graphql

LachlanStuart:
Fixes #227 
Closes #596 (I made a separate PR as the creator of a PR can't add approve/reject it)

Compared to Webapp, I've enabled the "requiring-type-checking" preset which does some extra checking, primarily validating that promises are used properly. I've removed the Vue-specific rules and also disabled the following:

* `lines-between-class-members` - TypeORM `model.ts` files get bigger without any readability gain when this is enforced
* `camelcase` - had several interoperability problems with the snake_case identifiers used in ES and sm-api requests. I looked at an alternative, `@typescript-eslint/naming-convention`, but it required a significant amount of manual configuration.
* `@typescript-eslint/explicit-module-boundary-types` - new "requiring-type-checking" rule. It would be nice to have, but would require a significant amount more types to be manually annotated.
* `@typescript-eslint/restrict-template-expressions` - new "requiring-type-checking" rule. Primarily complained about using `any`-typed variables in log statements. Wouldn't add much value to fix the code, as it would just mean a bunch of `x as string` or `String(x)`s throughout the logging statements.
* `@typescript-eslint/unbound-method` - Caused lots of unresolvable false positives with lodash functions
* `no-use-before-define` - Complained about imports that were only used in type annotations. I removed this as TypeScript does this check during compilation anyway.
* `no-useless-constructor` - had false positives with TypeScript classes

I disabled the following rules as they ban TypeScript's various escape hatches (`nullableVar!.x`, `as any`, etc.). Really not sure why they're on by default - it's obvious from syntax when code is circumventing type checking, and it's only done as a last resort, so these rules would just result in lots of `// eslint-disable-next-line`s through the codebase...
* `@typescript-eslint/no-explicit-any`
* `@typescript-eslint/no-unsafe-member-access`
* `@typescript-eslint/no-unsafe-return`
* `@typescript-eslint/no-unsafe-assignment`
* `@typescript-eslint/no-unsafe-call`
* `@typescript-eslint/ban-ts-comment`
* `@typescript-eslint/no-non-null-assertion`

You can filter by commit depending on how much you think warrants review. The first 4 commits are setting up the infrastructure, then the next 3 are automated, semi-automated and manual fixes respectively.

I didn't set up any file watching, etc. commands as ts-node-dev doesn't have tidy hooks for it in the same way that vue-cli / create-react-app have. I think it's probably best left to individual developers' IDEs anyway.


metaspace/graphql/server.js
```diff
@@ -43,190 +42,191 @@ const configureSession = (app) => {
...
     introspection: true,
-  });
-  apollo.applyMiddleware({ app });
+  })
+  apollo.applyMiddleware({ app })
 
-  configureSentryErrorHandler(app);
+  configureSentryErrorHandler(app)
 
-  app.use(function (err, req, res, next) {
```
richardgoater, line 191: I think this is a classic issue with `express`, I'm fairly sure the function needs all four arguments to work as an error handler. It's a terrible API and there ought to be a dedicated class for it like `express.Router`.

richardgoater (approved):
The formatted code looks really great, thanks for finishing this off 👍 

I think there's one issue with the express error handler that needs an exception to the rules.

metaspace/graphql/server.js
LachlanStuart: Nice catch! Thanks, I've pushed a fix.

PR #749, opened 2021-02-06
fix: Make sure engine works with minio object storage

intsco:
Refactor code and update configs to support local minio object storage in engine.

There is one thing I couldn't fix since I didn't have much time. Webapp formats dataset minio path and sends it to graphql
in the form `s3a://storage:9000/22735870-552d-4a56-9f03-75f6192b033a` while it should be
`s3a://<BUCKET_NAME>/22735870-552d-4a56-9f03-75f6192b033a`.
I guess it's caused by the webapp's `util.getS3Bucket` function that picks bucket name from the host name.

For the moldb uploads, the same fix is needed. Webapp should send moldb file paths without the endpoint component. `s3a://<BUCKET_NAME>/databases/ae679c95-8387-4463-a418-524ba1ca28ff/hmdb_small.csv` instead of `s3://storage:9000/<BUCKET_NAME>/databases/ae679c95-8387-4463-a418-524ba1ca28ff/hmdb_small.csv`


richardgoater:
@intsco thanks very much for this! Do you like the design of the util functions? Would you prefer to do this another way?

intsco:
@richardgoater No, I don't have any preferences for the webapp code. What I meant is that `getS3Bucket` should be fixed/updated to support extraction of bucket name from minio object paths.

richardgoater:
@intsco I meant your preference on the Python code :) I can work on the bucket "parsing", though I feel like the name of the bucket should be in the response separately.

intsco:
@richardgoater Based on the quick discussion during the sprint meeting, I have added support for moldb imports using minio object paths. It mostly works, I tested it locally but a small webapp fix is needed here too (see the updated PR description).

PR #753, opened 2021-02-09
Periodically force webapp to reload the page

LachlanStuart:
I tested it locally with `daysSincePageLoad > 0.0001` to reduce the wait time. The refreshing is a little bit disruptive - it shows a blank screen for ~1s while reloading everything. Since it's always during page transitions (unless they hit 7 days), it's unlikely that it will cause state loss.

A small issue is that the browser history doesn't behave correctly if the reload is caused by the user hitting back - it creates a new forward navigation instead of a backward navigation. I don't think it will happen often enough to impact anyone though.

I chose an interval of 1.75 days because refreshing every single day seems like it would become noticeable and annoying. Also, if it were an integer number of days, habitual users would get hit with a refresh at slightly later times each day, which could interrupt their flow. With 1.75, it's most likely that they'll get the refresh on their first use of the day.


richardgoater (approved):
Interesting one, see how it goes with the back button 👍

PR #754, opened 2021-02-11
RAM-based dataset segmentation

LachlanStuart:
The problem of dataset segmentation is that peaks in the `imzML` files are sorted by `spectrum_idx, mz`, and the annotation step needs them in `mz, spectrum_idx` order for efficient processing. Additionally, the spectrum is split into 128MB segments for to allow for easily loading small ranges of m/z values. 

Due to memory constraints, the old code would read chunks of the input file and distribute it across a set of temporary files, then re-read each temporary file, sort the peaks and save a segment to COS. Unfortunately disk access was so slow (~50MB/s) that large datasets would timeout.

This PR changes it to do the segmentation completely in RAM with no temporary files. This requires much more RAM(2-4x the size of the `.ibd` file), but it is overall much faster because it doesn't have the I/O bottleneck.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -383,8 +383,11 @@ workflows:
             - generate-graphql-schema
           filters:
             branches:
-              # Replace this with the branch to run sci-test against
-              only: feat/ibm-cloud
+              only:
+                - master
+                # Add additional branches to run sci-test against here. Clean up old branches if you notice any.
+                - feat/lithops-ds-sort-in-ram
+
 
 ## Disabled until we have time to redo these tests
 #  daily:
--- a/metaspace/engine/conf/scitest_config.json
+++ b/metaspace/engine/conf/scitest_config.json
@@ -55,7 +55,7 @@
       "include_modules": ["engine/sm"],
       "data_cleaner": true,
       "data_limit": false,
-      "workers": 2
+      "workers": 4
     },
     "localhost": {
     },
--- a/metaspace/engine/sm/engine/annotation_lithops/load_ds.py
+++ b/metaspace/engine/sm/engine/annotation_lithops/load_ds.py
@@ -109,9 +109,9 @@ def _sort_spectra(imzml_reader, mzs, ints, sp_lens):
     # sp_idxs in a compacted form with sp_lens until the last minute.
     sp_id_to_idx = get_pixel_indices(imzml_reader.coordinates)
     sp_idxs = np.empty(len(ints), np.uint32)
-    sp_lens = np.cumsum(sp_lens)
-    for sp_id, (start, end) in enumerate(zip(sp_lens[:-1], sp_lens[1:])):
-        sp_idxs[start:end] = sp_id_to_idx[sp_id]
+    sp_lens = np.insert(np.cumsum(sp_lens), 0, 0)
+    for sp_idx, start, end in zip(sp_id_to_idx, sp_lens[:-1], sp_lens[1:]):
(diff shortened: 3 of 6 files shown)


metaspace/engine/conf/scitest_config.json
```diff
@@ -55,7 +55,7 @@
       "include_modules": ["engine/sm"],
       "data_cleaner": true,
       "data_limit": false,
-      "workers": 2
+      "workers": 4
```
LachlanStuart, line 58: CircleCI seems to share CPU resources. Even though our selected instance says "2 CPUs", using 4 workers *sometimes* (but not always) causes it to take half as much time.

.circleci/config.yml
```diff
@@ -383,8 +383,11 @@ workflows:
             - generate-graphql-schema
           filters:
             branches:
-              # Replace this with the branch to run sci-test against
-              only: feat/ibm-cloud
+              only:
+                - master
```
LachlanStuart, line 387: I had accidentally merged `only: feat/ibm-cloud`, which disabled sci-test for master. This re-enables it and hopefully using a list here prevents that mistake from happening again.

metaspace/engine/tests/test_es_export.py
```diff
@@ -102,7 +102,7 @@ def test_index_ds_works(sm_config, test_db, es_dsl_search, sm_index, ds_config,
             ds_id=ds_id, moldb=moldb, isocalc=isocalc_mock,
         )
 
-    wait_for_es(sec=1)
+    wait_for_es(sec=1.5)
```
LachlanStuart, line 105: This is an unrelated fix. I found that CircleCI spontaneously started having intermittent failures in this test. It seems like ElasticSearch might not have enough time to finish indexing the document. e.g.
* https://app.circleci.com/pipelines/github/metaspace2020/metaspace/1183/workflows/a1e13dbe-df28-4021-9ee2-73948873b3db/jobs/13823
* https://app.circleci.com/pipelines/github/metaspace2020/metaspace/1162/workflows/f0947cb7-e718-49e3-b545-b993c7b9fc92/jobs/13696

Not sure what has changed, but hopefully increasing the delay makes it stable again.

metaspace/engine/sm/engine/annotation_lithops/load_ds.py
```diff
@@ -70,234 +66,113 @@ def get_spectra(
...
+    n_spectra = len(imzml_reader.coordinates)
+    mz_arrays = [np.array([], dtype=imzml_reader.mzPrecision)] * n_spectra
+    int_arrays = [np.array([], dtype=np.float32)] * n_spectra
+    sp_lens = np.empty(n_spectra, np.int64)
+
+    def read_spectrum_chunk(start_end):
+        spectra = get_spectra(storage, imzml_reader, ibd_cobject, list(range(*start_end)))
+        for sp_i, mzs, ints in spectra:
+            mz_arrays[sp_i] = mzs
+            int_arrays[sp_i] = ints.astype(np.float32)
```
sergii-mamedov, line 83: Why it called `int`, if type is float?

sergii-mamedov (approved):
LGTM

metaspace/engine/sm/engine/annotation_lithops/load_ds.py
LachlanStuart: As an abbreviation for "intensity". `ints` is also used in many places so that it doesn't shadow the `int` built-in type.

PR #756, opened 2021-02-12
Handle Lithops stalls

LachlanStuart:
Lithops stalls are hard to handle:
* It has several timeout mechanisms, but they seem to be under active development and have had a number of bugs and regressions
* Edge cases are hard to test - usually OOMs result in failures, but occasionally they're killed before they can write a status file to COS
* Lithops runs background processes and threads to monitor the jobs, so even if the current thread was interrupted by a `SIGINT` or injected exception, the background threads would continue running, making API calls and costing money

This PR runs Lithops in a separate thread, and if the thread takes an exceptionally long time (41 minutes, configurable) then it raises an exception, re-queues the message, and exits a process. The first retry will be with Lithops again, but if it fails a second time, the message will be transferred to the Annotate queue for Spark to handle.

Supervisor handles restarting the process as long as the exit code is non-zero. To actually achieve this I call `sys.exit(1)` from the interrupt handler, ensure that the main thread stays alive so that it can catch the `SystemExit` exception by `join`ing the daemon threads, and triggering a `SIGINT` from inside the daemon. This could probably be a lot cleaner if we used a framework for the daemons (e.g. Celery).

Additionally a bunch of small changes were rolled into this PR to resolve issues I found during testing:
* Fixed the Ansible step & Supervisor config template for running multiple sm-lithops-daemon processes
* Added a DB-based mutex around VM execution so that the multiple processes didn't try to use the VM simultaneously
* Limit the number of parallel workers on VM and localhost executors to be 128GB/`runtime_memory` and 32GB/`runtime_memory` respectively
* Add extra logging for deserialization errors. I found that the centroids cache in COS had some invalid pickle files because the `FDR` class had moved. The extra logging should help find exactly which files are affected if this happens again.
* increase the production `execution_timeout` from 40 to 60 minutes, as one of the DSs timed out at 40 mins
* Changed `TimeoutError` handling so that it retries with double memory, similar to `MemoryError`, instead of jumping straight to VMs. The reason is that `process_centr_segment` usually has too many jobs to run on the VM, but occasionally it will timeout instead of OOM when running on 2GB. On 4GB it seems to work without issues.

### DIFF
--- a/metaspace/engine/sm/engine/queue.py
+++ b/metaspace/engine/sm/engine/queue.py
@@ -471,10 +471,10 @@ class QueueConsumer(Thread):
         self.logger.info(' [*] Waiting for messages...')
 
         while True:
-            self.get_message()
-            self._failed_attempts = 0
             if self.stopped():
                 raise StopThread()
+            self.get_message()
+            self._failed_attempts = 0
             sleep(self._poll_interval)
 
     def stop(self):
--- a/metaspace/engine/tests/test_queue.py
+++ b/metaspace/engine/tests/test_queue.py
@@ -25,16 +25,17 @@ def delete_queue(sm_config):
     queue_pub.delete_queue()
 
 
-def run_queue_consumer_thread(config, callback, output_q, wait=1):
+def run_queue_consumer_thread(config, callback, output_q, wait=0.1):
     queue_consumer = QueueConsumer(
         config,
         QDESC,
         callback,
         lambda *args: output_q.put('on_success'),
         lambda *args: output_q.put('on_failure'),
-        poll_interval=0.1,
+        poll_interval=wait,
     )
     queue_consumer.start()
+    time.sleep(wait)
     queue_consumer.stop()
     queue_consumer.join()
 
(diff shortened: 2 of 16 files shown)


metaspace/engine/sm/engine/queue.py
```diff
@@ -471,10 +471,10 @@ def _poll(self):
         self.logger.info(' [*] Waiting for messages...')
 
         while True:
-            self.get_message()
-            self._failed_attempts = 0
             if self.stopped():
                 raise StopThread()
```
LachlanStuart, line 475: I moved this check to happen before `get_message` / after `sleep` because this was starting to process the retry message before the queues realized `stop()` had been called. This change broke tests that depended on the `QueueConsumer` always processing at least 1 message before stopping, so I added several `time.sleep(0.1)`s into the tests to wait for this check to pass before trying to stop the daemon.

metaspace/engine/tests/test_queue.py
```diff
@@ -56,11 +57,11 @@ def test_queue_msg_published_consumed_on_success_called(sm_config, delete_queue)
...
     output_q = Queue()
     run_queue_consumer_thread(
-        config, callback=lambda *args: output_q.put('callback'), output_q=output_q, wait=1
+        config, callback=lambda *args: output_q.put('callback'), output_q=output_q
     )
 
-    assert output_q.get() == 'callback'
-    assert output_q.get() == 'on_success'
+    assert output_q.get(block=False) == 'callback'
+    assert output_q.get(block=False) == 'on_success'
```
LachlanStuart, line 64: I added `block=False` here because it was painful to debug this - if the queue was empty, the calls without `block=False` just waited forever.

sergii-mamedov (approved):
This is pretty low-level code, and without actually working on it, I can only evaluate it at a high level. I don't see any obvious problems right now.

PR #761, opened 2021-02-17
Fix SVG in Safari

richardgoater:


metaspace/webapp/src/components/PrimaryIcon.vue
```diff
@@ -15,13 +15,16 @@ export default defineComponent({
   },
 })
 </script>
-<style scoped>
+<style scoped lang="postcss">
```
LachlanStuart, line 18: Why postcss? Can this be scss?

metaspace/webapp/src/components/PrimaryIcon.vue
```diff
@@ -15,13 +15,16 @@ export default defineComponent({
...
+i >>> svg {
+  width: 100%;
+}
 i >>> .primary,
 i.inverse >>> .secondary {
-  @apply text-blue-400;
+  @apply fill-current text-blue-400;
```
LachlanStuart, line 24: Are the `fill-current` changes a necessary part of the fix? I'm not asking for them to be reverted, just trying to understand what happened with Safari.

LachlanStuart (approved):
Fine to merge, but I'd prefer it if the `postcss` blocks were `scss`, just to minimize tech stack dilution.

metaspace/webapp/src/components/PrimaryIcon.vue
LachlanStuart: Nevermind, I just saw the commit message about this. Makes sense now.

metaspace/webapp/src/components/PrimaryIcon.vue
richardgoater: this was a blunt method of fixing linting and syntax highlighting, I can remove it.

PR #768, opened 2021-02-24
Fix graphql emails

LachlanStuart:
`AWS.SES` had the same issue as you encountered with `AWS.S3` not liking the `config.aws` object. I found the cause:

```
> console.log('config.aws:', Object.getOwnPropertyNames(config.aws))
config.aws: [
  'region',            'credentials',
  'util',              'get',
  'has',               'watch',
  'setModuleDefaults', 'makeHidden',
  'makeImmutable',     'getConfigSources',
  '_loadFileConfigs',  '_parseFile',
  '_attachProtoDeep',  '_cloneDeep',
  '_equalsDeep',       '_diffDeep',
  '_extendDeep',       '_stripYamlComments',
  '_stripComments',    '_isObject',
  '_initParam',        '_getCmdLineArg'
]

> console.log('config.aws copy:', Object.getOwnPropertyNames(_.cloneDeep(config.aws)))
config.aws copy: [ 'region', 'credentials' ]
```

The `config` package is doing some funky stuff with non-enumerable properties. When there's a `credentials.get` function, `aws-sdk` treats them as callback-style async credentials, so it calls `.get` with a callback function, causing `config`'s `get` to freak out because it's being called with a non-string.

For consistency I moved the SES construction into the same file and renamed it to `awsClient.ts`. I also updated the existing code to use `_.cloneDeep` as it's a simpler way to strip the unwanted non-enumerable properties.


richardgoater (approved):
sorry about this mess, and good find on enumerable properties.

PR #773, opened 2021-03-01
Feat: S3 image storage

intsco:
This PR implements S3 based image storage. For local development, S3 compatible minio server is used. This is a POC implementation and its design is subject to discussion. One of the aspects that I focused on was simplifying client code at expense of slight increase in internal `image_storage` module complexity.

The `image_storage` module will replace `image_store` but the later one is temporarily left for compatibility.

To save disk space, only image ids (not full urls) are stored in the database. Full urls are exported to ES so that graphql can use them without any knowledge about the image storage.

Not related changes:
* `boto3-stubs` and `mypy_boto3_builder` to improve coding experience with `boto3`. May slow down PyCharm so additional testing is needed.
* `pylint` and `black` updates which didn't lead to massive amount of code changes.

Not included features:
* Any modules/functions that use lithops.
* Handling of optical images.
* Serving iso images from graphql.


LachlanStuart (commented):
The overall architecture seems good. My only architectural concern is how/if test environments are going to share data. Do you think it's fine if we just don't copy the ion images when we clone prod, letting the copied datasets not have images?

Regarding code style though, I know it's not my codebase, so feel free to overrule me here, but I'm not a fan of the new module patterns:
* Doing `from sm.engine import image_storage` departs from the most Python code style I've seen, and makes usage more verbose, e.g. needing  `image_storage.ImageType.THUMB` instead of `ImageType.THUMB`
* The way `ImageStorage` is a class, but its methods are copied up to the module level. I don't see the advantage of this, but it seems to have more boilerplate code than doing a singleton class, or avoiding classes and having the private state at the module level. Additionally, it causes problems if it's not used with the above `import image_storage` pattern, because the `get_image`, etc. module variables may not have been initialized.

We have a lot of different patterns for dependency injection in this codebase:
* `bootstrap_and_run` / `with GlobalInit` / `with ConnectionPool`
* `SMConfig` used as a singleton
* `DB()` instances created inline whenever needed
* `DatasetManager` instances explicitly passed to all classes that need it
* `transaction_context` and `capture_logs` using thread/context-local storage to pass dependencies into wrapped code

I feel that if the above patterns don't cover all our needed use-cases, we should consider adding a Dependency Injection framework (e.g. [lagom](https://lagom-di.readthedocs.io/en/latest/)) that theoretically can cover most of the above patterns' use-cases with a single pattern, instead of adding new self-made patterns.

sergii-mamedov (approved):
LGTM

PR #778, opened 2021-03-08
Image storage: serve iso images, lithops support

intsco:
* Export iso image URLs to ES.
* Make image storage work with lithops.
* Remove dataset indexing from `ServerAnnotationJob`.
* Configure policy for image storage bucket on startup.
* Fix sci-test.
* Remove deprecated `bootstrap_and_run`, refactor dependent scripts.
* Update graphql `Annotation` resolver.
* Handle `pylint==2.7.0` complaints.

### DIFF
--- a/metaspace/engine/sm/engine/image_storage.py
+++ b/metaspace/engine/sm/engine/image_storage.py
@@ -13,7 +13,7 @@ import PIL.Image
 try:
     from mypy_boto3_s3.service_resource import S3ServiceResource
     from mypy_boto3_s3.client import S3Client
-except:  # pylint: disable=bare-except
+except ImportError:
     S3ServiceResource = object
     S3Client = object
 
--- a/metaspace/engine/sm/engine/image_storage.py
+++ b/metaspace/engine/sm/engine/image_storage.py
@@ -17,6 +17,7 @@ except ImportError:
     S3ServiceResource = object
     S3Client = object
 
+from sm.engine.config import SMConfig
 from sm.engine.storage import get_s3_resource, create_bucket
 
 logger = logging.getLogger('engine')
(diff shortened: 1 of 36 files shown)


metaspace/engine/sm/engine/image_storage.py
```diff
@@ -21,14 +30,32 @@ class ImageType(str, Enum):
...
+                {
+                    'Effect': 'Allow',
+                    'Principal': {'AWS': ['*']},
+                    'Action': ['s3:GetObject'],
+                    'Resource': [f'arn:aws:s3:::{config["bucket"]}/*'],
+                }
+            ],
+        }
+        self.s3_client.put_bucket_policy(Bucket=config['bucket'], Policy=json.dumps(bucket_policy))
+        self.bucket = self.s3.Bucket(config['bucket'])
```
LachlanStuart, line 58: This is only needed once, right? 

Could this be called from the module-level `init` instead of `ImageStorage.__init__` so that it only happens when the application starts? I'm a bit wary of this being run multiple times per annotation job, in case there are any weird race conditions, tiny blips of downtime, or errors updating the policy while we're not watching.

metaspace/graphql/src/modules/annotation/controller/Annotation.ts
```diff
@@ -125,14 +125,12 @@ const Annotation: FieldResolversFor<Annotation, ESAnnotation | ESAnnotationWithC
...
-    const { iso_image_ids, centroid_mzs, total_iso_ints, min_iso_ints, max_iso_ints } = hit._source
+    const { iso_image_urls, centroid_mzs, total_iso_ints, min_iso_ints, max_iso_ints } = hit._source
     return centroid_mzs
       .map(function(mz, i) {
         return {
           mz: parseFloat(mz as any),
-          url: iso_image_ids[i] !== null
-            ? `/${hit._source.ds_ion_img_storage}${config.img_upload.categories.iso_image.path}${iso_image_ids[i]}`
-            : null,
+          url: iso_image_urls[i],
```
LachlanStuart, line 133: Just thinking about the transition period while data is being migrated - it will take several days, or maybe even a full week to reindex ES. Were you planning on doing a live update-in-place, or using a separate checkout of the new code to populate the inactive index, then do a swap & code update with a short downtime?

If you're planning to do an update-in-place, it would be best if graphql checks for both fields and falls back to the old logic if `iso_image_urls` is undefined.

LachlanStuart (approved):
Thanks for cleaning up the lint issues. Generally looks good. 

My comment on `ImageStorage._configure_bucket` is non-critical - skip it if you don't have time.

Action is only needed on the migration comment if you don't already have a plan for how to deal with the transition period.

metaspace/graphql/src/modules/annotation/controller/Annotation.ts
intsco: Agree, I will probably do update-in-place, have already added needed changes to the next PR.

sergii-mamedov (approved):
LGTM

PR #784, opened 2021-03-10
Lithops hotfixes

### LINKED ISSUE #770: Lithops: Periodically restart daemons
Though there's no indication of a memory leak, this daemon can grow to use 1-2GB of memory for some datasets. Python often doesn't release memory when it's no longer used, so if these daemons run continually, they're likely to stabilize on their worst-case memory consumption. Periodically restarting them should reduce the average memory consumpiton.

Supervisorctl doesn't seem to be a good fit for this - it has only 1 timeout option (default: 10s) for stopping processes before it kills them, and it would be best if we allowed busy daemons to finish their current job before restarting. 

Instead, I think it would be best to add a command line option to the daemon that sends a `SIGINT` signal to itself after 23 hours + a random 0-60 minutes. This would allow the process to gracefully stop, then supervisor will restart it.

### LINKED ISSUE #771: Lithops: Raise open file limit on servers
Adding this as an issue so I don't forget - there's a temporary fix in place (I ran `ulimit -Sn 262144` on prod) but this needs to be made part of the Ansible provisioning step as well. My internet's shaky ATM so I can't easily test this.

Ansible has a [built-in module](https://stackoverflow.com/a/47101272) for setting this.

### LINKED ISSUE #774: Lithops: Support datasets >50GB
The Boto client defaults to 5MB parts when doing a multi-part upload to transfer a dataset to COS. As there's a 10000 part limit to multi-part uploads, this effectively means that datasets over 50000MB will fail.

As a temporary workaround for while we still have uploads set to go to S3, increase the part size to 20MB.

LachlanStuart:
This fixes several mostly Lithops-related issues:

#### `IOError: Too many open files` and `psycopg2.OperationalError: SSL connection has been closed unexpectedly` exceptions
Fixes #771

This raises the open file limit for the process on startup. I initially tried setting this on the server with `ulimit -n 127128`, but it appears that only sets it for the current terminal session. I decided to move this into the process as it's easier to manage code than infrastructure, and it affects only the Lithops daemon.

Additionally, I added code to `QueueConsumer` to exit the process if there is an error in the failure handler (i.e. there was an exception in the main task AND an exception handling the previous exception). `IOError: Too many open files` is actually the 2nd unrecoverable error we've seen. The first was `psycopg2.OperationalError: SSL connection has been closed unexpectedly`. It happens every time Postgres restarts (usually due to Ubuntu automatic updates), and every psycopg2 DB connection just stays in a broken state, raising exceptions on every call.

#### High long-term memory usage
Fixes #770

When large datasets are processed, the lithops-daemon process seems to grow to ~2GB and not shrink again after finishing. I added an optional `--exit-after` command-line parameter that stops the daemon after the specified number of seconds. Supervisorctl then restarts the daemon.

I also suspect that Lithops doesn't correctly clean up the background threads when there have been execution errors. Restarting periodically should prevent these threads from accumulating and becoming a problem.

To test this I ran it in Docker with `--exit-after 10` and verified that it shut down after 10s.

#### Increase the IBM COS multipart upload part size
Fixes #774 

Multipart uploads can have a maximum of 10000 parts. `upload_fileobj` doesn't know the file size, so it just guesses and uses 8MB parts, meaning the maximum file size that could be transferred from S3 to IBM COS is ~80GB. This change increases it to 20MB per part, so ~200GB maximum.

Additionally I increased some of the parameters for upload - parallelism from 10 to 20, and the read/write chunksize from 256k to 1MB. Hopefully this increases transfer speed.

To test this I uploaded a 300MB dataset on Staging and watched the logs to make sure there were no errors while it was transferring.

### DIFF
--- a/ansible/roles/sm_lithops_daemon/templates/sm-lithops-daemon.supervisor.j2
+++ b/ansible/roles/sm_lithops_daemon/templates/sm-lithops-daemon.supervisor.j2
@@ -4,7 +4,7 @@
 process_name = {{ sm_lithops_daemon_app_name }}-%(process_num)s
 environment = PATH="{{ miniconda_prefix }}/envs/{{ miniconda_env.name }}/bin"
 directory = {{ sm_home }}
-command = {{ miniconda_prefix }}/envs/{{ miniconda_env.name }}/bin/python scripts/run_sm_daemon.py --name=lithops
+command = {{ miniconda_prefix }}/envs/{{ miniconda_env.name }}/bin/python scripts/run_sm_daemon.py --name=lithops --exit-after 86400
 redirect_stderr = true
 stdout_logfile = {{ sm_home }}/logs/{{ sm_lithops_daemon_app_name }}-%(process_num)s.log
 numprocs = {{ sm_lithops_daemon_nprocs }}
--- a/metaspace/engine/scripts/run_sm_daemon.py
+++ b/metaspace/engine/scripts/run_sm_daemon.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 import argparse
 import logging
+import resource
 import signal
 import sys
 from functools import partial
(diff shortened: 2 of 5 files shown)


sergii-mamedov (approved):
LGTM.
Very good description of the problem and the solution method.

PR #786, opened 2021-03-11
Image storage: optical images

intsco:
This PR enables optical images to use new S3 image storage.

**Notes:**
* Add url database fields for optical images and thumbnails. Storing just ids in the database requires indexing urls in ES which makes things only more complicated.
* Implement fallback logic for graphql url fields to use old values if new ones are not available yet. It's expected to happen during data migration that might take several days to complete.

**Know issues:**
* Raw optical images are still uploaded to the old image web server. To fully migrate that part, webapp changes on the optical image alignment page are required.
* After uploading an optical image, a thumbnail is generated and saved but not used by the webapp. Probably some frontend changes are required.


LachlanStuart:
> After uploading an optical image, a thumbnail is generated and saved but not used by the webapp.

It is used, but it appears behind the ion thumbnail. It's only really visible on sparse datasets, and very hard to see:

image

The greenish off-sample area comes from [this image](https://metaspace2020.eu/fs/optical_images/fa0fb9255ce3a48e6a437021858dfaef) (`fa0fb9255ce3a48e6a437021858dfaef` comes from the `thumbnail` column in postgres / the `Dataset.thumbnailOpticalImageUrl` graphql field)

LachlanStuart (approved):

sergii-mamedov (approved):
LGTM.

PR #788, opened 2021-03-12
Hotfix: Use correct exit code when exiting daemons due to --exit-after flag

LachlanStuart:
I found that lithops-daemon wasn't restarting because it exited with a `0` exit code. It seems that when you call `sys.exit(1)` on a non-main thread, the exit code is ignored and the main thread's exit code is used instead.

This changes the timer behavior to raise a `SIGINT` signal, which is handled on the main thread, causing it to exit with the correct code.


sergii-mamedov (approved):
LGTM

Issue #783, opened 2021-03-10
Add pagination & sorting to Datasets page


theodev:
* In the datasets page, keep the limit of 100 but add pagination
* Add interface for sorting datasets by Last reprocessed date (as currently)/Upload date/Number of annotations for FDR 10%/User name/Dataset name (will require some work on GraphQL and ES)

Issue #777, opened 2021-03-08
LS: First task idea for Lucas: Add a "My datasets"/"Group datasets"/"All datasets" smart filter box to the datasets list (similar to the projects page)


theodev:
As discussed during the sprint

Similar to how "My projects" in projects

Ideally so that it will be remembered for each user

PR #789, opened 2021-03-12
Feat/dataset pg filter sorting

lmacielvieira:
### Description
This PR enables sorting, pagination and filtering by dataset owner. https://github.com/metaspace2020/metaspace/issues/783 https://github.com/metaspace2020/metaspace/issues/777

### New Features

* [x] The filter by dataset owner allows the user to filter the dataset by  "My datasets", "Group datasets" and "All datasets"
* [x] The Pagination allows the user with more than 100 datasets to paginate in chunks of 100 datasets
* [x] The Sorting allows the user to sort the datasets by "Last reprocessed date", "Upload date", "Number of annotations for FDR 10%", "User name" and "Dataset name" in asc or desc order

##### webapp

- [x] Treated conflict in submitter filter
- [x] Added My Group filter for dataset
- [x] Added simple filter type 
- [x] To Hide Filter when not logged
- [x] Added My dataset filter for dataset
- [x] Added All datasets filter for dataset
- [x] Added currentUser on store
- [x] New composed generic filter 
    - [x] Value and id - datasetOwner and datasetOwnerId
    - [x] Id is set according to datasetOwner type
- [x] Treat conflict in group filter

##### webapp

- [x] Added SortDropdown Component
- [x] Passing orderBy and sortingOrder variables to allDatasets  query 

##### graphql

- [x] Added "Sort by last reprocessed date"
- [x] Sort by "Sort by  uploaded date"
- [x] Added "Sort By user name"
- [x] Added "Sort By Number of annotations for FDR 10%"

##### webapp
- [x] Added pagination element at Dataset page
- [x] Passing offset queryVariable to allDatasets query
Screenshot 2021-03-15 at 07 11 34


metaspace/webapp/src/components/SortDropdown/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+          label: 'Last Reprocessed Date',
+        },
+        {
+          value: 'ORDER_BY_UP_DATE',
+          label: 'Uploaded Date',
+        },
+        {
+          value: 'ORDER_BY_DS_SUBMITTER_NAME',
+          label: 'User name',
+        },
```
LachlanStuart, line 47: These are a bit inconsistent in capitalization & usage of terminology. Additionally, I think the default should be the first item, and similar datatypes (date/string/number) should be grouped together to make reading it easier. I'd suggest changing them to the following:

* Last updated
* Upload date
* Dataset name
* Submitter name
* Annotation count

metaspace/webapp/src/modules/Datasets/list/DatasetTable.vue
```diff
@@ -105,33 +130,62 @@ export default Vue.extend({
...
         query: this.$store.getters.ftsQuery,
         inpFdrLvls: [10],
         checkLvl: 10,
-        offset: Math.max(0, (this.$store.getters.settings.datasets.page - 1) * 100),
+        offset: (this.currentPage - 1) * this.recordsPerPage, // Math.max(0, (this.$store.getters.settings.datasets.page - 1) * 100),
+        orderBy: this.orderBy,
+        sortingOrder: this.sortingOrder,
       }
     },
     nonEmpty() {
```
LachlanStuart, line 191: I found an awkward edge case. When there are datasets in the "queued" or "processing" status, they're always put at the top of the list. This is intended behavior for the default sort, but it looks very weird with e.g. sorting by upload date. e.g.

image

A few lines below this there's a `datasets()` computed method where this status-based sorting override is applied. I would suggest changing it to skip that override if the sort isn't `ORDER_BY_DATE` and `DESCENDING`.

metaspace/webapp/src/components/SortDropdown/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+        <div class="el-input-group__append sort-dp-btn">
+          <el-tooltip
+            content="Sorting order"
+            placement="right"
+          >
+            <el-button
+              class={`${!state.value ? 'cursor-not-allowed' : ''}`}
+              icon={state.orderBy === SortingOrder.Unsorted ? 'el-icon-sort' : (state.orderBy === SortingOrder.Desc
+                ? 'el-icon-sort-down' : 'el-icon-sort-up')}
+              onClick={state.value ? handleSort : null}
```
LachlanStuart, line 95: `el-button` gives console errors when this is `null`:
image

I think you may need to move this logic into `handleSort`

metaspace/webapp/src/components/SortDropdown/index.tsx
```diff
@@ -0,0 +1,102 @@
+import { defineComponent, reactive } from '@vue/composition-api'
```
LachlanStuart, line 1: Please use filenames that reflect the contents of the file. My workflow depends on me being able to jump between files by name, but that doesn't work well if most files are named "index" and "style". `SortDropdown.tsx` and `SortDropdown.css` would be better here.

LachlanStuart (approved):
Great work on getting this done so fast! Sorry it took a while for me to review this.

I've approved with comments. This means I've noticed some stuff where I have feedback that may need to be addressed, but it probably doesn't require a re-review after the changes, so you can merge after responding to the comments or making the suggested changes/fixes. 

I try to make it clear which feedback items I consider necessary to address before merging (either by you making the changes/fixes, or disagreeing and us having a conversation about it), and which are optional suggestions, etc. (which I'd appreciate a response to, but aren't really blockers). 

In this case, changing the filename for `SortDropdown` and fixing the console error are necessary. The rest are suggestions.

PR #798, opened 2021-03-22
Add image storage data migration script

intsco:


metaspace/engine/migrations/r1_8_20210322_image_storage.py
```diff
@@ -0,0 +1,255 @@
...
+    transfer_config = s3transfer.TransferConfig(
+        use_threads=True, max_concurrency=s3_client.meta.config.max_pool_connections
+    )
+    return s3transfer.create_transfer_manager(s3_client, config=transfer_config)
+
+
+def dataset_is_migrated(ds_id):
+    image_type = image_storage.ISO
+    resp = s3t_client.list_objects(Bucket=bucket_name, Prefix=f'{image_type}/{ds_id}', MaxKeys=1)
+    return 'Contents' in resp
```
LachlanStuart, line 55: This function appears to be unused - good. I think the approach with the `.txt` files is more reliable, because `dataset_is_migrated` would return `True` if a dataset failed after several images had successfully transferred. 

Just leaving this comment to make sure this function doesn't get re-added to the control flow.

LachlanStuart (approved):

LachlanStuart:
Looks good. I'm surprised it doesn't run jobs in parallel, but I guess as discussed on Slack it's probably not worth the extra time to optimize.

metaspace/engine/migrations/r1_8_20210322_image_storage.py
```diff
@@ -0,0 +1,260 @@
...
+):
+    s3t = create_s3t()
+    futures = []
+    for image_id in image_ids:
+        if image_id:
+            image_path = Path(image_base_path) / old_image_type / image_id[:3] / image_id[3:]
+            image_key = f'{image_type}/{ds_id}/{image_id}'
+            fut = s3t.upload(fileobj=str(image_path), bucket=bucket_name, key=image_key)
+            futures.append(fut)
+    s3t.shutdown()
```
intsco, line 63: This code does use multithreading, it's just hidden inside the `s3t` object. `s3t.upload` returns futures and `s3t.shutdown` waits for all tasks to finish.

metaspace/engine/migrations/r1_8_20210322_image_storage.py
```diff
@@ -0,0 +1,260 @@
...
+'''
+
+
+def _es_docs_migrated(es, ds_id):
+    query_body = {'query': {'match': {'ds_id': ds_id}}}
+    resp = es.search(
+        index='sm', doc_type='annotation', body=query_body, _source=['iso_image_urls'], size=1,
+    )
+    hits = resp['hits']['hits']
+    return hits and 'iso_image_urls' in hits[0]['_source']
```
intsco, line 86: @LachlanStuart I removed `dataset_is_migrated` but in addition to relying on txt files I have to check for already migrated images this way to handle datasets that were uploaded in between two migration starts.

metaspace/engine/migrations/r1_8_20210322_image_storage.py
LachlanStuart: Although the individual steps of the pipeline have parallelism (here and in the ElasticSearch library), the pipeline itself doesn't run in parallel (i.e. multiple datasets aren't run in parallel). This means it can only ever bottleneck on one resource at a time - while it's uploading images ElasticSearch will be idle, and while it's reindexing, S3 will be idle.

With Lithops & data analysis I generally run everything that takes a non-trivial amount of time through a `ThreadPoolExecutor` or `ProcessPoolExecutor` when working with multiple datasets at the same time, as it usually gives a 2x-4x speed boost

metaspace/engine/migrations/r1_8_20210322_image_storage.py
LachlanStuart: Won't those datasets already be handled by the text files, and the fact that new uploads will use the new code so won't need migration? Those text files aren't being deleted between runs, right?

The chances of a dataset being only partially reindexed in ES is much lower than partially uploaded to S3. I'm mainly concerned that this logic also makes it harder to fix any datasets that end up in `FAILED_DATASETS.txt` because it causes them to be skipped even when `force=True`.

intsco: >Won't those datasets already be handled by the text files, and the fact that new uploads will use the new code so won't need migration? Those text files aren't being deleted between runs, right?

New uploads submitted after the migration started don't get saved to the text file, so the next time the migration starts they are tried to migrate and fail because they have no images saved to the local FS. And unfortunately the succeeded and failed txt files are removed during deployment which made it difficult to experiment on staging.

>The chances of a dataset being only partially reindexed in ES is much lower than partially uploaded to S3. I'm mainly concerned that this logic also makes it harder to fix any datasets that end up in FAILED_DATASETS.txt because it causes them to be skipped even when force=True

This is true, however so far on staging I haven't got a single failed dataset so my hope that there won't be many failed ones on production. And I will be able to fix them by reprocessing if needed.

metaspace/engine/migrations/r1_8_20210322_image_storage.py
```diff
@@ -0,0 +1,283 @@
...
+            processed_ds_ids = read_ds_list('SUCCEEDED_DATASETS.txt') | read_ds_list(
+                'FAILED_DATASETS.txt'
+            )
+            dss_to_process = [ds for ds in dss if ds['id'] not in processed_ds_ids]
+
+            n = len(dss_to_process)
+            tasks = [(ds, i, n) for i, ds in enumerate(dss_to_process, 1)]
+            with ProcessPoolExecutor(max_workers=2) as executor:
+                for logs in executor.map(migrate_dataset, *zip(*tasks)):
+                    print(logs)
```
intsco, line 283: @LachlanStuart I added multiprocessing here as you advised. It made the code a bit more ugly but hopefully it speeds things up a bit so we won't have to wait for the migration to finish for 7 days but for 3 or 4. At least, all 4 cores of the machine seem to be pretty busy with this approach.

metaspace/engine/migrations/r1_8_20210322_image_storage.py
```diff
@@ -0,0 +1,283 @@
...
+        processed_ds_ids = read_ds_list('SUCCEEDED_DATASETS.txt') | read_ds_list(
+            'FAILED_DATASETS.txt'
+        )
+        dss_to_process = [ds for ds in dss if ds['id'] not in processed_ds_ids]
+
+        n = len(dss_to_process)
+        tasks = [(ds, i, n) for i, ds in enumerate(dss_to_process, 1)]
+        with ProcessPoolExecutor(max_workers=4) as executor:
+            for logs in executor.map(migrate_dataset, *zip(*tasks)):
+                print(logs)
```
intsco, line 283: @LachlanStuart I added multiprocessing here as you advised. It made the code a bit more ugly but hopefully it speeds things up a bit so we won't have to wait for the migration to finish for 7 days but for 3 or 4. At least, all 4 cores of the machine seem to be pretty busy with this approach.

metaspace/engine/migrations/r1_8_20210322_image_storage.py
LachlanStuart: Ah, cool. Thanks for clarifying.

metaspace/engine/migrations/r1_8_20210322_image_storage.py
LachlanStuart: The changes look good to me

Issue #803, opened 2021-03-24
Store raw optical images in S3


intsco:
After the new S3 image storage (https://github.com/metaspace2020/metaspace/pull/801) is deployed and all the data is migrated, extra efforts are needed to use S3 for storing raw optical images that users upload on the website. At the moment, they are stored the old way, locally on the web instance.

These are the extra steps:
- [ ] The optical image alignment page should switch to using `uppy` for uploading optical images like the dataset upload and database upload pages do.
- [ ] graphql file upload middleware should be set up to use S3 for (temp) storing raw optical images. These S3 image paths should be sent to the Python code for generating multiple scaled versions of the raw optical image.
- [ ] Old raw optical images should be migrated to the new location and the corresponding database records should be updated.
- [ ] The old image storage server (`storageServer.ts`) should be removed from graphql.
- [ ] Ansible config settings for Nginx should be updated, `/fs/iso_images/` like paths should be removed.
- [ ] The old image files on the web instances at `/opt/data/metaspace/public` should be removed and the instances storage capacity can be reduced from 1-2TB to 100-200GB.

PR #808, opened 2021-03-29
Update python-client to handle the new image URLs served by graphql

LachlanStuart:
As discussed in Slack

To test this I queried the ion images of datasets from Prod (where data has been migrated) and Beta (where it hasn't been migrated). I also checked that optical images on prod still work (they haven't been migrated yet - waiting for #803) and ran the relevant unit tests locally.

As this is urgent, I'll publish the package after opening this PR, however this should still be reviewed and if there's any feedback I'll publish again with the fixes.


sergii-mamedov (approved):
LGTM

Issue #723, opened 2021-01-04
Add python-client implementation & example for dataset uploading using Uppy


intsco:

LachlanStuart:
The old dataset upload code in the python client should be thrown out or overwritten. Only a few collaborators had access to the shared upload buckets, and AFAIK they've all finished what they wanted to do, so we shouldn't keep the old code around after the new code is available.

PR #810, opened 2021-03-30
Rewrite dataset creation using Uppy in python-client.

sergii-mamedov:
After integrating Uppy, we need to change the functions responsible for uploading files to S3 and starting the processing of a new dataset on python-client.

PR was created based on task [723](https://github.com/metaspace2020/metaspace/issues/723).

### DIFF
--- a/metaspace/python-client/docs/source/content/examples/submit-dataset.ipynb
+++ b/metaspace/python-client/docs/source/content/examples/submit-dataset.ipynb
@@ -216,9 +216,8 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "- After submission, you should see the following message\n",
-    "`{'createDataset': '{\"datasetId\": \"your_id\", \"status\": \"success\"}'}`.\n",
-    "Go to [metaspace](http://metaspace2020.eu/) to check your annotations!\n",
+    "- In case of successful submission, you should see the Dataset ID like `2021-01-01_10h11m12s`.\n",
+    "Go to [metaspace](https://metaspace2020.eu) to check your annotations!\n",
     "\n",
     "- If you get `KeyError: 'errors'` make sure that you provided API token and try it again\n",
     "\n",
--- a/metaspace/python-client/docs/source/content/examples/submit-dataset.ipynb
+++ b/metaspace/python-client/docs/source/content/examples/submit-dataset.ipynb
@@ -208,7 +208,7 @@
    "source": [
     "sm.submit_dataset_v3(\n",
     "    imzml_fn, ibd_fn, dataset_name,\n",
-    "    json.dumps(metadata), is_public, mol_dbs=databases\n",
+    "    json.dumps(metadata), is_public, databases\n",
     ")"
    ]
   },
--- a/metaspace/python-client/docs/source/content/examples/submit-dataset.ipynb
+++ b/metaspace/python-client/docs/source/content/examples/submit-dataset.ipynb
@@ -163,7 +163,7 @@
     }
    },
    "source": [
-    "## Dataset visibility (Private/Public)"
+    "## Dataset visibility (Public/Private)"
    ]
   },
   {
@@ -171,7 +171,9 @@
    "metadata": {},
    "source": [
     "We are currently allow users to choose if they want their datasets and annotations to be available publicly or not. \n",
-    "Set the value below to `True` if you want it to be public."
+    "Set the value below to `True` if you want dataset to be public.\n",
+    "If `False`, dataset will only be visible to yourself, other members of your Group,\n",
+    "METASPACE administrators, and members of any Projects you add it to."
    ]
   },
   {
@@ -184,7 +186,7 @@
    },
    "outputs": [],
    "source": [
-    "is_public = False # or True"
+    "is_public = True # or False"
    ]
   },
   {
@@ -206,7 +208,7 @@
    "source": [
     "sm.submit_dataset_v3(\n",
     "    imzml_fn, ibd_fn, dataset_name,\n",
-    "    json.dumps(metadata), is_public=is_public, mol_dbs=databases\n",
+    "    json.dumps(metadata), is_public, mol_dbs=databases\n",
     ")"
    ]
   },
--- a/metaspace/python-client/docs/source/content/examples/submit-dataset.ipynb
+++ b/metaspace/python-client/docs/source/content/examples/submit-dataset.ipynb
@@ -4,23 +4,10 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Submit dataset example"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "Please note that to submit a dataset you will need AWS S3 Metaspace bucket credentials. If you don't have any, please [contact us over email](mailto:contact@metaspace2020.eu)."
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
     "## Log in to METASPACE\n",
     "\n",
-    "To authenticate with METASPACE, generate an API key from your [account page](https://metaspace2020.eu/user/me) and enter it below"
+    "To authenticate with METASPACE, generate an API key from your\n",
+    "[account page](https://metaspace2020.eu/user/me) and enter it below"
    ]
   },
   {
@@ -29,7 +16,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "import json, pprint, getpass\n",
+    "import json, getpass\n",
     "from metaspace import SMInstance\n",
     "\n",
     "sm = SMInstance()\n",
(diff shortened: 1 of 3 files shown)


sergii-mamedov:
@LachlanStuart 
If the proposed option is generally satisfactory, then:

1. I will add a jupyter notebook as an example.
2. Remove deprecated functions: submit_dataset, `submit_dataset_v2`, create_dataset and rename create_dataset_v3.

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -137,33 +137,36 @@ def get_config(
         'moldb_url': '{}/mol_db/v1'.format(host),
         'signin_url': '{}/api_auth/signin'.format(host),
         'gettoken_url': '{}/api_auth/gettoken'.format(host),
-        'companion_url': f'{host}/database_upload',
+        'database_upload_url': f'{host}/database_upload',
```
sergii-mamedov, line 140: changed key for overall consistency

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -173,19 +176,21 @@ def sign_part_upload(key, upload_id, part):
...
+    def complete_multipart_upload(key, upload_id, etags, headers={}):
         query = urllib.parse.urlencode({'key': key})
         url = f'{companion_url}/s3/multipart/{upload_id}/complete?{query}'
         data = {'parts': [{'PartNumber': part, 'ETag': etag} for part, etag in etags]}
-        resp_data = send_request(url, 'POST', json=data)
+        resp_data = send_request(url, 'POST', json=data, headers=headers)
         location = urllib.parse.unquote(resp_data['location'])
         (bucket,) = re.findall(r'https://([^.]+)', location)
-        return f's3://{bucket}/{key}'
+        return f's3a://{bucket}/{key}'
```
sergii-mamedov, line 191: I was confused why we didn't use s3**a**

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -268,7 +273,8 @@ def get_primary_group_id(self):
               }
             }
         """
-        return self.query(query)['currentUser']['primaryGroup'].get('group', {}).get('id', None)
+        primary_group = self.query(query)['currentUser']
+        return primary_group.get('group', {}).get('id', None) if not primary_group else None
```
sergii-mamedov, line 277: An error occurred if the user is not a member of any group.

metaspace/python-client/metaspace/sm_annotation_utils.py
LachlanStuart: We enforce that dataset paths start with `s3a://` and database paths start with `s3://`. I think this needs to handle both cases, because uploading databases now fails graphql request validation because of the `s3a://`. Alternatively, change graphql to also allow `s3a` for databases.

AFAICT, the reason whether to use `s3` or `s3a` depends on which tool is used. Spark, Hadoop, etc. use it to determine which library/protocol to use, so `s3a` is needed to pick the most recent protocol. Lithops and the AWS CLI just use the URIs for addressing, so `s3://` just means a file is on S3, and no library/protocol is specified.

It doesn't matter so much anymore in our case, because it seems sm-engine now always manually parses the URLs into `bucket` and `key` before passing those values to boto/Lithops.

metaspace/python-client/metaspace/sm_annotation_utils.py
LachlanStuart: This now always returns `None`. It should be:
```python
return primary_group.get('primaryGroup', {}).get('group', {}).get('id', None) if primary_group else None
```

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -1435,6 +1498,42 @@ def submit_dataset_v2(
...
+        mol_dbs: Union[List[str], List[int]] = None,
+        adducts: List[str] = None,
+    ) -> dict:
+        """Submit a dataset for processing in Metaspace.
+
+        Args:
+            ibd_fn: ibd file name.
+            imzml_fn: imzml file name.
+            dataset_name: Dataset name.
+            metadata: Properly formatted metadata json string.
```
LachlanStuart, line 1517: Requiring the user to JSON-serialize `metadata` is passing unnecessary complexity to the user. Most of our users don't really deal with web APIs that much, so could we handle the JSON serialization internally to save them some difficulty?

It will also help if we decide to later add Python typings for the metadata parameter.

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -1435,6 +1498,42 @@ def submit_dataset_v2(
             adducts=adducts,
         )
 
+    def submit_dataset_v3(
+        self,
+        ibd_fn: str,
+        imzml_fn: str,
```
LachlanStuart, line 1504: Please put the imzML path first, as it's the "main" file.

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -1435,6 +1498,42 @@ def submit_dataset_v2(
...
+        dataset_name: str,
+        metadata: str,
+        is_public: bool = False,
```
LachlanStuart, line 1507: Please make `is_public` a required parameter with no default value. METASPACE's value as a community repository is driven by public datasets. We don't want datasets to be private by default, as it devalues the platform. 

However, I also don't think it should default to `True` here, because sharing the dataset without the user's explicit permission would be a breach of trust and probably a copyright violation. 

Please also revert the lines of the notebook that set `is_public = False` and update the API docstring to be something clearer like "is_public: If True, the dataset will be publicly visible. If False, it will only be visible to yourself, other members of your Group, METASPACE administrators, and members of any Projects you add it to".

metaspace/python-client/docs/source/content/examples/submit-dataset.ipynb
```diff
@@ -218,39 +204,24 @@
...
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "- After submission you should see the following message\n",
-    "{'createDataset': '{\"datasetId\": \"your_id\", \"status\": \"success\"}'}\n",
-    "Go to [metaspace](http://metaspace2020.eu/) to check your annotatoins!\n",
-    "\n",
-    "- If you get _KeyError: 'errors'_ make sure that you provided API token and try it again\n",
+    "- After submission, you should see the following message\n",
+    "`{'createDataset': '{\"datasetId\": \"your_id\", \"status\": \"success\"}'}`.\n",
```
LachlanStuart, line 218: Could you please change it to parse the output JSON and return the new dataset ID? We've had other projects where we've had to manually parse it in the calling code, and it's a hassle. Since this is a new API, we should try to clean it up.

It's safe to assume that only 2 return paths are valid: returning an ID, and throwing an exception.

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -1435,6 +1498,42 @@ def submit_dataset_v2(
...
+        """
+        polarity = json.loads(metadata)['MS_Analysis']['Polarity']
+        if not adducts:
+            if polarity == 'Positive':
+                adducts = ['+H', '+Na', '+K']
+            elif polarity == 'Negative':
+                adducts = ['-H', '+Cl']
+            else:
+                raise Exception(
+                    'Polarization set incorrectly, available only "Positive" or "Negative"'
```
LachlanStuart, line 1530: `Polarization` -> `Polarity`

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -1435,6 +1498,42 @@ def submit_dataset_v2(
...
         )
 
+    def submit_dataset_v3(
+        self,
+        ibd_fn: str,
+        imzml_fn: str,
+        dataset_name: str,
+        metadata: str,
+        is_public: bool = False,
+        mol_dbs: Union[List[str], List[int]] = None,
```
LachlanStuart, line 1508: It seems like there's an API error if `databases` is null. I think this should default to an empty array.

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -1435,6 +1498,42 @@ def submit_dataset_v2(
...
+        adducts: List[str] = None,
+    ) -> dict:
```
LachlanStuart, line 1510: Could you please add an argument for `project_ids`. If people are going to programmatically submit huge numbers of datasets, it'd be helpful if they can use projects to keep them organized.

LachlanStuart (approved):
I'm approving with feedback. This means that there are probably changes that need to be made, but I don't feel that a re-review will be necessary after the requested changes are made. Feel free to re-request a review if you add anything more, or feel they should be re-reviewed.

As with the other PR, I'll highlight the feedback items that must be addressed (either through changes, or conversation) before merging:
* All of the comments about the API of `submit_dataset_v3` (order of parameters, `is_public` default value & docstring, `mol_dbs` default value, `metadata` type, `project_ids`, parsing the return value)
* The `s3a://` and `get_primary_group_id` changes that broke database upload

Additionally, AFAIK nobody is using the database upload feature yet, so if you don't have time to make the fixes before you want to demo, feel free to publish this code as it is, as long as it's fixed later.

Issue #128, opened 2018-11-16
Improve optical image blending


LachlanStuart:
Generate a bunch of overlaid images and share with everyone

- Prop opaticy of opt img can be implemented by using white color for all pixels of the ion img and set their transparency inv. prop. to the intensity

- Examples:
- http://metaspace2020.eu/#/annotations?ds=2018-06-29_05h38m12s25_15h28m52s
- http://metaspace2020.eu/#/annotations?ds=2017-06-30_07h26m26s
- http://metaspace2020.eu/#/annotations?ds=2017-05-17_19h49m04s

Issue #812, opened 2021-03-31
Ion image download problem


lmacielvieira:
When in the annotation page, if we try to download the ion image does not work when the microscopy image is shown.

Tested browser:
Chrome

PR #815, opened 2021-04-07
Feat/plus one overlay

lmacielvieira:
### Description

In some cases, at the annotation viewer, it is difficult to see the ion image given some unique features from the correspondent optical image. In order to improve user experience, a new slider that can change the optical image visibility was added.

Issue #812

###### ImageViewer page
- [x] Added "Optical image visibility" slider
- [x] Changed "Opacity" slider label to "Ion image opacity"
- [x] Changed Sliders disposition. Opacity, visibility and lockIntensity sliders in a row disposition and as column on small screen
- [x] Hid "Optical image visibility" slider in IE, as the filter props do not work on it.

###### IonImageViewerComponent
- [x] Added crossOrigin props on optical img tag to enable download from s3 without cors problems
- [x] Added opticalOpacity props, which changes visibility of the optical image. The formula changes the image to grayscale and reduces the brightness with 75% of the correspondent grayscale percentage value

***Note
Although there is a code repeat on the filter opticalOpacity formula formula on both optical images tags. This was done so the props changes can be listened without needing a specific listener.

 #### Tests
 
 ##### Front end
 
- [ ] Unit Test
- [ ] e2e Test
 
- [x] Chrome
- [x] Safari
- [x] sm, md, lg, xl, lg
* Not supported on IE11

image
image

### DIFF
--- a/metaspace/webapp/src/components/IonImageViewer.tsx
+++ b/metaspace/webapp/src/components/IonImageViewer.tsx
@@ -339,6 +339,7 @@ const useBufferedOpticalImage = (props: Props) => {
       {opticalImageUrl.value
       && <img
         key={state.loadedOpticalImageUrl}
+        crossOrigin="anonymous"
         src={state.loadedOpticalImageUrl}
         class="absolute top-0 left-0 -z-10 origin-top-left"
         style={{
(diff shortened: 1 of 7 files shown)


LachlanStuart (approved):
Nice work! This also resolves #128

PR #833, opened 2021-04-26
Python client update_dataset support + docs improvements

LachlanStuart:
* Added `SMInstance.update_dataset` to make it possible to edit datasets' neutral losses, etc.
* Change `SMDataset.metadata` to be accessible as a dictionary. Previously it just held a property `Metadata.json` that returned the JSON string. I had to do some hackery to get it working as a TypedDict that also retained the `.json` property for backwards compatibility.
* Add `TypedDict` declarations for `SMDataset.config`
* Moved the `DatasetDownload` `TypedDict` declarations into `metaspace.types`, alongside the new `Metadata` and `DSConfig` declarations
* Explicitly declare most of the available arguments to `SMInstance.datasets`, because it's one of the main APIs new users will look at, and it wasn't generating useful documentation.
* Tidy up the docs by reordering module members and excluding a bunch of private methods & data
* Remove non-functional/unused `ion` function, `SMInstance.adducts` method, and `NYI` exception from `sm_annotation_utils`

As a team member is waiting for the `SMInstance.update_dataset` change, I've released this to PyPI already.

### DIFF
--- a/metaspace/python-client/metaspace/types.py
+++ b/metaspace/python-client/metaspace/types.py
@@ -1,0 +1,144 @@
+import json
+from typing import Optional, cast, List
+
+# NOTE: Python-client needs to support Python versions as low as Python 3.6.
+# Make sure to run PyCharm's "Code Compatibility" inspection when touching any imports from typing
+try:
+    from typing import TypedDict, Literal  # Python 3.8+
+
+    Polarity = Literal['Positive', 'Negative']
+    MetadataDataType = Literal['Imaging MS']
+except ImportError:
+    try:
+        from typing_extensions import TypedDict, Literal
+
+        Polarity = Literal['Positive', 'Negative']
+        MetadataDataType = Literal['Imaging MS']
+    except ImportError:
+        Polarity = str
+        MetadataDataType = str
+        TypedDict = dict
+
+
+class MetadataSampleInformation(TypedDict):
+    Organism: str
+    Organism_Part: str
+    Condition: str
+    Sample_Growth_Conditions: Optional[str]
+
+
+class MetadataSamplePreparation(TypedDict):
+    Sample_Stabilisation: str
+    Tissue_Modification: str
+    MALDI_Matrix: str
+    MALDI_Matrix_Application: str
+    Solvent: Optional[str]
+
+
+class MetadataResolvingPower(TypedDict):
+    mz: float
+    Resolving_Power: float
+
+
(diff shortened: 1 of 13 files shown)


metaspace/python-client/setup.py
```diff
@@ -19,15 +19,7 @@ def read_version():
...
-    install_requires=[
-        'pandas',
-        'plotly>=1.12',
-        'numpy',
-        'matplotlib',
-        'pyMSpec',
-        'pillow',
-        'requests',
-    ],
+    install_requires=['pandas', 'plotly>=1.12', 'numpy', 'matplotlib', 'pillow', 'requests'],
```
sergii-mamedov, line 22: Should I specify specific versions of libraries in the future? In my experience, this is usually done to minimize problems.

sergii-mamedov (approved):
I have no comments, just a couple of questions above.

metaspace/python-client/setup.py
LachlanStuart: My general rule: yes at the Application level (i.e. the `engine`, `graphql` and `webapp` projects), but no at the Library level (`python-client`, and the upcoming `msi_recal` library for recalibration).

The reason is that version stability is important for the stability of applications, so it needs to be strict somewhere. If versions are strict at the library level, it will often cause conflicts between libraries. Applications are usually packaged and tested as monolithic unit, so it makes sense to fix versions at the application level, and only add rules at the library level if there are known incompatibilities.

metaspace/python-client/metaspace/types.py
```diff
@@ -0,0 +1,156 @@
...
+    Polarity = Literal['Positive', 'Negative']
+    MetadataDataType = Literal['Imaging MS']
+except ImportError:
+    try:
+        from typing_extensions import TypedDict, Literal
+
+        Polarity = Literal['Positive', 'Negative']
+        MetadataDataType = Literal['Imaging MS']
+    except ImportError:
+        Polarity = str
```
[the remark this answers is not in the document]
LachlanStuart: I tested it on 3.6 and 3.8. I know 3.7 also doesn't fully support `TypedDict` and `Literal`. Python's typing is unfortunately still a work-in-progress, and even 3.9 has big shortcomings compared to what I'm used to with TypeScript.

Regarding which version to support - I have no strong opinions, but deprecating a version of Python will require some change management to ensure users aren't disrupted. I try to keep it 3.6-compatible just so I don't have to do this extra work.

PR #837, opened 2021-05-04
Set useLithops True for GraphQL API

sergii-mamedov:
At the moment, all clients of the GraphQL API used useLithops=false by default. In this regard, we have to explicitly register the processing of datasets through Lithops in the python-client, metaspace site, ... Since we do not see any obvious problems with the use of Lithops, it was decided to switch used to Lithops by default.

### DIFF
--- a/metaspace/python-client/metaspace/sm_annotation_utils.py
+++ b/metaspace/python-client/metaspace/sm_annotation_utils.py
@@ -667,7 +667,12 @@ class GraphQLClient(object):
         return self.query(query, variables)
 
     def update_dataset(
-        self, ds_id, input=None, reprocess=False, force=False, priority=1,
+        self,
+        ds_id,
+        input=None,
+        reprocess=False,
+        force=False,
+        priority=1,
     ):
         query = """
             mutation updateMetadataDatabases($id: String!, $reprocess: Boolean,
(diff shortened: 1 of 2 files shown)


sergii-mamedov:
It's a bit strange, why did the test coverage decrease if I didn't add or remove any functions?

LachlanStuart:
> It's a bit strange, why did the test coverage decrease if I didn't add or remove any functions?

CodeCov is quite flaky unfortunately. 

Most of the time when it behaves unexpectedly, it's due to problems detecting the correct base commit. If a branch wasn't branched from the latest commit in `master`, usually CodeCov will compare coverage against another commit from the same branch. However, that doesn't seem to have happened this time... 

Looking at [the coverage changes](https://app.codecov.io/gh/metaspace2020/metaspace/compare/837/changes), I have no idea what could have changed. It's very weird. I think it's safe to ignore in this case.

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -667,7 +667,12 @@ def delete_dataset(self, ds_id, force=False):
         return self.query(query, variables)
 
     def update_dataset(
-        self, ds_id, input=None, reprocess=False, force=False, priority=1,
+        self,
+        ds_id,
+        input=None,
+        reprocess=False,
+        force=False,
+        priority=1,
```
LachlanStuart, line 675: Thanks for fixing this. Sorry I let it get into master.

LachlanStuart (approved):

PR #838, opened 2021-05-04
Feat/ds comparison page

lmacielvieira:
### Description

In order to improve user experience and allow the user to compare annotation results among different datasets without having to merge the datasets into a single file, a page to compare Dataset annotations was created.

###### Dataset overview page
- [x] Added "Compare with other datasets..." option to DatasetActionsDropDown

###### DatasetComparisonDialog
- [x] Created component to build dataset comparison layout
- [x] Made the available datasets selectable for comparison
- [x] Created a grid layout based on number of rows and columns
- [x] Set the maximum grid elements to 15 cells
- [x] Enable the distribution of selected datasets to the created grid layout
- [x] Save the grid comparison settings using the existing saveImageViewerSnapshot mutation

###### DatasetComparisonPage
- [x] Created page to compare datasets according to settings chosen on the DatasetComparisonDialog
- [x] Enabled Filtering like in annotation page
- [x] Hid datasetIds filter 
- [x] Display DatasetComparisonAnnotationTable based in the allAggregatedAnnotations result
- [x] Display DatasetComparisonGrid based in the allAggregatedAnnotations result and DatasetComparisonAnnotationTable row selection
- [x] Display related molecules section

###### DatasetComparisonAnnotationTable
- [x] Display annotations passed by props
- [x] Enabled sorting by msmScore, fdrLevel and sumFormula
- [x] Persist row selection and sorting options by query parameters
- [x] Replicate CSV download from annotationTable
- [x] Emit row change event

###### DatasetComparisonGrid
- [x] Display IonImages according to annotations passed via props
- [x] Display IonImages of each dataset according to gridSettings received
- [x] Created multi state IonImageViewer state handler
- [x] Created Grid Image download feature

###### setupTestFramework
- [x] Changed ElementUi stub declaration schema as components like el-collapse were not being correctly replaced

######  Annotation
- [x] Added allAggregatedAnnotations, which is based on allAnnotations query, but uses es aggregation feature to aggregate by annotaion and db, and divide the annotations results by datasets

######  EsConnector
- [x] Added elasticSearch raw aggregation query
 
 
 #### Tests
 
 ##### Front end
 
- [x] Unit Test
- [ ] e2e Test
 
- [x] Chrome
- [x]  Safari
- [x] sm, md, lg, xl, lg

##### Remarks

- Es aggregation feature has a limitation, where dos not accept limit and offset parameters as the normal query. The pagination could be done on ES side, but it would also need to get all results first and aggregate it just afterwards. This can be a potential problem for comparison of datasets with a greater number of annotations.
- If we used the already existing aggregation query, we would have to return all the data points also, as they would be aggregated on the client side afterwards.
- A new AnnotationTable was created specifically to the dataset comparison page, as the existing one adds an overhead by being based on the existing graphql/Es structure, what would add an unnecessary complexity by activating it or not based on the props passed. The new AnnotationTable (DatasetComparisonAnnotationTable) does not make any graphql connection and its rendering, pagination and sorting are done in place according to the data passed via props.
- The feature of multi annotation selection for the grid Imageviewer and DatasetComparisonAnnotationTable are yet to be implemented.
- There are a lot variable checking (?. and ? : ) for the ImageViewer state handler in the DatasetComparisonGrid page (although you always recommend not to use it on your PRs review), given that the used components inside it gives a lot of errors and warning if not default value are passed. As the grid takes a time to be built, if removed same unexpected behaviour might happen, but I can take a deeper look to improve it later on.
- In order to try to improve the performance, the ImageViewerGrid and DatasetComparisonAnnotationTable have their internal state logic based on changes checked on the watchEffect hook. I will take a deeper look into scenarios that I might have not have mapped and are needed to correctly display then.
- In order to test the new page: 
    1. Access dataset list page with overview feature activated "datasets?feat=show_dataset_overview"
    2. Go to Dataset overview page, click the "Compare with other datasets..." in the action dropdown
    3. Set your grid settings
    4. You will be redirected to the DatasetComparison page 

image
image
image


LachlanStuart (commented):
This review is still WIP. I have a few more things to manually investigate, but I may not get back to it until late tomorrow and didn't want to sit on unsent feedback. Sorry once again the review is taking so long. 

So far it looks good. I'm still amazed you got this working so fast.

The only blocking bit of feedback so far is about the `datasets` field name in the graphql schema. The rest is of my feedback so far is UX and as it's all hidden behind a feature flag, it's not a problem if it gets deployed to get feedback from others.

I have a few areas of concern that I haven't fully investigated yet:
1. Sometimes the image viewer takes multiple seconds to update to the new ion image for me. I used a 1 column 2 row arrangement with 2 datasets, where both datasets had optical images, on Firefox in case that matters. It was most clear when moving from an annotation that only had an image for the bottom dataset to an annotation that had images for both. The bottom image viewer would update almost immediately, but the top one would just hang for a while with "No data". If I moused over the bottom image, the top image would immediately update, so I suspect it could be a reactivity issue. 
2. I haven't checked what the logic behind the columns is, but showing a single MSM or FDR value doesn't really make sense because the MSM/FDR values will be different for each dataset. IMO there are 3 valid options here:
    a. If it's an aggregate statistic, make it clearer, i.e. "Best MSM/FDR" or "Average MSM/FDR"
    b. Just show the m/z. It's (theoretically) the same for all annotations with the same ion. Technically it might be slightly different between the datasets, but it's inconsequential so it's fine to just pick a value
    c. Show one column per dataset for each value. IMO this should be the eventual goal, but it would be a lot of columns to show at once, so we'd also want a UI for selecting which field to show.
3. I think the default sort is backwards. It seemed to show me the worst annotations first. The default sort order should be FDR ascending, then MSM descending (e.g. 5% 0.999 then 5% 0.800 then 10% 0.999 then 10% 0.800).
4. I suspect scientists would prefer the Export to CSV to be in a pivot table format (i.e. 1 row per ion, and columns for each dataset+field combination), but I suggest you check with them when you get feedback.

metaspace/webapp/src/modules/Datasets/comparison/DatasetComparisonPage.tsx
```diff
@@ -0,0 +1,207 @@
...
+  // @ts-ignore
+  setup(props, { refs, root }) {
+    const fetchImageViewerSnapshot = gql`query fetchImageViewerSnapshot($id: String!, $datasetId: String!) {
+      imageViewerSnapshot(id: $id, datasetId: $datasetId) {
+        snapshot
+      }
+    }`
+    const { $route, $store } = root
+    const gridNode = ref(null)
+    const state = reactive<any>({
```
LachlanStuart, line 52: It's better to mark individual fields as `any` than to mark the whole object as `any`. If you do `reactive<any>`, it disables all type inference for the whole `state`, so it won't even detect typos like `state.gird`. If you mark individual fields as `any`, TypeScript can suggest field names, detect typos, and ensure the correct types for fields that are simple types.
e.g.

```
const state = reactive({
    selectedAnnotation: -1, // TypeScript can infer this is a Number
    gridState: {} as any, // TypeScript can't guess the right type here, so override it to "any"
    annotations: [] as any[], // "any[]" is preferable to "any" for arrays
})
```

LachlanStuart (changes requested):
The code generally looks good, but I'm concerned that the UX hasn't been tested by users yet. I would suggest asking some scientists on our team for feedback. I suspect it may take some time to get feedback and may lead to significant changes, so I would suggest doing this ASAP, skipping any feedback I've given if it would take too long to address.

As this is feature-flagged, it's not an issue if this is deployed with bugs. The only critical feedback items are:
* Renaming the GraphQL field from `datasets` to `annotations` (as it's much more work to rename fields after a graphql schema has been deployed, due to backwards compatibility issues)
* Moving the ES query structure into `esConnector.ts`

metaspace/webapp/src/modules/Datasets/comparison/DatasetComparisonGrid.tsx
```diff
@@ -0,0 +1,565 @@
...
+import fitImageToArea from '../../../lib/fitImageToArea'
+import FadeTransition from '../../../components/FadeTransition'
+import OpacitySettings from '../../ImageViewer/OpacitySettings.vue'
+import RangeSlider from '../../../components/Slider/RangeSlider.vue'
+import IonIntensity from '../../ImageViewer/IonIntensity.vue'
+import ImageSaver from '../../ImageViewer/ImageSaver.vue'
+import getColorScale from '../../../lib/getColorScale'
+import { THUMB_WIDTH } from '../../../components/Slider'
+import { isEqual } from 'lodash-es'
+import { processingSettingsChanged } from '../../../../../graphql/src/modules/dataset/controller/Mutation'
```
LachlanStuart, line 18: This line

LachlanStuart (approved):
Looks good. Sorry for the last-minute change request, but please update the graphql field name/type to match other instances of `databaseId`. I'm happy for this to be merged after that's addressed.

Issue #830, opened 2021-04-26
[ProjectListPage] Remember state of My projects drop-down


lmacielvieira:
In order to improve the user experience, so he can find the desired project faster, the last option selected on the my datasets filter must be saved.

- [ ] Remove 'my projects' option persistence by url parameter
- [ ] Persist the 'my projects' option on user settings state

Issue #829, opened 2021-04-26
[DatasetsListPage] Remember the state of My Datasets drop-down


lmacielvieira:
In order to improve the user experience, so he can find the desired dataset faster, the last option selected on the my datasets filter must be saved.

PR #851, opened 2021-05-10
Feat/group projects filter persist

lmacielvieira:
### Description

In order to improve user experience and allow the user filter settings regarding "My projects" and "My datasets" sould be persisted even without passing a query param.
#830 #829 

###### Dataset Table page
- [x] Setting initial value of "My datasets" filter with local storage stored value if query param not set and user logged

###### Project List page
- [x] Setting initial value of "My projects" filter with local storage stored value if query param not set and user logged

###### Filter Panel
- [x] Saving "datasetOwner" filter value on local storage
- [x] Saving "simpleFilter" filter value on local storage

 
 #### Tests
 
 ##### Front end
 
- [x] Unit Test
- [ ] e2e Test
 
- [x] Chrome
- [x]  Safari
- [x] sm, md, lg, xl, lg


LachlanStuart (approved):
Looks good

Issue #849, opened 2021-05-10
Fix load annotations from reprocessed annotations permalink


lmacielvieira:
In some cases, when the dataset is reprocessed some annotations can change or even be removed. In a case where a user had previously generated a permalink to access the annotation page and the annotation was removed/changed, it can happen to lead the image viewer to a misleading behaviour where the image is not shown and the user does not have indications of what happened.

- [ ] Check if the annotations loaded from the permalink (snapshotId) still exist
- [ ] Display the image without the ion layers if the annotation does not exist anymore

Issue #790, opened 2021-03-15
Lithops: Do partial reprocessing when new databases are added


LachlanStuart:
The Lithops annotation pipeline currently [reprocesses all databases](https://github.com/metaspace2020/metaspace/blob/master/metaspace/engine/sm/engine/annotation_lithops/annotation_job.py#L310-L312). This can cause significant repeated work. Copy across the logic that's in the Spark implementation: 
* If the processing settings (the `ds_config` excluding the database field) has changed, always reprocess all DBs (IIRC this is also indicated by a `reprocess` flag with the annotate message)
* If databases have been removed, delete their corresponding `job` entries in postgres/elasticsearch but don't reannotate
* If databases have been added, annotate only the new databases

PR #855, opened 2021-05-14
Bug/fix reprocessed permalink

lmacielvieira:
### Description

In some cases, when the dataset is reprocessed some annotations can change or even be removed. In a case where a user had previously generated a permalink to access the annotation page and the annotation was removed/changed, it can happen to lead the image viewer to a misleading behaviour where the image is not shown and the user does not have indications of what happened.
#849 

###### AnnotationsPage
- [x] Displays an alert warning the user that: "This dataset has been reprocessed and some of the annotations used for this link no longer exist. Please select another annotation and generate the link again." if the viewId and datasetId query params are passed, but the related annotation was reprocessed
- [x] Hides the new filter "AnnotationIds" if the annotations are loaded from a snapshot
- [x] Displays the new filter "AnnotationIds" if the annotations are loaded from a snapshot

###### AnnotationTable
- [x] Added annotationIds param to allAnnotations query
- [x] Stopped selecting a row and updating the imageViewer at first load when loading from a snapshot
- [x] Selecting default row as the one from the snapshot if loading from snapshot
- [x] Selecting no row as if the one from the snapshot is not being rendered on the current table page
- [x] Filtering by annotationIds if loading from snapshot with multiple annotations

###### Filter
- [x] Added annotationIds filter
- [x] Added hiddenFilters option, which hide the filter passed on it

 ###### ionImageState
- [x] Returns default intensity if annotation does not exist, so it not breaks the viewer if annotation is not passed

 ###### useRestoredState
- [x] Set default annotation as loaded from snapshot
- [x] Marks annotation as reprocessed if it does not exists anymore

 ###### Share link
- [x] Adding ions and related databases to snapshot, so if the dataset is reprocessed and the annotation deleted we can inform which ions it was related to

###### Annotation
- [x] Added annotation ids term filter (annId)

##### Front end
 
- [x] Unit Test
- [ ] e2e Test
 
- [x] Chrome
- [x]  Safari
- [x] sm, md, lg, xl, lg

#### Scenarios

Some improvements and behaviour changes were applied to the load from snapshot. The previous behaviour always create a link with the molecule shared, but this could be confusing if you are sharing a multi molecule comparison link. Although the share link generated link was kept and still works the same when the query param "mol" is passed, I added the following support to the link, when the datasetIds and viewId params are used (and the mol one is not):
1 - If the dataset was reprocessed, the annotation does not exists anymore and the mol query param is set, the page is not breaking anymore and shows the first annotation filtered by mol with the snapshot settings applied.
2 - If the dataset was reprocessed, the annotation does not exists anymore and the mol query param is not set, the page is showing an alert on the image viewer saying: "This dataset has been reprocessed and some of the annotations used for this link no longer exist. Please select another annotation and generate the link again.".
3 - If the dataset was uploaded after this change and it was reprocessed, the annotation does not exists anymore and the mol query param is not set, the page is showing an alert on the image viewer saying: "This dataset has been reprocessed and some of the annotations used for this link no longer exist. Please select another annotation and generate the link again." and the list of ions and related databases related to the previous annotation from the link.
4 - If the annotation is loaded from the snapshot and the mol query param is not set, the page is showing the annotation and viewer settings according to the loaded snapshot (on the previous behaviour this was being  replaced by the annotation table initial load).
5 - If the annotation is loaded from the snapshot, the mol query param is not set, and the snapshot is related with many molecules, the page is showing the annotation and viewer settings according to the loaded snapshot (on the previous behaviour this was being  replaced by the annotation table initial load) and the annotationIds filter is automatically applied.

image
exists
image
image


LachlanStuart (changes requested):
The new error message is a big UX improvement over the current JS errors, so it should still be merged after the requested fixes, but I feel there is some more work to be done on this task.

This doesn't really solve the problem that has been discussed. My understanding was that this task was to prevent links from breaking due to reprocessing (except in the rare cases where molecules weren't reannotated as a result of processing changes), but they still break even with this PR. If this task has turned out to be harder than initially expected to fix this, we should discuss it again at Sprint XL. One alternative solution would be to prioritize #790, which would mitigate this on the back end by reducing the unnecessary deletion & recreation of annotations.

However, this PR still gives an overall improvement - the error message is good, and saving the selected molecules at least gives us the opportunity to fix future links. There are just a couple things I feel need to be addressed before merging:
* The graphql parameter name `annId` -> `annotationId`
* The longer URLs due to the added `annotationId` parameter. The ideal fix would be prevent filters from being added to the URL if they're already contained in the snapshot. However, it would also be acceptable to change the share link menu to directly give the user the generated URL (e.g. YouTube and Reddit's "Share" menus that show the URL in a textbox or have a "Copy link" button) so that they aren't likely to use the long URL for sharing.

metaspace/webapp/src/store/mutations.js
```diff
@@ -41,7 +42,7 @@ function sortFilterKeys(keys) {
   keys.sort((a, b) => compare(FILTER_SPECIFICATIONS[a].sortOrder || 100, FILTER_SPECIFICATIONS[b].sortOrder || 100));
 }
 
-function updateFilter(state, filter, routerAction = null) {
+async function updateFilter(state, filter, routerAction = null) {
```
LachlanStuart, line 45: This `async` doesn't seem to be needed. As Vuex is a reducer-pattern state store, mutations should never have async logic - that should go in actions instead.

LachlanStuart (approved):
One fix needed before merging - the unused `async` in mutations.js. Otherwise LGTM.

Issue #858, opened 2021-05-25
Allow HMDB to be excluded for API-submitted datasets


LachlanStuart:
We've had several internal projects where HMDB has been unwanted, but the current logic makes it impossible to exclude databases with the `default` flag. They just get re-added in the sm-api Python code. Ideally we'd make it so that the defaults were applied for all web-submitted datasets, but were easy to bypass for API-submitted datasets. The cleanest way to do this is to make it impossible to de-select HMDB when submitting a database through webapp, but have no code in graphql/sm-api to enforce this.

**Upload page**
- [ ] Make it so that HMDB-v4 is selected by default and cannot be un-selected
    * Ideally this logic would be based on the graphql `default` field - all `default` datasets should be selected. However, IDK how complex the dataflow is for this field, so if it's too much trouble to use the `default` field, it's also acceptable to just hard-code a match for the `name == 'HMDB' && version == 'v4'`. 
- [ ] The maximum database limit should be raised by 1 (or the default DBs shouldn't count towards it), because some people relied on the trick of removing HMDB to get 1 more selectable DB

**Edit Metadata page**
- [ ] HMDB should only be un-selectable if the dataset was annotated with it. It should be possible to edit a dataset that doesn't have HMDB without the page forcing HMDB to be added. It should be impossible to remove HMDB if the dataset already had HMDB.

**sm-api**
- [ ] Remove [the logic that adds the default DBs](https://github.com/metaspace2020/metaspace/blob/master/metaspace/engine/sm/rest/dataset_manager.py#L71-L72). In case the line numbers change, it's these two lines in the `add` method:
```python
if 'moldb_ids' in doc:
    doc['moldb_ids'] = self._add_default_moldbs(doc['moldb_ids'])
```

**Python client**
- [ ] Change the default value for the [`submit_dataset` `databases` parameter](https://github.com/metaspace2020/metaspace/blob/master/metaspace/python-client/metaspace/sm_annotation_utils.py#L1465) to `[DEFAULT_DATABASE]`

PR #861, opened 2021-05-27
Changed default db logic from backend to frontend

lmacielvieira:
### Description

"We've had several internal projects where HMDB has been unwanted, but the current logic makes it impossible to exclude databases with the default flag. They just get re-added in the sm-api Python code. Ideally we'd make it so that the defaults were applied for all web-submitted datasets, but were easy to bypass for API-submitted datasets. The cleanest way to do this is to make it impossible to de-select HMDB when submitting a database through webapp, but have no code in graphql/sm-api to enforce this".
#858

###### MetadataEditor
- [x] Add default database to database list if new dataset
- [x] The default database can not be removed from the database list if new dataset
- [x] The default database can not be removed from the database list if updating a dataset which was already annotated with the default db
- [x] The default database can be removed/added from the database list if updating a dataset which was already annotated without the default db
- [x] Increased database limit to 4

###### dataset_manager
- [x] Removed the logic that adds the default DBs. 

###### sm_annotation_utils
- [x] Changed the default value for the submit_dataset databases parameter to [DEFAULT_DATABASE]

##### Front end
 
- [x] Unit Test
- [ ] e2e Test
 
- [x] Chrome
- [x]  Safari
- [x] sm, md, lg, xl, lg

image

### DIFF
--- a/metaspace/webapp/src/lib/constants.ts
+++ b/metaspace/webapp/src/lib/constants.ts
@@ -1,4 +1,4 @@
-export const MAX_MOL_DBS = 3
+export const MAX_MOL_DBS = 4
 export const MAX_MOL_DBS_EXT = 5
 export const MAX_NEUTRAL_LOSSES = 3
 export const MAX_CHEM_MODS = 3
--- a/metaspace/webapp/src/lib/constants.ts
+++ b/metaspace/webapp/src/lib/constants.ts
@@ -1,5 +1,5 @@
 export const MAX_MOL_DBS = 4
-export const MAX_MOL_DBS_EXT = 5
+export const MAX_MOL_DBS_EXT = 6
 export const MAX_NEUTRAL_LOSSES = 3
 export const MAX_CHEM_MODS = 3
 export const DEFAULT_SCALE_TYPE = 'linear'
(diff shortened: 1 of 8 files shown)


metaspace/webapp/src/lib/constants.ts
```diff
@@ -1,4 +1,4 @@
-export const MAX_MOL_DBS = 3
+export const MAX_MOL_DBS = 4
 export const MAX_MOL_DBS_EXT = 5
```
LachlanStuart, line 2: Please increment `MAX_MOL_DBS_EXT` as well. It's used when the `moldb_limit_ext` is set - we allow powerusers to select more DBs, and several people in the team currently rely on this plus the trick of removing HMDB to get 5 non-HMDB DBs

metaspace/webapp/src/modules/MetadataEditor/MetadataEditor.vue
```diff
@@ -145,6 +146,7 @@ export default {
       schema: null,
       loadingPromise: null,
       localErrors: {},
+      defaultDb: {},
```
LachlanStuart, line 149: Please use `null` instead of `{}` when you mean "no value" / "nothing". It makes the intent of the code much clearer, and is easier to work with in TypeScript (e.g. with `SomeType | null` you can just use `?.`, whereas with `SomeType | {}` you need an awkward `if ('id' in defaultDb)` guard to access values in a safely-typed way)

LachlanStuart (approved):
Looks good, but please address both comments before merging.

Issue #864, opened 2021-06-07
Change permalink storing logic


lmacielvieira:
Currently, the permalink generated when sharing a specific annotation saves internally the dsId, settings and annotationsIds. When a dataset is reprocessed, the annotation ids are updated and the permalink info is not enough to rebuild the shared annotation. In order to solve this problem, the permalink logic should change, so that instead of the annotation id, the used ions are saved.

- [ ] Retrieve annotation using the ion formula
- [ ] Build a migration to update the already existing permalinks to the new schema

PR #879, opened 2021-06-16
Change permalink storing logic

lmacielvieira:
### Description

Currently, the permalink generated when sharing a specific annotation saves internally the dsId, settings and annotationsIds. When a dataset is reprocessed, the annotation ids are updated and the permalink info is not enough to rebuild the shared annotation. In order to solve this problem, the permalink logic should change, so that instead of the annotation id, the used ions are saved.

#864 

###### ShareLink
- [x] Saving ions on imageViewerSnapshot create
- [x] Saving dbIds on imageViewerSnapshot create

###### esConnector
- [x] Added esAnnotationByIon to retrieve snapshot annotations by IonFormulas and dbIds

###### imageViewerSnapshot
- [x] Added ionFormulas field
- [x] Added dbIds field
- [x] Created mutation to add ionFormulas and dbId columns to the db and to populate it  according to its annotationsIds

###### imageViewerSnapshotController
- [x] Retrieve snapshot annotations by IonFormulas and dbIds if present
- [x] Retrieve snapshot annotations by annotationId if IonFormulas and dbIds not present

##### Front end
 
- [x] Unit Test
- [ ] e2e Test
 
- [x] Chrome
- [x]  Safari
- [x] sm, md, lg, xl, lg


LachlanStuart (changes requested):
In general looks good. I've left two comments on where the API & DB specification should be improved. I'm marking as "request changes" because the nullable columns in the model/migration need to be fixed before merging, otherwise it will cause headaches because the model & migrations will be out of sync.

PR #897, opened 2021-07-02
Group dataset editing permissions

### LINKED ISSUE #776: Make dataset editable by group
At the moment, editing the dataset (changing metadata, loading optical images, ...) is available only to the owner of the dataset.
Increasingly, there is a need to be able to edit a dataset by multiple users. Alternatively, the following strategy is proposed:

1. All the same, the dataset will have one owner.
2. The owner can "share" the dataset within his own group (primary?) only.
3. Deleting the dataset, as before, will be possible only by the owner.
4. Also make the PI a co-owner (per #693 )

At the frontend level, an action `Share dataset with group` is required.
At the GraphQL level, we will need:
 - a new endpoint that will accept a request from the frontend and change the status of the dataset (an additional column in the table?)
 - additional logic that will check whether the user has permission to edit the dataset, as belonging to the same groups as the owner of the dataset.

LachlanStuart:
Resolves #776 

## Business logic changes
* When a dataset is in a group, other members of the group may edit the dataset
* When a dataset is in a group, the group admins may delete the dataset
* (Not part of the original task) When a dataset is queued/processing, users may edit metadata / delete the dataset
    * This was actually made (mostly) safe a while ago - sm-api uses a dataset lock while re-indexing a dataset in ElasticSearch, so there should be no race conditions when metadata is edited
    * Submitting for reannotation still causes a race condition (annotation jobs may be processed in parallel and will overlap and cause chaos), so I added a new error message to block this case. It's not 100% fool-proof as it relies on client state, but the worst case issue with a race condition is obvious and recoverable

## Code changes
* I refactored a bunch of the dataset permissions functionality, consolidating most checks into `getDatasetForEditing` (the "source of truth" function), with optimized subsets in `canDeleteEsDataset` and `canEditEsDataset` for cases where ElasticSearch's potentially out-of-date data is acceptable (i.e. the datasets list)
* I added some test functions to allow exhaustively testing every type of relationship between two users

## Screenshots
New error when you try to reprocess a queued/annotating dataset
image


metaspace/graphql/src/modules/dataset/operation/publicationChecks.ts
```diff
@@ -1,23 +1,52 @@
...
+const fetchDatasetProjectsInStatus = async(
+  ctx: Context, datasetId: string, statuses: PublicationStatus[]
+): Promise<DatasetProjectModel[]> => {
+  const dataLoader = ctx.contextCacheGet(
+    'fetchDatasetProjectsInStatusDataLoader',
+    [statuses],
+    (statuses) => new DataLoader(
+      (datasetIds: string[]) => fetchDatasetProjectsInStatusUncached(ctx.entityManager, datasetIds, statuses)
+    ))
+  return await dataLoader.load(datasetId)
```
LachlanStuart, line 39: @lmacielvieira Sorry for the commit spam. This cache was the cause of the bug you mentioned yesterday. Previously it was:
```
const dataLoader = ctx.contextCacheGet('findProjectsByDatasetIdDataLoader', [], () => {
```
That cache key `'findProjectsByDatasetIdDataLoader'` was already used in `ProjectSourceRepository.findProjectsByDatasetId`, which caused the cached results (unfiltered projects for the dataset) to be returned by this function, which was interpreted by `isDatasetInPublicationStatus` as meaning there were projects in the review state. 

Additionally, the cache entry wasn't correctly dependent on the `statuses` argument, which would have caused calls with different `statuses` values to share the same cached results.

I also added a comment up above about why `ProjectSourceRepository.findProjectsByDatasetId` wasn't used here, as it wasn't clear to me either and I ended up mistakenly trying to use it a few times before realizing the problems with it.

lmacielvieira (approved):

PR #898, opened 2021-07-02
Package updates & TS startup time optimizations

LachlanStuart:
I found updating ts-node and ts-node-dev gives a nice performance boost to graphql's startup:

* `time yarn start`: 23.54s -> 10.15s
* `time yarn run dev`: 20.68s -> 10.42s
* `time yarn run test`: 24.63s -> 20.11s

(Averaged over 3 runs, with a line of code added to `server.js` to make it quit after everything finished starting, so that the timings were consistent)

I tried updating ts-jest and jest, but there is currently a [significant performance regression](https://github.com/kulshekhar/ts-jest/issues/2696), so I skipped this.

Hoping the same could happen for webapp, I also updated vue-cli, webpack, babel, etc. but the compilation time didn't really change. I kept the changes anyway, as updates are good.

Lastly, I optimized the CI a bit - skipping type checking on the `gen-binding` and `gen-sql-schema` commands, and allowing CircleCI to cache the `node_modules` directories. Together these save ~1 minute in the generate-graphql-schema step.


lmacielvieira (approved):
Nice optimization :)

Issue #883, opened 2021-06-21
Ds comparison page improvements after first feedback


lmacielvieira:
In order to improve the ds comparison page, some small features should be added.

- [ ] The user should be able to select the datasets to be compared from the dataset list page
- [ ] Add m/z value into the annotation table
- [ ] Add  a popup explaining which MSM is shown in the annotations table
- [ ] Show individual MSM next to each dataset/image, next to its intensity
- [ ] Add a link for each image (annotation within individual dataset) that can be clicked and open the annotation within the selected dataset in a regular view
- [ ] Fix export to CSV

LachlanStuart:
Bug fixes:
- [ ] Fix the default sort order (should be FDR ascending, then MSM descending)
- [ ] Fix the colorization of zero-intensity pixels. On Staging they're showing up as transparent - ion images should only have transparency when there's an optical image.
- [ ] Fix the positioning of the intensity popup when hovering over the image in Firefox

UX changes:
- [ ] Only show a single "cog menu" (intensity scaling, colormap, scale bar color) that affects all images simultaneously
- [ ] Show the name of each dataset above their image so that it's possible to tell them apart if you don't remember how you arranged the grid

I'd expect scientists to want to compare the molecules between datasets in tabular form, so that they don't need to open each annotation and look at all the images one-by-one. However, showing it all at once would be too many columns to display by default, so it'd need to be customizable by the user.
- [ ] Allow user to select a field (MSM, FDR, maxIntensity or totalIntensity) to show in the table with one column per dataset.
- [ ] For dataset-specific fields, show a separate column for each dataset in the CSV (i.e. if you have 5 datasets, then the CSV should have 5 `minIntensity` columns, 5 `maxIntensity` columns, etc.

One suggestion that IIRC came from GM:
- [ ] Ability to filter and/or sort by the number of datasets that include an annotation, so that users can easily see molecules that are shared and/or not shared

Additionally, I'd suggest changing the "Add a popup explaining which MSM is shown in the annotations table" task - IMO the issue isn't that it's unclear which dataset it comes from. The issue is that showing the MSM for the first non-null annotation just isn't very useful, because there's multiple values and instead of summarizing them, one is being arbitrarily picked. I think people would find it more useful to see the best value out of the datasets.
- [ ] Change the MSM/FDR columns to consistently show the best value (highest MSM / lowest FDR) out of the datasets. Also rename them to "Best MSM" / "Best FDR" to reflect this.
[...]

PR #900, opened 2021-07-05
Ds comparison page improvements #1 after first feedback

lmacielvieira:
### Description

In order to improve the ds comparison page, some small features should be added.

Theo
- [x] Add m/z value into the annotation table
- [x] Add  a popup explaining which MSM is shown in the annotations table
- [x] Show individual MSM next to each dataset/image, next to its intensity
- [x] Add a link for each image (annotation within individual dataset) that can be clicked and open the annotation within the selected dataset in a regular view
- [x] Fix export to CSV

Mans
- [x] Only show a single "cog menu" (intensity scaling, colormap, scale bar color) that affects all images simultaneously

Lachlan
 Bug fixes:
- [x] Fix the default sort order (should be FDR ascending, then MSM descending)
- [x] Fix the colorization of zero-intensity pixels. On Staging they're showing up as transparent - ion images should only have transparency when there's an optical image.
- [x] Fix the positioning of the intensity popup when hovering over the image in Firefox

UX changes:
- [x] Only show a single "cog menu" (intensity scaling, colormap, scale bar color) that affects all images simultaneously
- [x] Show the name of each dataset above their image so that it's possible to tell them apart if you don't remember how you arranged the grid

I'd expect scientists to want to compare the molecules between datasets in tabular form, so that they don't need to open each annotation and look at all the images one-by-one. However, showing it all at once would be too many columns to display by default, so it'd need to be customizable by the user.
- [ ] Allow user to select a field (MSM, FDR, maxIntensity or totalIntensity) to show in the table with one column per dataset.
** Added single MSM and FDR valus on each grid annotation image
- [ ] For dataset-specific fields, show a separate column for each dataset in the CSV (i.e. if you have 5 datasets, then the CSV should have 5 `minIntensity` columns, 5 `maxIntensity` columns, etc.
** Print all datasets annotations, where each rows corresponds to a single annotation

One suggestion that IIRC came from GM:
- [x] Ability to filter and/or sort by the number of datasets that include an annotation, so that users can easily see molecules that are shared and/or not shared

Additionally, I'd suggest changing the "Add a popup explaining which MSM is shown in the annotations table" task - IMO the issue isn't that it's unclear which dataset it comes from. The issue is that showing the MSM for the first non-null annotation just isn't very useful, because there's multiple values and instead of summarizing them, one is being arbitrarily picked. I think people would find it more useful to see the best value out of the datasets.
- [x] Change the MSM/FDR columns to consistently show the best value (highest MSM / lowest FDR) out of the datasets. Also rename them to "Best MSM" / "Best FDR" to reflect this.
- [x] Additionally, allow Best MSM/Best FDR to be used for filtering or sorting, because most users will still want to compare images if an annotation has e.g. 5% FDR in one dataset and 50% in another.  (Talk to Veronika if you want more information about this use-case, as she has tried to use METASPACE for dataset comparison in the past, and ran into this inconsistent FDR issue every time)

#883

###### MainImageHeader
- [x] Added props to hide control options

###### DatasetComparisonAnnotationTable
- [x] Added mz value
- [x] Added Best FDR/MSM labels
- [x] Changed default order to FDR ascending
- [x] Exporting each dataset annotation per row to csv

###### DatasetComparisonGrid
- [x] Changed to get ion image settings (colormap, resetviewPort, scalebarColor and resetViewport) from props
- [x] Displays annotation individual MSM value
- [x] Displays annotation individual FDR value
- [x] Adapts ion Image width/height on resize event
- [x] Fixed the colorization of zero-intensity pixels

###### DatasetComparisonPage
- [x] Added global image settings controller

##### Front end
 
- [x] Unit Test
- [ ] e2e Test
 
- [x] Chrome
- [x] Safari
- [x] sm, md, lg, xl, lg


metaspace/webapp/src/modules/Datasets/comparison/DatasetComparisonGrid.tsx
```diff
@@ -234,10 +295,62 @@ export const DatasetComparisonGrid = defineComponent<DatasetComparisonGridProps>
...
+          if (state.gridState[key] && state.gridState[key].colormap
+            && state.gridState[key].colormap !== props.colormap) {
+            diffColormapCount += 1
+          }
+        })
+
+        if (diffColormapCount > 0) {
+          handleGlobalColormapChange()
+        }
+      }
```
LachlanStuart, line 318: This `watchEffect` makes the code much more complex than it needs to be. E.g. it wouldn't need the `if (props.colormap && state.gridState)` and `diffColormapCount` boilerplate logic, and so many layers of `handle***Change` if the `state.gridState[...].ionImageLayers` were set to a `computed` based on `ionImageLayers()` - it would just get recalculated lazily when one of the input parameters changed.

Also, `gridState[...].colormap`, `scaleType` and `colorBarColor` now get their values copied out of `props` - could you please make this component access these fields from `props` directly so that the synchronization code isn't needed? You shouldn't need to pass `onScaleBarColorChange`, `onColormapChange` and `onScaleTypeChange` to `<MainImageHeader` as they'll never be called now that `hideOptions` is used.

metaspace/webapp/src/modules/Datasets/comparison/__snapshots__/DatasetComparisonGrid.spec.ts.snap
```diff
@@ -3,56 +3,46 @@
...
-          <transition-stub mode="out-in" enterclass="opacity-0" leavetoclass="opacity-0" enteractiveclass="duration-150 transition-opacity ease-in-out" leaveactiveclass="duration-150 transition-opacity ease-in-out" class="absolute bottom-0 right-0 mt-3 ml-3 dom-to-image-hidden"></transition-stub>
-          <transition-stub mode="out-in" enterclass="opacity-0" leavetoclass="opacity-0" enteractiveclass="duration-150 transition-opacity ease-in-out" leaveactiveclass="duration-150 transition-opacity ease-in-out" class="absolute top-0 right-0 mt-3 ml-3 dom-to-image-hidden"></transition-stub>
-        </div><button title="Export image" class="button-reset rounded-full bg-gray-100 shadow-xs h-8 w-8 flex items-center justify-center cursor-not-allowed absolute top-0 left-0 mt-3 ml-3 dom-to-image-hidden"><i class="el-icon-download text-xl"></i></button>
-      </div>
+    </div></div> <!----></div></div></mock-el-option> <mock-el-option value="hidden" label="Hidden"></mock-el-option></mock-el-select><transition-stub name="el-zoom-in-top"><!----></transition-stub></div></div></form></span>
+      <!---->
+      </mock-el-popover>
+      <!---->
+      <!---->
```
LachlanStuart, line 16: BTW, if you do `expect(wrapper.element).toMatchSnapshot()` instead of `expect(wrapper).toMatchSnapshot()`, Jest will pretty-print the DOM so that it doesn't stack so many elements on a single line in the snapshots. This makes it a lot more diff-friendly.

The difference is that `expect(wrapper).toMatchSnapshot()` uses the `@vue/test-utils` custom snapshot handler, which just returns the `.innerHTML` directly as a string, whereas `wrapper.element` is a JSDom Element, which Jest can reformat so that each element & attribute goes on its own line.

LachlanStuart (approved):
I like the UI changes, but I think the data flow around `gridState` deserves a cleanup, now that several fields just come from `props` - details in comments. 

If you don't get what I mean about the `computed`, or you have any performance issues with it, LMK. I'm happy to show you show you what I mean, and I think it makes a big difference in the readability of the code.

metaspace/webapp/src/modules/Datasets/comparison/DatasetComparisonPage.tsx
```diff
@@ -140,9 +154,34 @@ export default defineComponent<DatasetComparisonPageProps>({
...
+
+    const handleScaleBarColorChange = (scaleBarColor: string) => {
+      Vue.set(state.globalImageSettings, 'scaleBarColor', scaleBarColor)
+    }
+
+    const handleScaleTypeChange = async(scaleType: string) => {
+      Vue.set(state.globalImageSettings, 'scaleType', scaleType)
+    }
+
+    const handleColormapChange = async(colormap: string) => {
```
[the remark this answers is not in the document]
lmacielvieira: I copied them from dsComparisonGrid, which called a promise inside. It is not needed here, so I will remove. Thx for pointing.

PR #909, opened 2021-07-16
TIC Backend

### LINKED ISSUE #885: TIC image generation (Back-end)
METASPACE currently provides no way to get the TIC of individual spectra. The TIC can be very useful for downstream analysis for a number of reasons:
* Normalizing ion images by the TIC often reduces noise and makes spatial effects clearer
* ML techniques can often use spectrum TICs as a proxy for spectrum quality, density, or limit of detection
* The TIC image can be helpful to diagnose acquisition problems (e.g. someone bumped the instrument and parts of the image were acquired with reduced sensitivity)

TIC normalization is simply dividing each intensity value for each peak by the sum of all intensities in the same spectrum. E.g. if a peak has 1e4 intensity and the total spectrum is 1e5, the TIC-normalized intensity for that peak becomes `1e4 / 1e5 = 0.1` or `10%`.

This task has several parts:

#### Record TIC information while loading ImzML files
- [x] If the information is available in the ImzML file metadata, use that value. It appears like this:
```xml
<cvParam cvRef="MS" accession="MS:1000285" name="total ion current" value="112074.63582611084"/>
```
- [x] If it's not available in the ImzML file metadata, simply sum the values in the intensity array from the ibd file

#### Save the result
- [x] Reuse the [database structure from the shelved Diagnostics branch](https://github.com/metaspace2020/metaspace/commit/d703477283d51f2f246e6f6ba83e6f0adfccafb6#diff-8bbdb7ed07e56df4effabaabdca379120af1618d3aa09e76f22cbb58f97a9a4a), as this will help with integrating with future diagnostics information. 
* This structure was designed so that additional metadata could be supplied along with each image (not needed in this case, but used for many other proposed diagnostics), and so that the images could be easily cleaned up when a dataset is deleted.
* `jobId` should be `NULL` for TIC images, because they do not vary with database/annotations.

- [x] The output should be a 2D image with 1 value per spectrum. 
* This image must align with ion images, so the same coordinate system math should be used
* The format should be a [`.npy`-serialized](https://numpy.org/doc/stable/reference/generated/numpy.save.html?highlight=save#numpy.save) float32 array, where missing pixels are NaN values
    * The reason for using this format is that `.npy` files encode shape & dtype information, and often don't need any additional metadata to decode. There are also several JS libraries that can load `.npy` files, making it easy to use from the front-end.
* Use the existing image service for saving the image. It may need changes for saving binary files, but it's still the best way to manage dataset-related objects in S3, so it should be reused.

#### Migrate existing datasets

- [x] Make a migration script that creates TIC images for existing datasets
* Most datasets are still only on S3, so it would be best if this can also run on S3

#### Add a GraphQL API

- [x] For now it should be enough to just mirror the database structure into sub-resolvers in the `Dataset` query resolver.

#### Integrate with Python Client

- [x] Add an additional function to `SMDataset` to retrieve & load the TIC image. Make sure to pass `ignore_pickle=True` to `np.load` to avoid any security hazards.
- [x] Add a way to automatically normalize images when they're downloaded through `SMDataset.isotope_images` and `all_annotation_images`. It would make sense to expose this through the existing `scale_intensity` parameter, e.g. `scale_intensity='tic_normalize'`

LachlanStuart:
Resolves #885

Mostly went according to plan (as described in #885)

Notes:
* The big increase in sm-engine coverage (69% -> 83% :sunglasses: ) comes from `test_annotation_job`, which covers most of the Lithops pipeline. Previously this pipeline was only covered by sci-test, which isn't included in codecov. 
* I upgraded TypeORM in sm-graphql so that I could specify `createForeignKeyConstraints: false` on several relations, as our DB schema avoids the `public.dataset` and `graphql.dataset` tables. This is a risk - TypeORM releases aren't always stable. I specifically chose version 0.2.31 because the latest version has several unresolved issues around migrations.
* I refactored the way that ImzML files are loaded so that code could be shared between Lithops, Spark, and the `update_diagnostics` migration script. This conflicts with #785 . When work resumes on that PR, I suggest to scrap its ImzML parsing code changes and use the [new `ImzMLParser.polarity` field](https://github.com/alexandrovteam/pyimzML/pull/27) instead.
* I removed the temporary code in sm-graphql that handled non-S3 images during the #773 migration, as it's no longer needed because that migration has finished. I did this while trying to figure out how to convert image IDs to URLs in sm-graphql, but found that it wasn't needed in the end so I just kept the cleanups.

### DIFF
--- a/metaspace/engine/scripts/db_schema.sql
+++ b/metaspace/engine/scripts/db_schema.sql
@@ -192,6 +192,29 @@ CREATE TABLE "public"."job" (
   CONSTRAINT "PK_456bc75d32d741774ab85e6606a" PRIMARY KEY ("id")
 );
 
+CREATE TABLE "public"."dataset_diagnostic" (
+  "id" uuid NOT NULL DEFAULT uuid_generate_v1mc(), 
+  "type" text NOT NULL, 
+  "ds_id" text NOT NULL, 
+  "job_id" integer, 
+  "updated_dt" TIMESTAMP NOT NULL DEFAULT (now() at time zone 'utc'), 
+  "data" json, 
+  "error" text, 
+  "images" json NOT NULL, 
+  CONSTRAINT "PK_2ddb635004b2e08782649e0c638" PRIMARY KEY ("id")
+);
+
+CREATE UNIQUE INDEX "IDX_8064cf8bd1bc774a3d53db893f" ON "public"."dataset_diagnostic" (
+  "ds_id", 
+  "type"
+) WHERE job_id IS NULL;
+
+CREATE UNIQUE INDEX "IDX_2e8c4885755304fd125f7c4815" ON "public"."dataset_diagnostic" (
+  "ds_id", 
+  "type", 
+  "job_id"
+) ;
+
 CREATE TABLE "public"."annotation" (
   "id" SERIAL NOT NULL, 
   "job_id" integer NOT NULL, 
@@ -259,29 +282,6 @@ CREATE TABLE "graphql"."dataset_project" (
   CONSTRAINT "PK_9511b6cda2f4d4299812106cdd4" PRIMARY KEY ("dataset_id", 
   "project_id")
 );
-
-CREATE TABLE "graphql"."dataset_diagnostic" (
-  "id" uuid NOT NULL DEFAULT uuid_generate_v1mc(), 
-  "type" text NOT NULL, 
-  "ds_id" text NOT NULL, 
-  "job_id" integer, 
-  "updated_dt" TIMESTAMP NOT NULL DEFAULT (now() at time zone 'utc'), 
-  "data" json, 
-  "error" text, 
-  "images" text array NOT NULL, 
-  CONSTRAINT "PK_604d02805f10893127e79ef1250" PRIMARY KEY ("id")
-);
-
-CREATE UNIQUE INDEX "IDX_810c0fd2cbcfcb4395f7f120ef" ON "graphql"."dataset_diagnostic" (
-  "ds_id", 
-  "type"
-) WHERE jobId IS NULL;
-
-CREATE UNIQUE INDEX "IDX_f3da139561a67f5656bcf6f786" ON "graphql"."dataset_diagnostic" (
-  "ds_id", 
-  "type", 
-  "job_id"
-) WHERE jobId IS NOT NULL;
 
 CREATE TABLE "graphql"."user" (
   "id" uuid NOT NULL DEFAULT uuid_generate_v1mc(), 
@@ -352,6 +352,14 @@ ALTER TABLE "public"."job" ADD CONSTRAINT "FK_07f17ed55cabe0ef556bc0e0c93" FOREI
   "moldb_id") REFERENCES "public"."molecular_db"("id"
 ) ON DELETE CASCADE ON UPDATE NO ACTION;
 
+ALTER TABLE "public"."dataset_diagnostic" ADD CONSTRAINT "FK_e13bfa279306d04d020139312d1" FOREIGN KEY (
+  "ds_id") REFERENCES "public"."dataset"("id"
+) ON DELETE CASCADE ON UPDATE NO ACTION;
+
+ALTER TABLE "public"."dataset_diagnostic" ADD CONSTRAINT "FK_0216974d86c145e2a08ea291c5d" FOREIGN KEY (
+  "job_id") REFERENCES "public"."job"("id"
+) ON DELETE CASCADE ON UPDATE NO ACTION;
+
 ALTER TABLE "public"."annotation" ADD CONSTRAINT "FK_bfed30991918671d59fc1f5d5e4" FOREIGN KEY (
   "job_id") REFERENCES "public"."job"("id"
 ) ON DELETE CASCADE ON UPDATE NO ACTION;
@@ -383,14 +391,6 @@ ALTER TABLE "graphql"."dataset_project" ADD CONSTRAINT "FK_8bb698a02c945dc2a67a2
 ALTER TABLE "graphql"."dataset_project" ADD CONSTRAINT "FK_e192464449c2ac136fd4f00b439" FOREIGN KEY (
   "project_id") REFERENCES "graphql"."project"("id"
 ) ON DELETE NO ACTION ON UPDATE NO ACTION;
-
-ALTER TABLE "graphql"."dataset_diagnostic" ADD CONSTRAINT "FK_c43b99d980f560e82f711562c73" FOREIGN KEY (
-  "ds_id") REFERENCES "public"."dataset"("id"
-) ON DELETE CASCADE ON UPDATE NO ACTION;
-
-ALTER TABLE "graphql"."dataset_diagnostic" ADD CONSTRAINT "FK_1b73cd1c238b2e78080031618e6" FOREIGN KEY (
-  "job_id") REFERENCES "public"."job"("id"
-) ON DELETE CASCADE ON UPDATE NO ACTION;
 
 ALTER TABLE "graphql"."user" ADD CONSTRAINT "FK_1b5eb1327a74d679537bdc1fa5b" FOREIGN KEY (
   "credentials_id") REFERENCES "graphql"."credentials"("id"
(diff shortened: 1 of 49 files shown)


metaspace/engine/scripts/db_schema.sql
```diff
@@ -80,7 +80,7 @@ CREATE TABLE "graphql"."project" (
   "project_description" text, 
   "review_token" text, 
   "review_token_created_dt" TIMESTAMP, 
-  "publish_notifications_sent" integer NOT NULL DEFAULT 0, 
+  "publish_notifications_sent" integer NOT NULL DEFAULT '0',
```
LachlanStuart, line 83: Many of the changes in this file (including this one) are likely due to the TypeORM upgrade.

metaspace/engine/sm/engine/annotation/job.py
```diff
@@ -38,27 +38,28 @@ def del_jobs(ds: Dataset, moldb_ids: Optional[Iterable[int]] = None):
...
+            'SELECT iso_image_ids '
+            'FROM annotation m '
+            'JOIN job j ON j.id = m.job_id '
+            'JOIN dataset d ON d.id = j.ds_id '
+            'WHERE ds_id = %s AND j.moldb_id = %s',
+            (ds.id, moldb.id),
+        )
+
+        image_ids = [img_id for img_ids in img_id_rows for img_id in img_ids if img_id is not None]
+        image_storage.delete_images(image_storage.ISO, ds.id, image_ids)
```
LachlanStuart, line 58: Ignore whitespace to see the real changes here - `del_diagnostics` was added above, and I added a new function `image_storage.delete_images` so that callers don't need to set up their own `ThreadPoolExecutor`

metaspace/engine/sm/engine/annotation_lithops/annotate.py
```diff
@@ -104,12 +101,12 @@ def finish(self) -> Tuple[pd.DataFrame, pd.DataFrame]:
...
 
 
-def gen_iso_image_sets(sp_inds, sp_mzs, sp_ints, centr_df, nrows, ncols, isocalc_wrapper):
+def gen_iso_image_sets(sp_inds, sp_mzs, sp_ints, centr_df, nrows, ncols, isocalc_wrapper, n_peaks):
     # pylint: disable=too-many-locals
     # assume sp data is sorted by mz order ascending
 
     def yield_buffer(buffer):
-        while len(buffer) < ISOTOPIC_PEAK_N:
+        while len(buffer) < n_peaks:
```
LachlanStuart, line 109: Unrelated bug fix. The hard-coded `ISOTOPIC_PEAK_N` caused a minor bug - when `n_peaks` is less than 4, `annotation.metrics` would still have arrays of length 4. When `n_peaks` was greater than 4, the arrays in `annotation.metrics` would have inconsistent length, depending on the number of peaks that had images.

metaspace/engine/sm/engine/annotation_lithops/executor.py
```diff
@@ -87,12 +87,13 @@ def _save_subtask_perf(
...
     else:
-        exec_times = [-1]
+        # debug_run_locally=True doesn't make futures
+        exec_times = [sum(perf.entries.values()) for perf in subtask_perfs]
     inner_times = subtask_data.pop('inner time', [-1])
     mem_befores = subtask_data.pop('mem before', [-1])
     mem_afters = subtask_data.pop('mem after', [-1])
     perf_data = {
-        'num_actions': len(futures) if futures else -1,
+        'num_actions': len(exec_times),
```
LachlanStuart, line 96: These are bug fixes for the `debug_run_locally` mode of `Executor`. When that flag is set, `futures` is None, so the `num_actions` needs some other way to be calculated. `subtask_perfs` was the easiest option, as it's correctly set in `debug_run_locally` mode.

metaspace/engine/sm/engine/annotation_lithops/segment_centroids.py
```diff
@@ -21,6 +21,8 @@
 )
 from sm.engine.annotation.isocalc_wrapper import IsocalcWrapper
 
+MIN_CENTR_SEGMS = 32
```
LachlanStuart, line 24: I made this a constant so that tests could change it to a lower value. 32 segments causes a lot of console spam

metaspace/engine/sm/engine/annotation_lithops/segment_centroids.py
```diff
@@ -261,7 +263,8 @@ def get_segm_stats(storage, segm_cobject):
 
     warnings = []
 
-    segm_formula_is, stats = fexec.map_unpack(get_segm_stats, db_segms_cobjs, runtime_memory=1024)
+    args = [(cobj,) for cobj in db_segms_cobjs]
+    segm_formula_is, stats = fexec.map_unpack(get_segm_stats, args, runtime_memory=1024)
```
LachlanStuart, line 267: With `debug_run_locally=True`, `map` doesn't correctly handle passing e.g. `[1,2,3]` as arguments for single-argument functions. Lithops also deprecated this pattern, preferring `[(1,), (2,), (3,)]`, so I fixed the callers instead of changing `Executor.map` to support the deprecated style.

metaspace/engine/sm/engine/annotation_spark/annotation_job.py
```diff
@@ -30,11 +30,6 @@
 
 logger = logging.getLogger('engine')
 
-JOB_ID_MOLDB_ID_SEL = "SELECT id, moldb_id FROM job WHERE ds_id = %s AND status='FINISHED'"
-TARGET_DECOY_ADD_DEL = (
-    'DELETE FROM target_decoy_add tda WHERE tda.job_id IN (SELECT id FROM job WHERE ds_id = %s)'
-)
```
LachlanStuart, line 36: These were unused

metaspace/engine/sm/engine/util.py
```diff
@@ -32,7 +32,7 @@ def split_cos_path(path):
         tuple[string, string]
     Returns a pair of (bucket, key)
     """
-    return re.sub(r'^cos?://', '', path).split(sep='/', maxsplit=1)
+    return re.sub(r'^cos://', '', path).split(sep='/', maxsplit=1)
```
LachlanStuart, line 35: Unrelated fix - I think `cos?` was a copy+paste mistake, because the function above uses `s3a?://` to match both `s3://` and `s3a://`. `co://` is not a valid protocol, so I removed the `?`

metaspace/engine/tests/conftest.py
```diff
@@ -26,14 +31,24 @@
...
+    SMConfig._config_dict['elasticsearch']['index'] = test_id
+    SMConfig._config_dict['rabbitmq']['prefix'] = f'test_{worker_id}__'
+
+    for path in SMConfig._config_dict['lithops']['sm_storage'].values():
         # prefix keys with test ID so they can be cleaned up later
         path[1] = f'{test_id}/{path[1]}'
 
-    return sm_config
+    # __LITHOPS_SESSION_ID determines the prefix to use for anonymous cloudobjects
+    os.environ['__LITHOPS_SESSION_ID'] = f'{test_id}/cloudobjects'
```
LachlanStuart, line 49: These Lithops lines are new. The rest of the `sm_config()` changes are from [the parallelize tests PR](https://github.com/metaspace2020/metaspace/pull/905/files#diff-dc498e079dd80f381f636c0d4f7293b768cbf1df68e1aca4ad47632bc5c29f69R29-R44).

metaspace/graphql/config/default.js
```diff
@@ -28,26 +28,11 @@ module.exports = {
...
       raw_optical_image: {
         type: 'image/jpeg',
         path: '/raw_optical_images/',
         storage_types: ['fs'],
       },
-      ion_thumbnail: {
-        type: 'image/png',
-        path: '/ion_thumbnails',
-        storage_types: ['fs'],
-      },
```
LachlanStuart, line 50: This is cleaning up after #773 's migration. `iso_image`, `optical_image` and `ion_thumbnail` categories are no longer needed, but `raw_optical_image` is still used.

sergii-mamedov (approved):
The PR is a bit big to understand, but in general I have no comments.

PR #923, opened 2021-08-11
Feat/imzml browser

lmacielvieira:
### Description

A new feature which makes possible to browse the dataset's imzml file, by making possible to visualize the ion Image and the selected pixel spectrum chart.
#582 

### How to test it

As the imzml browser backend routes are a POC, in order to test the front-end, you need to run the backend as a standalone server. To test it, follow these steps:

1 - Run the metaspace project
2 - Go to the the folder metaspace/metaspace/browser and register the sm dev package:  python3 setup.py develop
3 - Go to the folder /metaspace/metaspace/browser/sm/browser and run the server: uvicorn api:app --reload 
4 - Select the inputPath of the dataset you want to check the functionality, add its value to the attribute s3_path and pre-process it (via postman):
     route: http://127.0.0.1:8000/preprocess
     body: {
             "s3_path": "s3://sm-engine-dev/ddd7f0e5-ff74-4f61-96d4-60f40dfb4c3c",
     }
     * This step is needed, because I still didn't create a back-end route to check if the temp files generated already exists (i didnt do because I dont know how this will be in the production env).
5 - After pre-process the dataset, you can access the imzml browser page on the front-end: http://localhost:8999/dataset/<dataset_id>/browser    

###### DatasetBrowserPage
- [x] Added spectrum chart filters
> - [x] Filter by database
> - [x] Filter by only annotated by FDR
- [x] Added image filters
> - [x] Filter by m/z
> - [x] Filter by formula
- [x] Querying ion image by s3_path, mz and ppm
- [x] Querying spectrum chart by selected x,y pixel
To do:
- [ ] Check if imzml temp files exists before doing the search query
- [ ] Add TIC normalization
- [ ] Handle back-end errors
- [ ] Normalize ion image with max and min mz returned from backend

###### DatasetBrowserSpectrumChart
- [x] Showing instructions to visualize the chart when empty
- [x] Showing spectrum chart
- [x] Added spectrum chart image download feature
- [x] Added spectrum y axis zoom by slider
- [x] Added spectrum x axis zoom by mouse scroll
- [x] Added chart zoom full reset by double clicking the chart
- [x] Added chart zoom full reset by clicking the restore button
- [x] Added chart zoom last step reset by clicking the reset button
- [x] Added chart zoom with brushing tool by clicking the zoom button

###### Dataset api
- [x] Returning input_path in datasetDetail fragment

###### Ion image viewer
- [x] Added keep pixel selected logic and callback
To do:
- [ ] Move selected pixel by pressing keyboard arrows

###### Simple ion image viewer
- [x] Created simple wrapper to ion image with embed intensity slider handler

###### Api
- [x] Enabled CORS for testing on localhost
- [x] Added search_pixel route to get spectrum data by passing pixel coordinates

###### Main
- [x] Added search_pixel by spectrum function

###### utils
- [x] Enabled minio usage

##### Front end
 
- [ ] Unit Test
- [ ] e2e Test
 
- [x] Chrome
- [ ] Safari
- [ ] Firefox
- [x] md, lg, lg
- [ ] sm, xl

#### Considerations
* A warning from echarts is displayed on the console ("import echarts from 'echarts/lib/echarts'" is not supported anymore. Use "import * as echarts from 'echarts/lib/echarts'" instead;) as the vue-echarts wrapper is importing it on the old way. But according to the code this warning is not shown on production mode. I can downgrade echarts version so this warning is not shown, but when downgraded the label overlapping function does not work.
* Apparently, by using onResult feature from useQuery, even when setting the localVue, jest does not seem to be able to parse the file, so thats why i removed the DatasetBrowserPage tests for now.


metaspace/webapp/src/modules/Datasets/imzml/SimpleIonImageViewer.tsx
```diff
@@ -0,0 +1,400 @@
...
+                    class="range-slider p-3 bg-gray-100 rounded-lg box-border shadow-xs">
+                    {
+                      state.imageSettings.userScaling
+                      && state.rangeSliderStyle
+                      && <RangeSlider
+                        class="ds-comparison-opacity-item"
+                        value={state.imageSettings.userScaling}
+                        min={0}
+                        max={1}
+                        step={0.01}
```
LachlanStuart, line 366: I would suggest a step of `0.1`. The UI makes it easy to just click and hold these buttons, and if a user did that trying to go from 3.0 to 2.0, it would send 100 requests. Under real-world network conditions this would cause the requests to queue up and cause a lag spike for them.

It may be worth also adding a [throttle](https://lodash.com/docs/4.17.15#throttle) to `handleUserScalingChange` so that the user can't queue up lots of requests by holding the button down.

metaspace/browser/sm/browser/api.py
```diff
@@ -29,6 +39,11 @@ async def preprocess(item: DatasetPreprocessItem):
...
 
 
+class PixelSearchItem(BaseModel):
+    s3_path: str
+    x: int
+    y: int
+
 class MzSearchItem(BaseModel):
     s3_path: str
     mz: float
```
LachlanStuart, line 49: Not your code, but one line down the `ppm` is an `int` but should be a `float`, and it's quite obvious there's a problem when playing with the UI. FastAPI coalesces values into their annotated type, so when you switch from 3.00ppm to 2.99ppm in the UI the image suddenly jumps, because it's actually snapping to 2ppm due to the integer conversion.

metaspace/webapp/src/lib/formulaParser.tsx
```diff
@@ -0,0 +1,95 @@
...
+      const mass = periodicTable[elementKey][2][0]
+      mz += nOfElements * mass
+    }
+  })
+
+  return mz
+}
+
+export const generateIonFormula = (molecularFormula: string) => {
+  const cleanFormula = molecularFormula.toUpperCase().trim().replace(/\s/g, '')
```
LachlanStuart, line 63: This `.toUpperCase()` also needs to be removed to fix the case insensitivity problem

metaspace/webapp/src/lib/periodicTable.tsx
```diff
@@ -0,0 +1,134 @@
+export const periodicTable : any = {
+  H: [1, 1, [1.007825032, 2.014101778], [0.999855, 0.000145]],
+  HE: [2, 0, [3.01602932, 4.002603254], [0.000002, 0.999998]],
```
LachlanStuart, line 3: These elements should all only have their first letter capitalized, e.g. `He` for helium.

metaspace/browser/sm/browser/main.py
```diff
@@ -72,12 +76,30 @@ def search(self, mz_lo: int, mz_hi: int) -> np.ndarray:
...
+        mz_image, alpha, mz_max, mz_min = mz_search.create_mz_image(mz_peaks, self.coordinates)
+        rgba_image = plt.get_cmap("gray")(mz_image)
         rgba_image[:, :, 3] = alpha
         log(start, "done")
         return rgba_image
 
+    def search_pixel(self, x: int, y: int) -> np.ndarray:
+        start = time.time()
+        log(start, f"pixel parsing imzml at {self.imzml_path}")
+        p = ImzMLParser(self.imzml_path)
```
LachlanStuart, line 88: FYI (not part of your task job, but in case you're interested), this line is the reason this function is slow - it parses the .ImzML file, and XML parsing is horribly slow.

The solution is to save a pickled version of `imzml_reader.portable_spectrum_reader()` (a lightweight version of the class built so that it could be easily pickled) during the `preprocess` step, and load that here instead.

metaspace/webapp/src/lib/formulaParser.tsx
```diff
@@ -0,0 +1,95 @@
...
+        ionElements[elem] += 1
+      } else {
+        ionElements[elem] = 1
+      }
+    } else if (adduct.indexOf('-') !== -1) {
+      if (Object.keys(ionElements).includes(elem)) {
+        ionElements[elem] -= 1
+      }
+    }
+  })
```
LachlanStuart, line 92: This doesn't seem to support adducts with more than one element, or with element counts (e.g. `+NH4` or `-H2O`)

LachlanStuart (commented):
Looking good so far. Sorry, I've run out of time before going on holiday, so this review isn't complete. Here's my feedback so far:

* When in "Show peaks annotated at ... FDR" mode, I still expected the intensities to change when moving between pixels, but they didn't. It looks like you're using the annotation's `totalIntensity` or something like that - it would IMO make more sense to use the intensity from the spectrum for the selected pixel.
* The spectrum zoom resets when changing pixels, which is quite disruptive when you're only interested in a small range of the spectrum
* Switching from "All peaks" to "Show peaks annotated at ..." resets the m/z of the image view, which is disruptive
* I couldn't get the formula entry to work at all... even with the example H2O+H There's something weird about that text box - I can't paste text into it. Also, when a formula is entered, the m/z value above should update to that formula's m/z.

lmacielvieira:
> * When in "Show peaks annotated at ... FDR" mode, I still expected the intensities to change when moving between pixels, but they didn't. It looks like you're using the annotation's `totalIntensity` or something like that - it would IMO make more sense to use the intensity from the spectrum for the selected pixel.

I couldn't reproduce this behaviour. Both "All annotations" and "Show peaks annotated at ... FDR" modes seems to be changing as I change the selected pixel

metaspace/webapp/src/lib/formulaParser.tsx
```diff
@@ -0,0 +1,95 @@
+import { periodicTable } from './periodicTable'
+
+export const parseFormula = (formula: string) => {
+  const regexp = /(?<element>[A-Z][a-z]{0,2})(?<n>[0-9]*)/ig
```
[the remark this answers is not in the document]
lmacielvieira: I did this way expecting the the user could type all upper or lower case characters to make it easier (i,e CLNAH), but it really introduces a lot of strange behaviours. So I fixed as you suggested. Thx

metaspace/webapp/src/lib/formulaParser.tsx
lmacielvieira: I changed to support it now. thx

metaspace/webapp/src/modules/Datasets/imzml/DatasetBrowserPage.tsx
```diff
@@ -0,0 +1,482 @@
...
+                    state.invalidFormula = false
+                  }
+                  state.moleculeFilter = value
+                }}
+                onChange={() => {
+                  const { moleculeFilter } : any = state
+                  if (!state.invalidFormula) {
+                    const newMz = calculateMzFromFormula(moleculeFilter as string)
+                    state.mzmScoreFilter = newMz
+                    requestIonImage(newMz)
```
LachlanStuart, line 407: This calculates the formula's neutral mass, but should actually calculate the charged mass. The difference is +/- one electron mass. Charge can be determined from the dataset's polarity:

* For positive-mode datasets, subtract 1 electron mass
* For negative-mode datasets, add 1 electron mass

(Yes, this looks like it's the wrong way around. It's because electrons are negatively charged)

The electron mass value is the number in `periodicTable.Ee[2][0]`

metaspace/webapp/src/modules/Datasets/imzml/SimpleIonImageViewer.tsx
```diff
@@ -0,0 +1,400 @@
...
+    }
+
+    return () => {
+      return (
+        <div class={'simple-ion-image-container'}>
+          {
+            state.imageSettings
+            && <MainImageHeader
+              class='simple-ion-image-item-header dom-to-image-hidden'
+              annotation={state.annotation}
```
LachlanStuart, line 319: `state.annotation` never seems to be set anywhere. Did you mean `annotation.value`?

metaspace/webapp/src/modules/Datasets/imzml/SimpleIonImageViewer.tsx
```diff
@@ -0,0 +1,400 @@
...
+              slot="title"
```
LachlanStuart, line 320: `slot` is meaningless here as the parent is just a `div`. I'm surprised that Vue didn't complain about invalid usage of its special props.

LachlanStuart (approved):
Wow, this actually feels really magical to play with, now that everything is working for me. I've left some comments where I think the code could be improved, but you're free to merge once you've addressed them.

Issue #796, opened 2021-03-22
Implement draft version of recalibration


LachlanStuart:
- [ ] Copy across current recalibration & alignment code
- [ ] Either try to improve & package MSIWarp for integration, or copy across the necessary parts
- [ ] Save recal stats, sum spectrum, TIC image to DB as Diagnostic blobs
- [ ] Add feature-flagged UI

LachlanStuart:
This has been shelved as we found it didn't provide a compelling argument for adoption in its current state. (~15% of good-quality and ~100% of bad-quality datasets got worse results, overall improvement usually less than [Raphaël La Rocca's implementation](https://github.com/LaRoccaRaphael/MSI_recalibration)). We decided that it should be re-evaluated with METASPACE ML FDR (#797) once that's available, as the inclusion of mass error in the MSM score is likely to provide a clearer advantage.

My notes, if/when this gets picked up again:
* The most important outstanding shortcoming is explainability. The end-to-end diagnostics in `evaluation.py` need to be re-tooled to plot mass errors (& stddevs) on a ppm scale, and track peaks through the pipeline by peak index instead of m/z so that a reliable before & after view can be shown.
* Unanswered question: Is it possible to detect and abort if recalibration seems to worsen the results?
* There's a bug that causes excessive memory usage somewhere - seems to be late in the pipeline, possibly while writing the ImzML file.
* Several as-yet undocumented ideas for how to further improve the algorithm:
    * Find recalibration targets based on distinct spectra instead of sum spectra. The spectrum summation step is the most likely culprit for why many known peaks didn't work at all for recalibration in the spotting datasets.
    * Likely reason for underperformance compared to Raphaël's algorithm: that algorithm pre-filters outliers, so RANSAC has less work to do.

PR #924, opened 2021-08-12
Recalibration tool

LachlanStuart:
This adds the recalibration command-line/library that I've been working on in #796 . That task has been shelved (reasons given in a comment on that issue). This project has already been [made public](https://pypi.org/project/msi-recal/) though not officially announced, and not integrated into METASPACE in any way.

The PyPI package has an awkward build script, because it vendorizes MSIWarp, which requires C++ code to be compiled. I did this so that all major OSs and Python versions could be supported - MSIWarp only has published packages for Win/macOS with Python 3.8. This PR includes a GitHub action for packaging & publishing to PyPI that handles all the OS/Python version combinations.

This PR also includes some molecular database files, as they are bundled into the PyPI package. The `matrix_` files were manually transcribed from [this paper](https://analyticalsciencejournals.onlinelibrary.wiley.com/doi/10.1002/rcm.9110). HMDB/CM3/LIPID_MAPS were just copied from the files that Ansible uses to install these DBs. Technically these could be downloaded during package compilation to avoid having them in Git, I was just lazy. It's up to you whether I should do this.

It's up to you how much reviewing this needs. As this code isn't used by METASPACE and isn't publicly advertised, I feel it's low-risk because if there were any bugs, they wouldn't have much impact.


.github/workflows/msi_recal_publish.yml
```diff
@@ -1,13 +1,68 @@
...
+        path: metaspace/recal/MSIWarp
+        submodules: true
+
+    - uses: actions/setup-python@v2
+
+    - name: Build wheels
+      uses: joerick/cibuildwheel@v1.10.0
+      env:
+        CIBW_BUILD: "*_amd64 *_x86_64"
+        CIBW_SKIP: "cp27* cp35* pp*"
```
sergii-mamedov, line 30: do we really want to support 2.7?

metaspace/recal/msi_recal/evaluate.py
```diff
@@ -0,0 +1,126 @@
...
+    #             for set_name, peaks_dict in self.collected_peak_sets.items():
+    #                 peaks = peaks_dict.get(mol_idx)
+    #                 if peaks is not None:
+    #                     peaks = peaks.loc[idxs]
+    #                     if len(peaks): WIP
+    #
+    #
+    #                 if mol_idx in peaks_dict:
+    #
+    #                 peaks = peaks_dict[mol_idx]
```
sergii-mamedov, line 116: perhaps it should be removed if it is no longer used?

sergii-mamedov (approved):
LGTM

.github/workflows/msi_recal_publish.yml
LachlanStuart: `CIBW_SKIP` is the list of versions to exclude. Only [CPython 3.6-3.9](https://pypi.org/project/msi-recal/#files) are built with this combination of options.

PR #926, opened 2021-08-13
Optimize Python client upload

LachlanStuart:
I got bored while waiting for a batch of datasets to upload through Python Client, so I optimized it:
* Upload 8 chunks in parallel
* Use a `Session` so that HTTPS connections are reused between chunks
* Retry failing requests. Some team members have had occasional timeout issues (~1 per 2000-3000 requests) with parallel downloads on their home internet - parallel uploads are likely to be much worse, so this should hopefully prevent some headaches.
* Support files >50GB by increasing the chunk size
* Report the total number of upload parts in the progress messages

### DIFF
--- a/metaspace/python-client/metaspace/__init__.py
+++ b/metaspace/python-client/metaspace/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '1.8.9'
+__version__ = '1.8.10'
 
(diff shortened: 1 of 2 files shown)


sergii-mamedov (approved):
LGTM

Issue #910, opened 2021-07-16
[DatasetComparison] Simplify ionImage rendering logic


lmacielvieira:
In order to simplify the datasetComparisonGrid readability, the code must be improved to use a lazy ionImage update logic instead of the hard called update and the watcheffect

- [ ] Fix intensity lock system
- [ ] Lock all grid images with same intensity once one is locked
- [ ] Remove filters from dataset name query (leave only the id one)

PR #927, opened 2021-08-13
Feat/ds comparison improvements

lmacielvieira:
### Description

In order to simplify the datasetComparisonGrid readability, the code must be improved to use a lazy ionImage update logic instead of the hard called update and the watcheffect.
 Fix intensity lock system
 Lock all grid images with same intensity once one is locked
 Remove filters from dataset name query (leave only the id one)

#910 

###### DatasetComparisonGrid
- [x] Show empty cell if selected annotation index is greater than the number of annotations
- [x] Fixes intensity lock system
- [x] Persist intensity lock on row change
- [x] Applies intensity lock to all grid cells
- [x] Changed watcheffect ionImage build system to individual watch effects bases on annotation selection change
- [x] Fixed opacityMode update on toggle optical Image
- [x] Added text overflow with ellipsis  when dataset name too big

###### DatasetComparisonPage
- [x] Displaying selected formula and m/z on top of image grid
- [x] Waiting for datasetIds response to trigger datasetDetail query
- [x] Rendering no data in compounds container if compounds not found

###### EditIntensity
- [x] Stops onKey propagation (so the table do not change pages if moving arrows to edit intensity)

##### Front end
 
- [x] Unit Test
- [ ] e2e Test
 
- [x] Chrome
- [ ] Safari
- [ ] Firefox
- [x] md, lg, lg
- [ ] sm, xl

#### Considerations
* I still wasn't able to remove the handleImageLayerUpdate calls to update the ionImages and update it lazily with computed values. I'm still trying to figure out why the IonImageViewer does not listen to the updates but its wrapper like the MainImage does. My plan is to use the SimpleIonImageViewer created on the browser branch also on the DatasetComparisonGrid, as it is listening to the props update and it can have separated state handlers (not like the ones used on the ImageViewer).

image

### DIFF
--- a/metaspace/webapp/src/components/IonImageViewer.tsx
+++ b/metaspace/webapp/src/components/IonImageViewer.tsx
@@ -412,6 +412,7 @@ const useImageSize = (props: Props) => {
 }
 
 export default defineComponent<Props>({
+  name: 'IonImageViewer',
   props: {
     ionImageLayers: Array,
     isLoading: { type: Boolean, default: false },
--- a/metaspace/webapp/src/lib/ionImageRendering.ts
+++ b/metaspace/webapp/src/lib/ionImageRendering.ts
@@ -211,7 +211,7 @@ const quantizeScaleBar = (minIntensity: number, maxIntensity: number,
   return quantizeIonImage(linearDistribution, minIntensity, maxIntensity, rankValues, scaleMode)
 }
 
-export const loadPngFromUrl = async(url: string) => {
+export const loadPngFromUrl = async(url: string): Promise<Image> => {
   const response = await fetch(url, { credentials: 'omit' })
   if (response.status !== 200) {
     throw Object.assign(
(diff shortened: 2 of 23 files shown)


LachlanStuart (changes requested):
This code is still more complex than it needs to be, and I found several bugs.

Instead of adding complaints against every place where I think there's something wrong, I made a PR showing how I recommend this code be written: https://github.com/metaspace2020/metaspace/pull/932

While I hope you take all my suggested changes, these are the ones I feel are most important:
* Removing the unnecessary `watch` functions and duplicated state for `props.colormap`, `props.scaleType`, and `props.scaleBarColor`
* Replacing all the cases of `Vue.set(state.gridState, key, { ...state.gridState[key], changedField: newVal })` with `state.gridState[key].changedField = newVal`
* Removing the unnecessary uses of `async`/`await`, including removing the call to `loadPngFromUrl` from `ionImage` and the call to `ionImage` from `imageFit`
* Fixing `pixelAspectRatio` which is currently hard-coded to `1`, and the default `zoom: 0.7` which seems to be used because either because the `pixelAspectRatio` is wrong, or the `dimensions`' width and height are swapped.
* Fixing the unhandled promise rejection messages and "missing required props" errors logged to console during tests
(some of these aren't from changes in this PR, but I'm including them as I take it this is the cleanup PR for the original one that was merged with known issues to get faster user feedback)

PR #932, opened 2021-08-25
Suggested changes to the DatasetComparisonGrid

LachlanStuart:
As I've previously mentioned, the code in these components is more complicated than it needs to be, and a lot of it is from fighting against Vue's reactivity paradigm and instead trying to push state changes imperatively. This PR shows how to reduce complexity and code quantity by removing duplicated state and using `computed`s instead of `watch`es.

### DIFF
--- a/metaspace/webapp/src/modules/Datasets/comparison/DatasetComparisonGrid.tsx
+++ b/metaspace/webapp/src/modules/Datasets/comparison/DatasetComparisonGrid.tsx
@@ -420,7 +420,9 @@ export const DatasetComparisonGrid = defineComponent<DatasetComparisonGridProps>
       if (event) {
         event.stopPropagation()
       }
-      state.gridState[key]!.imagePosition = defaultImagePosition()
+      if (state.gridState[key]) {
+        state.gridState[key]!.imagePosition = defaultImagePosition()
+      }
     }
 
     const toggleOpticalImage = (event: any, key: string) => {


metaspace/webapp/src/components/IonImageViewer.tsx
```diff
@@ -412,6 +412,7 @@ const useImageSize = (props: Props) => {
 }
 
 export default defineComponent<Props>({
+  name: 'IonImageViewer',
```
LachlanStuart, line 415: Not part of your changes, I just noticed Vue DevTools was breaking because this component didn't have a name.

metaspace/webapp/src/modules/Annotations/AnnotationView.ts
```diff
@@ -33,31 +33,11 @@ import LockSvg from '../../assets/inline/refactoring-ui/icon-lock.svg'
...
-   imagePosition: ImagePosition
-   opticalSrc: string | null
-   opticalTransform: number[][] | null
-   pixelAspectRatio: number
-   opticalOpacity: number
-   // scaleType is deliberately not included here, because every time it changes some slow computation occurs,
-   // and the computed getters were being triggered by any part of the ImageSettings object changing, such as opacity,
-   // causing a lot of jank.
-   // scaleType?: ScaleType
- }
```
LachlanStuart, line 59: These were moved so they could be made public and used in `DatasetComparisonGrid`

metaspace/webapp/src/modules/Datasets/comparison/DatasetComparisonGrid.spec.ts
```diff
@@ -6,7 +6,90 @@ import { sync } from 'vuex-router-sync'
 import store from '../../../store'
 import router from '../../../router'
 
-describe('DatasetComparisonAnnotationTable', () => {
+// Mock loadPngFromUrl to prevent network requests, but keep the rest of ionImageRendering as it's actually used
+jest.mock('../../../lib/ionImageRendering', () => Object.assign(
+  {},
+  jest.requireActual('../../../lib/ionImageRendering'),
+  { loadPngFromUrl: jest.fn() },
+))
```
LachlanStuart, line 14: This fixes the uncaught promise rejection errors that appear while tests are running

metaspace/webapp/src/modules/Datasets/comparison/DatasetComparisonGrid.tsx
```diff
@@ -163,34 +176,33 @@ export const DatasetComparisonGrid = defineComponent<DatasetComparisonGridProps>
...
+      const { width = dimensions.width, height = dimensions.height } = gridCell?.ionImagePng || {}
 
       return fitImageToArea({
         imageWidth: width,
-        imageHeight: height / (state.gridState[key]?.pixelAspectRatio || 1),
-        areaWidth: dimensions.height,
-        areaHeight: dimensions.width,
+        imageHeight: height / (gridCell?.pixelAspectRatio || 1),
+        areaWidth: dimensions.width,
+        areaHeight: dimensions.height,
```
LachlanStuart, line 187: Note: WebStorm caught that these variables were mismatched. I highly recommend using it if you're not already.

metaspace/webapp/src/modules/Datasets/comparison/DatasetComparisonGrid.tsx
```diff
@@ -163,34 +176,33 @@ export const DatasetComparisonGrid = defineComponent<DatasetComparisonGridProps>
       return Object.assign(safeJsonParse(annotation.dataset.metadataJson), datasetMetadataExternals)
     }
 
-    const imageFit = async(annotation: any, key: string, pixelAspectRatio: number = 1) => {
-      const finalImage = await ionImage(state.gridState[key]?.ionImagePng, annotation.isotopeImages[0])
-      const { width = dimensions.width, height = dimensions.height } = finalImage || {}
+    const imageFit = (key: string) => {
+      const gridCell = state.gridState[key]
+      const { width = dimensions.width, height = dimensions.height } = gridCell?.ionImagePng || {}
```
LachlanStuart, line 181: Removing the `await ionImage(...)` fixes the bug I found where images are loaded from the network multiple times. The image data was already available and loaded in `gridCell.ionImagePng`... Even if this PR isn't merged, I suggest you include this fix.

metaspace/webapp/src/modules/Datasets/comparison/DatasetComparisonGrid.tsx
```diff
@@ -204,111 +216,88 @@ export const DatasetComparisonGrid = defineComponent<DatasetComparisonGridProps>
...
       })
 
       Promise.all(settingPromises)
-        .then((values) => {
-          // pass
-        })
-        .catch((e) => {
-          // pass
-        })
+        .catch(console.error)
```
LachlanStuart, line 300: The empty `.catch((e) => {})` was causing exceptions to be completely hidden. The test for this component was actually broken. I suggest using `.catch(console.error)` instead for these cases.

metaspace/webapp/src/modules/Datasets/comparison/DatasetComparisonGrid.tsx
```diff
@@ -321,55 +310,28 @@ export const DatasetComparisonGrid = defineComponent<DatasetComparisonGridProps>
...
+          resetViewPort(null, key)
+        })
       }
     })
 
-    const defaultImagePosition = {
-      zoom: 0.7,
+    const defaultImagePosition = () => ({
+      // This is a function so that it always makes a separate copy for each image
+      zoom: 1,
```
LachlanStuart, line 331: I made several fixes so that `pixelAspectRatio` is correctly calculated and passed to `<IonImageViewer>`. With those fixes, `zoom` can now be set to its correct default value of `1`

metaspace/webapp/src/modules/Datasets/comparison/DatasetComparisonGrid.tsx
```diff
@@ -415,63 +377,60 @@ export const DatasetComparisonGrid = defineComponent<DatasetComparisonGridProps>
           ),
         }
       }
-      return null
+      return {
+        min: getIntensityData(0, 0, 0, 0, 0, false),
+        max: getIntensityData(0, 0, 0, 0, 0, false),
+      }
```
LachlanStuart, line 383: I'm unsure if this is the correct way to fix it, but when this was returning `null`, `<IonIntensity` would have errors.

metaspace/webapp/src/modules/Datasets/comparison/DatasetComparisonGrid.tsx
```diff
@@ -204,111 +216,88 @@ export const DatasetComparisonGrid = defineComponent<DatasetComparisonGridProps>
...
+          pixelAspectRatio: pixelSizeX && pixelSizeY && pixelSizeX / pixelSizeY,
+          imageZoom: 1,
+          showOpticalImage: true,
+          userScaling: [0, 1],
+          imageScaledScaling: [0, 1],
+          scaleBarUrl: computed(() => renderScaleBar(
+            ionImageLayersAux.value[0]?.ionImage,
+            createColormap(props.colormap),
+            true,
+          )),
```
LachlanStuart, line 260: Optimization: cache the scale bar in a `computed` so that it only gets re-rendered when needed. Otherwise it was getting re-rendered for every image for every render, which caused a lot of lag.

metaspace/webapp/src/modules/Datasets/comparison/DatasetComparisonGrid.tsx
```diff
@@ -204,111 +216,88 @@ export const DatasetComparisonGrid = defineComponent<DatasetComparisonGridProps>
...
-      await handleImageLayerUpdate(state.annotationData[key], key)
+        const pixelSizeY = metadata?.MS_Analysis?.Pixel_Size?.Yaxis || 0
+
+        gridCell = reactive({
+          intensity,
+          ionImagePng,
+          pixelSizeX,
+          pixelSizeY,
+          ionImageLayers: ionImageLayersAux,
+          imageFit: computed(() => imageFit(key)),
```
LachlanStuart, line 247: Note that because `gridCell` is a `reactive`, this `computed` gets unwrapped. It's possible to directly access `gridCell.imageFit.zoom` instead of needing `gridCell.imageFit.value.zoom`.

metaspace/webapp/src/modules/Datasets/comparison/DatasetComparisonGrid.tsx
```diff
@@ -481,16 +440,13 @@ export const DatasetComparisonGrid = defineComponent<DatasetComparisonGridProps>
...
-    const handleImageMove = ({ zoom, xOffset, yOffset }: any, imageFit: any, key: string) => {
-      Vue.set(state.gridState, key, {
-        ...state.gridState[key],
-        imagePosition: {
-          ...state.gridState[key].imagePosition,
-          zoom: zoom / imageFit.imageZoom,
-          xOffset,
-          yOffset,
-        },
-      })
```
LachlanStuart, line 493: I know this pattern is common in React with `setState`, but it's actually really bad to do this in Vue. Recreating the full object like this causes every `computed` that accesses that object to be recomputed. If you only modify the fields that have changed, then only the `computed`s that depend on the changed fields will update.

metaspace/webapp/src/modules/Datasets/comparison/DatasetComparisonGrid.tsx
```diff
@@ -649,52 +570,29 @@ export const DatasetComparisonGrid = defineComponent<DatasetComparisonGridProps>
...
-        handleScaleBarColorChange(scaleBarColor, key)
+    const handleIonIntensityLockChangeForAll = (value: number, type: string) => {
+      // apply max lock to all grids
+      Object.keys(state.gridState).forEach((gridKey) => {
+        handleIonIntensityLockChange(value, gridKey, type)
+        if (value) {
+          handleIonIntensityChange(value, gridKey, type, true)
+        }
       })
     }
```
LachlanStuart, line 594: I split this into a separate method because `applyToAll` dramatically changed the behavior of the function and made it unnecessarily recursive. 

Since `applyToAll` is always supplied as a constant value by the caller, it's cleaner to just put that logic into separate function,

metaspace/webapp/src/modules/ImageViewer/IonIntensity.vue
```diff
@@ -83,14 +83,6 @@ export default defineComponent<Props>({
...
 
     const status = computed(() => props.intensities.status)
-    const display = computed(() => {
-      if (status.value === 'LOCKED') {
-        return props.intensities.user
-      } else if (status.value === 'CLIPPED') {
-        return props.intensities.clipped
-      }
-      return props.value
-    })
```
LachlanStuart, line 93: This code was unused

metaspace/webapp/src/modules/Datasets/comparison/DatasetComparisonGrid.tsx
```diff
@@ -204,111 +216,88 @@ export const DatasetComparisonGrid = defineComponent<DatasetComparisonGridProps>
...
         intensity.max.scaled = intensity.max.status === 'CLIPPED'
           ? intensity.max.clipped : intensity.max.image
       }
-
-      const settings = {
-        intensity,
-        ionImagePng,
-        metadata,
+      let gridCell: GridCellState
+      if (hasPreviousSettings) {
```
LachlanStuart, line 230: BTW, look how much cleaner this code is if you can just guarantee that `gridCell` is either `null` or it has all of its fields. No need to do `hasPreviousSettings && state.gridState[key].annotImageOpacity !== undefined ? state.gridState[key].annotImageOpacity : ...`  for every field.

metaspace/webapp/src/modules/Datasets/comparison/DatasetComparisonGrid.tsx
```diff
@@ -204,111 +216,88 @@ export const DatasetComparisonGrid = defineComponent<DatasetComparisonGridProps>
...
-            , key, 'max', false)
+        if (gridCell.lockedIntensities[1] !== undefined) {
+          await handleIonIntensityLockChange(gridCell.lockedIntensities[1], key, 'max')
         }
       }
     }
 
-    const unsetAnnotation = (key: string) => {
-      Vue.set(state.annotationData, key, { empty: true })
-      Vue.set(state.gridState, key, { empty: true })
```
LachlanStuart, line 281: BTW, I replaced this `{ empty: true }` state with just `null`. It's a lot easier to work with `null` as you can do `?.` and `state.gridState[key] ? ...`

Issue #888, opened 2021-06-22
TIC Images & Normalization (Front-End)


LachlanStuart:
Blocked by #885 - cannot be started until the back end is ready.

#### Show the TIC Image
- [ ] In the Dataset Overview page, add it to the diagnostics gallery
- [ ] In the Optical Image Alignment page, allow it to be selected as a reference ion image

#### TIC Normalization
- [ ] Add an additional option to the image viewer options for TIC normalization. It makes sense to put this as a "TIC normalization" checkbox under the "Intensity scaling" option in the cog menu
    * When this option is enabled, every pixel in each ion image should be divided by the value for the same pixel in the TIC image. This step should happen before hotspot removal.
    * If this option is selected and the TIC image fails to load, show an error message instead of the ion image. It's better to show nothing than show incorrect data
    * While TIC normalization is active, the "Hotspot removal" and "Outlier clipping" warning messages (when you hover over red numbers in the intensity color scale) should be changed to not refer to these values as "intensities". Ask a scientist how to describe these... maybe "relative intensities"? IDK.
    * While TIC normalization is active, the popup intensity display (when mousing over the ion image) should be changed to show both the non-normalized intensity and the normalized relative intensity.
- [ ] Add a new feature popup directing people to try it out

PR #935, opened 2021-09-01
Feat/ion image tic normaliztion

lmacielvieira:
### Description

Add an option to normalize ion intensities using TIC.
#888  

###### Annotations api
- [x] Returning diagnostics data on dataset detail item

###### ImageLoader
- [x] Added normalization data props
- [x] Creating normalized ionImage if normalization props passed

###### IonImageViewer
- [x] Displaying TIC normalized data on pixel tooltip

###### IonImageRendering
- [x] Applying tic normalization when props passed

###### npyHandler
- [x] Added function to parse npy file

###### AnnotationTable
- [x] Calculating normalization data on selected annotation change

###### IonImageSettings
- [x] Added TIC normalization checkbox

###### MainImageHeader
- [x] Added TIC normalized badge

###### ImageAligner
- [x] Added TIC normalization option

###### Url
- [x] Added norm query filter

###### ImageViewer
- [x] Added normalization error box

##### Front end
 
- [x] Unit Test
- [ ] e2e Test
 
- [x] Chrome
- [x] Safari
- [x] Firefox
- [x] md, lg, lg
- [ ] sm, xl

image
image
image


metaspace/webapp/src/modules/ImageAlignment/ImageAlignmentPage.vue
```diff
@@ -239,10 +258,12 @@ export default {
       annotationIndex: 0,
       file: null,
       opticalImgUrl: null,
+      ticData: null,
       alreadyUploaded: false,
       initialTransform: [[1, 0, 0], [0, 1, 0], [0, 0, 1]],
       padding: 100,
       angle: 0,
+      enableNormalization: false,
```
LachlanStuart, line 266: BTW, the task on this page wasn't to allow normalization. It was to add the TIC image as a selectable reference image (alongside all the other annotations). Now that it's implemented, it's worth keeping, but you should still add selecting the TIC image as a reference image for alignment

LachlanStuart (changes requested):
Generally looks good. The only critical feedback item is addressing the performance hit that comes from including diagnostics in the `AnnotationDetailItem` fragment that's used by the annotation list query.

metaspace/webapp/src/api/annotation.ts
```diff
@@ -205,6 +193,32 @@ gql`query GetRelatedAnnotations($datasetId: String!, $filter: AnnotationFilter!,
     }
   }`
 
+export const annotationsDiagnosticsQuery =
+gql`query GetAnnotationsDiagnostics($orderBy: AnnotationOrderBy, $sortingOrder: SortingOrder,
+  $offset: Int, $limit: Int, $query: String,
+  $filter: AnnotationFilter, $dFilter: DatasetFilter) {
+  allAnnotations(filter: $filter, datasetFilter: $dFilter, simpleQuery: $query,
+    orderBy: $orderBy, sortingOrder: $sortingOrder,
+    offset: $offset, limit: $limit) {
```
LachlanStuart, line 202: Why does this query through the `allAnnotations` query instead of going directly to `dataset`? The `datasetId` should already be available through `currentAnnotation.dataset.id`.

Querying it through the annotation makes the cache much less effective because it re-fetches for different annotations even if the dataset is the same. If the query is based on the dataset, it only needs to re-fetch when `datasetId` changes.

Also, please clean up code when you copy+paste it. This query for a single annotation doesn't need all the extra parameters `$orderBy: AnnotationOrderBy, $sortingOrder: SortingOrder,   $offset: Int, $limit: Int, $query: String,   $filter: AnnotationFilter, $dFilter: DatasetFilter`. In fact it shouldn't even go through the `allAnnotations` query - there's an `annotation` query that lets you select a single annotation by ID.

metaspace/webapp/src/lib/ionImageRendering.ts
```diff
@@ -120,20 +122,22 @@ const extractIntensityAndMask = (png: Image, min: number, max: number, normaliza
...
       if (
         normalizationData && normalizationData.data
-          && normalizationData.data.length === numPixels
-          && normalizationData.data[i] && !isNaN(normalizationData.data[i])) {
-        intensity = (intensity / normalizationData.data[i]) * 100000000
+        && normalizationData.data.length === numPixels
+        && normalizationData.data[i] && !isNaN(normalizationData.data[i])
+      ) {
+        intensity = normalizationData.showFullTIC
+          ? normalizationData.data[i] : (intensity / normalizationData.data[i]) * TIC_MULTIPLIER
```
LachlanStuart, line 134: Why does this entangle another unused ion image with the TIC image rendering? This adds a lot of logic to what is supposed to be an optimized loop that's called for every pixel. It makes no sense to involve some random other ion image that doesn't even get used into this codepath.

It would be much cleaner to have an alternative function for the `showFullTIC` codepath, e.g.
```typescript
const getNormalizationIntensityAndMask = (normalizationData: Normalization) => {
  const intensityValues = normalizationData.data
  const mask = new Uint8ClampedArray(intensityValues.length)
  for (let i = 0; i < intensityValues.length; i++) {
    mask[i] = Number.isNaN(intensityValues[i]) ? 0 : 255
  }
  const [width, height] = normalizationData.shape

  return { intensityValues, mask, width, height }
}
```
So that `processIonImage` can do this:
```ts
const { width, height, intensityValues, mask } = normalizationData?.showFullTIC 
    ? getNormalizationIntensityAndMask(normalizationData)
    : extractIntensityAndMask(png, minIntensity, maxIntensity, normalizationData)
```

(Some small fixes to the types of the fields in `Normalization` were required, and `width`/`height`-extraction were moved into `extractIntensityAndMask` for this to work)

This would also mean that unused ion image that currently comes through the `png` parameter wouldn't be necessary at all.

metaspace/webapp/src/modules/ImageAlignment/ImageAlignmentPage.vue
```diff
@@ -381,12 +400,12 @@ export default {
...
     renderAnnotation(annotation) {
       const { ion } = annotation
-      return renderMolFormulaHtml(ion)
+      return annotation.type === 'TIC' ? 'TIC' : renderMolFormulaHtml(ion)
     },
 
     renderLabel(annotation) {
       const { ion } = annotation
-      return renderMolFormula(ion)
+      return annotation.type === 'TIC' ? 'TIC' : renderMolFormula(ion)
```
LachlanStuart, line 408: Please use "TIC Image" to describe it. I've found saying "TIC" alone is confusing to a lot of people

LachlanStuart (approved):

LachlanStuart (approved):

PR #943, opened 2021-09-10
Feat/ds comparison pre release

lmacielvieira:
### Description

Improvement of ds comparison page lock intensities and permalink features before release.
#910  

###### DatasetComparisonPage
- [x] Handling lock all intensities mechanism by props
- [x] Using saved settings by reading snapshot

###### DatasetList
- [x] Showing new feature badge only on first dataset card

##### Front end
 
- [x] Unit Test
- [ ] e2e Test
 
- [x] Chrome
- [ ] Safari
- [ ] Firefox
- [x] md, lg, lg
- [ ] sm, xl

### DIFF
--- a/metaspace/webapp/src/modules/Datasets/comparison/DatasetComparisonGrid.tsx
+++ b/metaspace/webapp/src/modules/Datasets/comparison/DatasetComparisonGrid.tsx
@@ -20,6 +20,7 @@ import { ExternalWindowSvg } from '../../../design/refactoringUIIcons'
 import { Popover } from '../../../lib/element-ui'
 import { ImagePosition } from '../../ImageViewer/ionImageState'
 import { range } from 'lodash-es'
+import config from '../../../lib/config'
 
 const RouterLink = Vue.component('router-link')
 
@@ -226,6 +227,7 @@ export const DatasetComparisonGrid = defineComponent<DatasetComparisonGridProps>
       const ionImagePng = await loadPngFromUrl(annotation.isotopeImages[0].url)
       let gridCell: GridCellState
 
+      console.log('HALLo')
       if (hasPreviousSettings) {
         gridCell = state.gridState[key]!
         gridCell.ionImagePng = ionImagePng
@@ -235,7 +237,7 @@ export const DatasetComparisonGrid = defineComponent<DatasetComparisonGridProps>
         const pixelSizeX = metadata?.MS_Analysis?.Pixel_Size?.Xaxis || 0
         // eslint-disable-next-line camelcase
         const pixelSizeY = metadata?.MS_Analysis?.Pixel_Size?.Yaxis || 0
-
+        console.log('pixelSizeXY', pixelSizeX, pixelSizeY)
         gridCell = reactive({
           intensity: null, // @ts-ignore // Gets set later, because ionImageLayers needs state.gridState[key] set
           ionImagePng,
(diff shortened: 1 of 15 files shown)


metaspace/webapp/src/modules/Datasets/comparison/DatasetComparisonGrid.tsx
```diff
@@ -666,7 +681,8 @@ export const DatasetComparisonGrid = defineComponent<DatasetComparisonGridProps>
           ? props.datasets.find((dataset: any) => dataset.id === settings.value.grid[`${row}-${col}`])
           : null
       return (
-        <span class='dataset-comparison-grid-ds-name'>{dataset?.name}</span>
+        <span class='dataset-comparison-grid-ds-name'>{dataset?.name?.substring(0, 39)}
+          {dataset?.name?.length > 40 ? '...' : ''}</span>
```
LachlanStuart, line 685: It's usually best to handle this [with CSS](https://tailwindcss.com/docs/text-overflow#truncate) so that it can adapt to any size of text or any number of items in the grid. 

There's an example of this on [dataset titles in the dataset list](https://github.com/metaspace2020/metaspace/blob/master/metaspace/webapp/src/modules/Datasets/list/DatasetItemMetadata.tsx#L73-L77) though it only partially uses tailwind. Note that it also [sometimes requires some of the parent elements to have extra styles like `min-width: 0px`](https://css-tricks.com/flexbox-truncated-text/) so that the browser is able to calculate how much is available.

metaspace/webapp/src/modules/Datasets/comparison/DatasetComparisonGrid.tsx
```diff
@@ -265,17 +270,20 @@ export const DatasetComparisonGrid = defineComponent<DatasetComparisonGridProps>
...
+        ? globalLockedIntensities.value[1] : (intensity.max.clipped || intensity.max.image)
       gridCell.intensity = intensity
-      gridCell.lockedIntensities = state.globalLockedIntensities
+      gridCell.lockedIntensities = globalLockedIntensities.value as [number | undefined, number | undefined]
+
       // persist ion intensity lock status
       if (gridCell.lockedIntensities !== undefined) {
         if (gridCell.lockedIntensities[0] !== undefined) {
           await handleIonIntensityLockChange(gridCell.lockedIntensities[0], key, 'min')
+          // await handleIonIntensityChange(gridCell.lockedIntensities[0], key, 'min')
```
LachlanStuart, line 282: Is this meant to be commented out? It's not commented out in the `max` branch

metaspace/webapp/src/modules/Datasets/comparison/SimpleShareLink.tsx
```diff
@@ -19,37 +40,116 @@ export const SimpleShareLink = defineComponent<SimpleShareLinkProps>({
...
+    nCols: { type: Number },
+    nRows: { type: Number },
+    settings: { type: String },
+    colormap: { type: String },
+    scaleType: { type: String },
+    sourceDsId: { type: String },
+    scaleBarColor: { type: String },
+    selectedAnnotation: { type: Number },
+    annotations: { type: Array },
+    datasets: { type: Array },
```
LachlanStuart, line 53: `annotations` and `datasets` seem to be unused and should be cleaned up.

It's also really weird passing all this data to what should be a "Simple" link. Have you considered making the parent component handle building the `ImageViewerSnapshotInput` object, so that `handleClick` can just do `const variables = { input: inputSnapshot }` and not need to manage so many props?

LachlanStuart (approved):
Generally looks good. I've left some comments where I feel code could be improved, but none are blockers.

PR #949, opened 2021-09-20
[TIC] added tic feature as feature flag

lmacielvieira:
### Description

Added a feature flag to show TIC normalization feature (feat=tic), renamed TIC intensity label value to TIC-relative intensity and moved TIC normalization explanation to TIC normalized checkbox.
#888

### DIFF
--- a/metaspace/webapp/src/modules/Annotations/annotation-widgets/default/IonImageSettings.vue
+++ b/metaspace/webapp/src/modules/Annotations/annotation-widgets/default/IonImageSettings.vue
@@ -74,10 +74,23 @@
       >
         <el-checkbox
           :value="normalization"
-          class="font-thin"
           @change="onNormalizationChange"
         >
-          TIC normalization
+          <div class="font-thin flex flex-row items-center">
+            TIC normalization
+            <el-popover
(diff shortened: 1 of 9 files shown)


metaspace/webapp/src/modules/Annotations/annotation-widgets/default/IonImageSettings.vue
```diff
@@ -74,10 +74,23 @@
...
         >
-          TIC normalization
+          <div class="font-thin flex flex-row items-center">
+            TIC normalization
+            <el-popover
+              trigger="hover"
+              placement="right"
+            >
+              <div slot="reference">
+                <i class="el-icon-question metadata-help-icon ml-1" />
```
LachlanStuart, line 86: `metadata-help-icon` is only included in the upload page bundle, which means the `cursor: pointer` style will only be applied here if the user has visited that page without reloading. I noticed you've copy-pasted this in a few other places (`IonImageSettings.vue`, etc.). I suggest you replace it with Tailwind's `cursor-pointer`

LachlanStuart (approved):

PR #977, opened 2021-11-02
[MetadataEditor] Added template selector for dataset upload

lmacielvieira:
### Description

Added option to pre fill upload metadata info from previous uploaded dataset at upload page #963.

###### MetadataEditor
- [x] Added remote search to load previous uploaded datasets
- [x] Filling metadata according to selected dataset template

##### Front end
 
- [ ] Unit Test
- [ ] e2e Test
 
- [x] Chrome
- [ ] Safari
- [ ] Firefox
- [x] md, lg, lg
- [ ] sm, xl

image

### DIFF
--- a/metaspace/webapp/src/modules/MetadataEditor/MetadataEditor.vue
+++ b/metaspace/webapp/src/modules/MetadataEditor/MetadataEditor.vue
@@ -7,47 +7,42 @@
       <div id="md-section-list">
         <div
           v-if="isNew"
-          class="flex flex-row w-full flex-wrap mt-6"
+          class="flex flex-row w-full flex-wrap mt-6 justify-end"
         >
-          <div class="metadata-section__title w-3/12">
-            Metadata template
-            <el-popover
-              trigger="hover"
-              placement="top"
-            >
-              <div class="max-w-sm">
-                Type the name of a previous uploaded dataset and select
-                it to use its metadata as a template for you new upload!
-              </div>
-              <i
-                slot="reference"
-                class="el-icon-question metadata-help-icon"
-              />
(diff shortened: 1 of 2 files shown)


metaspace/webapp/src/modules/MetadataEditor/MetadataEditor.vue
```diff
@@ -5,6 +5,50 @@
...
+            >
+              <div class="max-w-sm">
+                Type the name of a previous uploaded dataset and select
+                it to use its metadata as a template for you new upload!
+              </div>
+              <i
+                slot="reference"
+                class="el-icon-question metadata-help-icon"
+              />
+            </el-popover>
```
LachlanStuart, line 26: Now that I've had a chance to use it, I feel this looks a bit too much like it's part of the form - users might think that they need to fill it out, not realizing that it's not necessary to use this to submit, and that selecting an option from the list destructively overwrites any data they've already entered without warning... 

Any ideas for how to distinguish this from the form input fields? All I can think of is making a "Copy metadata from another dataset..." button that opens a dialog box with this dropdown inside of it...

LachlanStuart (changes requested):
I tried it locally and it seems to work well. I've made some suggestions around the query but none of them are critical. The only objection I have is about the UX - to me it feels too much like it's a form field, rather than a discrete action. I think this will confuse users and some may unintentionally overwrite their entered data by trying to put a value in this field. 

Feel free to ask Theo or some users for a second opinion though - if I'm the only one who thinks it's a confusing UX then I'll withdraw the objection.

LachlanStuart (approved):

Issue #973, opened 2021-10-21
[m/z filter] Persist trailing zeros after the decimal place


lmacielvieira:
when entering m/z value ending with 0, the filter form doesn’t preserve it and skips it. e.g. entering 267.090 gives 267.09. This is important because +- m/z is estimated based on the number of zeros. It should not skip zeros at the end.

PR #979, opened 2021-11-05
[MzFilter] Added raw input option to enable trailing zeros

lmacielvieira:
### Description

Changed mz filter to keep trailing zeros  #973 .

###### NumberFilter
- [x] Added normal input option to enable string input, so the trailing zeros can persist

##### Front end
 
- [x] Unit Test
- [ ] e2e Test
 
- [x] Chrome
- [ ] Safari
- [ ] Firefox
- [x] md, lg, lg
- [ ] sm, xl

### DIFF
--- a/metaspace/webapp/src/lib/util.ts
+++ b/metaspace/webapp/src/lib/util.ts
@@ -81,7 +81,7 @@ export function decodePayload(jwt: JWT) {
 
 export function mzFilterPrecision(value: number | string): string {
   // Using parseFloat to remove any extra decimal places that won't actually count toward the precision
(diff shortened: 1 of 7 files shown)


LachlanStuart (changes requested):
This fixes the UI but doesn't seem to fix the request sent to graphql. E.g. if I enter `296.0000` it shows the correct filter "296.0000 ± 0.0001", but I actually see all annotations between 295 and 297, the graphql request shows why: `"mzFilter":{"min":295,"max":297}`. Tested on Firefox and Chrome.

LachlanStuart (approved):

Issue #988, opened 2021-11-12
Pressing Enter in the Create Project dialog refreshes the page without creating a project


LachlanStuart:
## Steps to reproduce:
1. Go to the user page: https://metaspace2020.eu/user/me
1. Click Create Project
1. Enter a project name
1. Press enter (this is where the bug happens - the page refreshes, maybe it's saving the user details form in the background or something)

Most likely cause is that either there's no HTML `<form>` around the name input, or that form isn't handling and `e.preventDefault()`ing the `submit` event.

PR #991, opened 2021-11-15
[EditProjectForm] fixed validations and enter press bug

lmacielvieira:
### Description

Pressing Enter in the Create Project dialog refreshes the page without creating a project #988 

* Highlights
 - It seems there is a bug on vue form where pressing enter key in input causes app to reload when there is only one input in the form (i.e https://forum.framework7.io/t/vue-pressing-enter-key-in-input-causes-app-to-reload/2585), so a hidden input was added to prevent it.
 - The project name validation wasn't being triggered, so I added the validation in the forms submit and the el-form-item wrapper

### DIFF
--- a/metaspace/webapp/src/modules/Project/EditProjectForm.vue
+++ b/metaspace/webapp/src/modules/Project/EditProjectForm.vue
@@ -11,12 +11,20 @@
     <div>
       <label>
         <primary-label-text>Title</primary-label-text>
-        <el-input
-          v-model="value.name"
-          class="py-1"
-          :max-length="50"
-        />
+        <el-form-item prop="name">
+          <el-input
+            v-model="value.name"
+            class="py-1"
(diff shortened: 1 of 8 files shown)


metaspace/webapp/src/modules/Project/EditProjectForm.vue
```diff
@@ -11,12 +11,20 @@
...
+          <el-input
+            v-model="value.name"
+            class="py-1"
+            :max-length="50"
+            :min-length="2"
+            validate-event
+          />
+        </el-form-item>
       </label>
+      <!--      empty and hidden input to prevent press enter reload vue bug-->
```
LachlanStuart, line 24: Please include a link to that issue in the comment so we can find more info if this code is edited later.

LachlanStuart (approved):

Issue #996, opened 2021-11-26
Fix Metadata Additional info bug


lmacielvieira:
The Additional info in the 'Show full metadata' dialog is not showing the correct description related to the dataset

PR #1000, opened 2021-12-03
Feat/metadata additional info bug

lmacielvieira:
### Description

The additional information field displayed on 'Show full metadata' should display the dataset description and not the old value as it is now #996 

##### Front end
 
- [ ] Unit Test
- [ ] e2e Test
 
- [x] Chrome
- [x] Safari
- [ ] Firefox
- [x] md, lg, lg
- [ ] sm, xl


LachlanStuart (commented):
Could you please benchmark the rendering performance of the Datasets page (with at least 10 datasets that have descriptions set) before & after this change?

I don't know if this will affect performance, but we need to be careful because the datasets page rendering is already noticeably slow. I've been surprised a few times by ElementUI components like `el-dialog`, `el-tooltip` and `el-collapse` rendering their contents even when they're not visible.

PR #1004, opened 2021-12-06
Set Lithops 2.5.8

sergii-mamedov:


sergii-mamedov:
I checked CF and VPC - everything is ok.

metaspace/engine/sm/engine/annotation_lithops/executor.py
```diff
@@ -24,7 +24,7 @@
 #: manually updating their config files every time it changes. The image must be public on
 #: Docker Hub, and can be rebuilt using the scripts/Dockerfile in `engine/docker/lithops_ibm_cf`.
 #: Note: sci-test changes this constant to force local execution without docker
-RUNTIME_DOCKER_IMAGE = 'metaspace2020/metaspace-lithops:1.9.0'
+RUNTIME_DOCKER_IMAGE = 'metaspace2020/metaspace-lithops:1.9.4'
```
LachlanStuart, line 27: Did the Dockerfile change at all? It's fine if not - it's just that it doesn't include Lithops, so normally the container wouldn't change between Lithops versions unless the Dockerfile or some other dependencies changed.

LachlanStuart (approved):

metaspace/engine/sm/engine/annotation_lithops/executor.py
sergii-mamedov: I missed that we don't include lithops in the temporary `requirements` file. Thank you for paying attention.

Issue #1005, opened 2021-12-07
Adduct filter is disappearing in dataset comparison page


lmacielvieira:
It was reported that when typing some adducts on ds comparison page filter, it is disappearing. i.e hcl

PR #1006, opened 2021-12-07
[DsComparisonTable] removed onkey hl keys from listener

lmacielvieira:
### Description

It was reported that when typing some adducts on ds comparison page filter, it is disappearing. i.e hcl #1005 

##### Front end
 
- [ ] Unit Test
- [ ] e2e Test
 
- [x] Chrome
- [x] Safari
- [ ] Firefox
- [x] md, lg, lg
- [ ] sm, xl


LachlanStuart (approved):
This change is fine, but wouldn't it also be a good idea to limit the scope of the event listener, instead of attaching it to `window`, so that pressing left/right while updating a filter doesn't cause issues? `AnnotationTable.vue`, for example, puts the keyup listener on the `el-table` element.

LachlanStuart (approved):

PR #1007, opened 2021-12-08
Fix Python Client issue with slow uploading & add save_login function

LachlanStuart:
#### Fixes issue with slow uploads

Måns found that uploads that take longer than 5 minutes would fail because the pre-signed URLs would expire. The cause for this was that `ThreadPoolExecutor` evaluated `iterate_file` eagerly, loading the entire file into memory and pre-signing each of the chunks before there was an uploader thread ready to upload it. 

This was fixed by using a semaphore to prevent `iterate_file` from reading ahead when all the upload threads are busy, and by moving the pre-signing into the `upload_part` function.

#### Adds save_login function

Loading login details from a config file was implemented but never really documented. Many users actually just directly put their api_key in the source code, which is really bad. The config file approach isn't perfect, but it's much better than risking committing the api_keys to a public repository. 

This change adds a public API for creating the config file, and uses it in all the example notebooks.

### DIFF
--- a/metaspace/python-client/docs/source/content/examples/fetch-dataset-annotations.ipynb
+++ b/metaspace/python-client/docs/source/content/examples/fetch-dataset-annotations.ipynb
@@ -57,9 +57,36 @@
    ],
    "source": [
     "from metaspace import SMInstance\n",
-    "sm = SMInstance()\n",
-    "sm"
+    "sm = SMInstance()\n"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "source": [
+    "### Enter your API Key (only required for private datasets)\n",
+    "\n",
+    "To access private datasets on METASPACE, generate an API key from your [account page](https://metaspace2020.eu/user/me) and enter it when prompted below.\n",
+    "This step can be skipped for accessing public datasets."
+   ],
+   "metadata": {
+    "collapsed": false
+   }
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "outputs": [],
+   "source": [
+    "# This will prompt you to enter your API key if needed and it will save it to a config file.\n",
+    "# Note that API keys should be kept secret like passwords.\n",
+    "sm.save_login()"
+   ],
+   "metadata": {
+    "collapsed": false,
+    "pycharm": {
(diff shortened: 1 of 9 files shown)


metaspace/python-client/docs/source/content/examples/fetch-dataset-annotations.ipynb
```diff
@@ -57,10 +57,37 @@
...
+   "outputs": [],
+   "source": [
+    "# This will prompt you to enter your API key if needed and it will save it to a config file.\n",
+    "# Note that API keys should be kept secret like passwords.\n",
+    "sm.save_login()"
+   ],
+   "metadata": {
+    "collapsed": false,
+    "pycharm": {
+     "name": "#%%\n"
```
sergii-mamedov, line 87: Good feature. 
Maybe not now, but I would move all the authorization information to a separate notebook. We are now duplicating this across all notebooks.

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -1460,11 +1481,50 @@ def __init__(
...
+            else:
+                print(f'Failed to connect to {self._config["host"]}: {ex.message}')
 
     def __repr__(self):
         return "SMInstance({})".format(self._config['graphql_url'])
 
+    def save_login(self, overwrite=False):
+        """Saves login credentials to the config file so that they will be automatically loaded
+        in future uses of SMInstance()"""
+        config_path = Path.home() / '.metaspace'
```
sergii-mamedov, line 1504: The `save_login` function name is a little misleading as we are storing the KEY API.

sergii-mamedov (approved):
LGTM, thanks for these improvements

metaspace/python-client/metaspace/sm_annotation_utils.py
LachlanStuart: I try to avoid using technical terms in the API - we have many users with non-technical backgrounds, and some of them already have trouble reading the documentation. With "login" in the name, they're more likely to remember it and understand what it does.

I would have preferred to just use `SMInstance.login` as it's more direct, but that's already taken by a deprecated method...

Issue #990, opened 2021-11-15
[AnnotationTable] Make annotation table columns customizable


lmacielvieira:

LachlanStuart:
To help with planning this, here's my wishlist for improvements to the Annotations table. This doesn't all have to be implemented as part of this task, I'm just listing ideas I've had over the years.

* Ability to add columns: 
    * Molecules (comma-separated list of names)
    * Database
    * Max intensity
    * Total intensity
    * Adduct (separately from the formula so that it can be used for sorting, or so that you can see the molecule name and adduct without the formula taking up space)
    * Spatial/Spectral/Chaos metrics
    * On-/Off-sample
    * If isomers/isobars are present (Just a column that's usually empty but shows a caution icon for molecules with isomers/isobars)
* Ability to hide each existing column
* Ability to adjust the size of the table so that the image view gets more screen space (e.g. with a draggable divider), or even hide the annotations table completely

PR #1016, opened 2021-12-15
Feat/customizable annotation table

lmacielvieira:
### Description

A new feature which makes possible select which columns to be shown on annotation table.
#990  

###### Annotation Table
- [x] Added table full screen mode
- [x] Added dynamic column rendering
- [x] Persisting column settings on local storage

###### esConnector
- [x] Added sort by adduct
- [x] Added sort by group
- [x] Added sort by dataset name
- [x] Added sort by database name
- [x] Added sort by chaos metric
- [x] Added sort by spatial metric
- [x] Added sort by spectral metric
- [x] Added sort by max intensity
- [x] Added sort by total intensity
- [x] Added sort by isomers/isobars

##### Front end
 
- [x] Unit Test
- [ ] e2e Test
 
- [x] Chrome
- [x] Safari
- [ ] Firefox
- [x] md, lg, lg
- [ ] sm, xl

image
image
image


metaspace/webapp/src/modules/Annotations/AnnotationTable.vue
```diff
@@ -147,32 +165,142 @@
         property="offSampleProb"
         label="Off-sample %"
         sortable="custom"
-        min-width="60"
+        min-width="120"
       >
         <template slot-scope="props">
           <span> {{ (props.row.offSampleProb * 100).toFixed(0) }}% </span>
```
LachlanStuart, line 171: This seems to show `0%` when `row.offSampleProb` is `null`. Please change this to handle nulls - maybe just showing `- %`.

Nulls happen because annotations are visible before off-sample classification has finished running. Also, sometimes datasets hit an annotation limit and it skips classification for the lowest-quality ion images. Previously this field was completely hidden, but now since all users can see it it's important that it doesn't show potentially misleading data.

metaspace/webapp/src/modules/Annotations/AnnotationTable.vue
```diff
@@ -504,6 +780,9 @@ export default Vue.extend({
       },
     },
   },
+  created() {
+    this.columns = getLocalStorage('annotationTableCols') || this.columns
```
LachlanStuart, line 784: The previous behavior was to hide the Lab and Dataset columns if there is a dataset filter and show them if there is no dataset filter.  Please bring this behavior back somehow, as it's annoying to have to manually show/hide them.

metaspace/webapp/src/modules/Annotations/AnnotationTable.vue
```diff
@@ -339,6 +533,88 @@ export default Vue.extend({
...
+          src: 'rhoSpectral',
+          selected: false,
+        },
+        {
+          label: 'Chaos metric',
+          src: 'rhoChaos',
+          selected: false,
+        },
+        {
+          label: 'On/Off-sample %',
```
LachlanStuart, line 608: I suggest changing this to "Off-sample probability %" to be clearer. It's the probability that the whole image shows an off-sample detection, not the percentage of the image that is off-sample.

metaspace/webapp/src/modules/Annotations/AnnotationTable.vue
```diff
@@ -339,6 +533,88 @@ export default Vue.extend({
...
+          src: 'maxIntensity',
+          selected: false,
+        },
+        {
+          label: 'Total Intensity',
+          src: 'totalIntensity',
+          selected: false,
+        },
+        {
+          label: 'Adduct',
```
LachlanStuart, line 588: It would be good if the order of items here matched the order of columns in the grid. Adduct especially feels out of place - it should be next to Annotation, like in the grid.

metaspace/webapp/src/modules/Annotations/AnnotationTable.vue
LachlanStuart, line 784: Overwriting the whole `columns` field with the saved value from localStorage isn't future-proof. If we change/add/remove columns, existing users won't see those changes because they'll always be overwritten by the copy saved in their localStorage.

Could you please change this storage strategy somehow so that future updates/additions/removals to `this.columns` can't be overwritten by old localStorage values?

LachlanStuart (changes requested):
Overall looks really good, but I've listed a few small-but-necessary changes in the comments. The critical ones are:

* Handle null `offSampleProb`s
* Hide Lab/Dataset when user enters the page with a dataset filter
* Future-proof the localStorage field

LachlanStuart (approved):

PR #1017, opened 2021-12-20
Save user_id and input_path into table

sergii-mamedov:
As part of this PR I am adding the following functionality:

1. Create table migration and add columns `user_id` and` input_path`.
2. Add functionality to save `user_id` and` input_path` for new custom DBs.


metaspace/graphql/src/modules/moldb/controller.ts
```diff
@@ -20,6 +21,10 @@ const MolecularDbResolvers: FieldResolversFor<MolecularDB, MolecularDbModel> = {
   hidden(database): boolean {
     return database.archived || !database.isPublic
   },
+
+  user(database): User {
```
sergii-mamedov, line 25: I'm not completely sure of the correctness, I will just comment.

LachlanStuart (changes requested):
The new `MolecularDB.user` field always returns an error. Here's how to test the graphql API manually:
1. Set up data in your local environment
2. Go to http://localhost:8888/graphql (change the port if needed)
3. On the settings page set `"request.credentials": "include"` so that your browser cookies are used for authentication. You can log in via the web app.
4. Enter a query that includes the new/changed fields, e.g.
```graphql
query {
  molecularDB (databaseId:48) {
    user { id name email }
  }
}
```

Critical changes:
* Fix the graphql schema type of the `user` field
* Make `MolecularDB.user` resolver return a `UserSource` object
* Make sure it runs

metaspace/graphql/src/modules/moldb/controller.ts
```diff
@@ -22,8 +24,47 @@ const MolecularDbResolvers: FieldResolversFor<MolecularDB, MolecularDbModel> = {
...
+    const databaseGroupId = database.groupId!.toString()
+
+    if (ctx.isAdmin || userGroupIds.includes(databaseGroupId)) {
+      return getUserSourceById(ctx, database.userId!)
+    } else {
+      return null
+    }
+  },
+
+  async downloadLink(database) {
```
LachlanStuart, line 38: Per discussion IRL/in Slack: we should confirm with Theo whether additional authorization is needed here to restrict download access to people in the same group as the database.

metaspace/graphql/src/modules/moldb/controller.ts
```diff
@@ -20,6 +23,49 @@ const MolecularDbResolvers: FieldResolversFor<MolecularDB, MolecularDbModel> = {
   hidden(database): boolean {
     return database.archived || !database.isPublic
   },
+
+  user: async function(database, args: any, ctx): Promise<User|null> {
```
LachlanStuart, line 27: A tip for working with resolver functions: TypeScript can usually type-check the parameters, but WebStorm/Pycharm often don't understand them and can't give feedback/code suggestions. It makes life a lot easier if you put type annotations on the 1st and 3rd arguments, e.g. `user: async function(database: MolecularDbModel, args: any, ctx: Context)`.

metaspace/graphql/src/modules/moldb/controller.ts
```diff
@@ -20,6 +23,49 @@ const MolecularDbResolvers: FieldResolversFor<MolecularDB, MolecularDbModel> = {
...
+        .map(obj => obj.Key!)
+        .filter(key => key && /(\.tsv|\.TSV|.csv)$/i.test(key.toLocaleLowerCase()))
+
+      const file = fileKey.map(key => ({
+        filename: key.replace(/.*\//, ''),
+        link: s3.getSignedUrl('getObject', { Bucket: bucket, Key: key, Expires: 1800 }),
+      }))
+      return JSON.stringify({
+        file,
+      })
```
LachlanStuart, line 64: `database.inputPath` appears to save the full path to the exact S3 object. Why does this need to list objects and find the csv files? I'm not aware of any scenario where there would be multiple files for a molDB. It would also be better and easier for the front-end if this function didn't return an array in the JSON.

LachlanStuart (changes requested):
Overall it looks much better, but I think these changes are needed:
* Remove `!` non-null assertions in `MolecularDB.user` (mainly to fix the `TypeError` on dbs with a `null` groupId)
* Only return a single item in `MolecularDB.downloadLink` (or justify returning an array)
* Wait for Theo's feedback about whether access controls are needed in `MolecularDB.downloadLink`

LachlanStuart:
This change is still required:

> * Remove ! non-null assertions in MolecularDB.user (mainly to fix the TypeError on dbs with a null groupId)

LachlanStuart (approved):

PR #1019, opened 2022-01-07
ML Scoring

LachlanStuart:
We've already discussed what/why/etc. so here I'll just describe how it was implemented and what remains to be done

@sergii-mamedov please review the back-end changes and @lmacielvieira please review the front-end changes

# Setup

After restarting graphql to add the new table, the `v3_default` scoring model can be imported by running this:

```
dc exec api sh  # If you don't have Python set up outside Docker, open a shell inside the API container

cd /opt/dev/metaspace/metaspace/engine

python -m scripts.import_scoring_model v3_default ../scoring-models/v3_default/model-2022-01-05T13-45-26.947188-416b1311.cbm sm-engine-dev
```

If there are any problems with not being able to import catboost, try rebuilding the docker images:
```
dc rm -fs api lithops-daemon annotate-daemon update-daemon
dc build api lithops-daemon annotate-daemon update-daemon
dc up -d api lithops-daemon annotate-daemon update-daemon
```

There are 2 feature flags:
* http://localhost:8999/upload?feat=advanced_ds_config,save  to show the "analysis version" field in the upload page
* http://localhost:8999/annotations?feat=raw_fdr to show the raw FDR values (e.g. 3%, 18% instead of 5%, 20%)

# Architecture Overview

## `annotation.stats` JSON column

The schema hasn't changed, but `msm` has been removed (it was unused & also in a column) and more fields have been added:

Old style:
```json
{
  "chaos": 0.9819444444444444,
  "spatial": 0.048468438199788286,
  "spectral": 0.9251577868703323,
  "msm": 0.04403132469,
  "total_iso_ints": [1504149.575683593, 720507.0854492179, 24994.709472656, 117403.16796875],
  "min_iso_ints": [0, 0, 0, 0],
  "max_iso_ints": [75321.7109375, 45423.57421875, 20302.056640625, 42440.9453125],
}
```

New style:
```json
{
  "chaos": 0.9976041666666666,
  "spatial": 0.005921380836044965,
  "spectral": 0.7906196869421936,
  "total_iso_ints": [348247.875, 1093766.0, 1901134.625, 36702.76953125],
  "min_iso_ints": [0.0, 0.0, 0.0, 0.0],
  "max_iso_ints": [4206.11474609375, 1619.6641845703125, 1990.939453125, 1607.725341796875],
  "mz_mean": [623.3374633789062, 624.34033203125, 625.3348999023438, 626.3374633789062],
  "mz_stddev": [0.0003070708771701902, 0.0006269849836826324, 0.0007836259319446981, 0.0011551531497389078],
  "mz_err_abs": 0.0016151343006640673,
  "mz_err_rel": -0.0008197362767532468,
  "theo_mz": [623.3358764648438, 624.3392333984375, 625.3342895507812, 626.3370361328125],
  "theo_ints": [100.0, 37.65089797973633, 37.747135162353516, 13.01543140411377],
}
```

* `mz_mean` and `mz_stddev` are new per-image mass error fields that are interesting for diagnostics
* `mz_err_abs` and `mz_err_rel` are the 2 new metrics
* `theo_mz` and `theo_ints` are technically redundant, as they can be recalculated, however it has always been a pain to recalculate these fields during data analysis as they depend on dataset metadata.

TBH I'm not happy that this field has become so big. It used to be ~300B/row and now it's ~750B. It contains 33 64-bit floats and 7 arrays, so if it was split into non-JSON columns it would be ~320B with the new data. However, the annotation table is very infrequently accessed (all Annotation queries go to ElasticSearch instead), so it wasn't pragmatic for me to spend time optimizing this.

## ScoringModel

This table/class/`ds_config` item represents a named model that transforms some set of features into a score. By default this is `None` for V1/V2 and `'v3_default'` for V3. Specifying `None` will revert to the old MSM calculation (spatial*spectral*chaos). Using any non-null string will look up the ScoringModel row in the database try to load it as a CatBoost model. 

The features passed to CatBoost depend on the features specified in the ScoringModel metadata. This is to support new models that may use different configurations of features. Some "derived features" (e.g. the FDR-normalized features) need to be calculated based on the overall results for the dataset, so there's a function `add_derived_features` that adds these. Additionally, there's `remove_uninteresting_features`, which removes temporary features that don't need to be saved to `stats`.

The model that is included in the repository does not use the FDR-normalized metrics I've previously described.

### Continuous FDR values + "Rule of succession" approximation

V3 datasets now save exact FDR values (e.g. 0.003812) instead of snapping to the 5/10/20/50% thresholds. Webapp has been changed to snap to 5/10/20/50%, but the raw values can be viewed through the API, CSV export or with the `raw_fdr` feature flag.

### FDR_RESULTS Diagnostics

Every `job` now produces a `FDR_RESULTS` diagnostics entry, which links to a set of 3 parquet files for mapping target/decoy adducts to formulas to metrics. The main documentation is in `diagnostics.py`. 

### Decoy adduct selection

Under V1, each target adduct gets its own independent ranking where 20 decoy adducts are randomly selected. With 3 target adducts this means 3x20 random selections, averaging 53 decoy adducts per formula after deduplication. This is a wasteful, as there's no dependency between the rankings, so there's no reason to resample the decoy adducts

If V3 mode is selected, the randomly selected decoy adducts are reused for all target adduct rankings. This means  ~60% fewer ion formulas are searched for with 3 target adducts, or ~40% fewer with 2 target adducts.

## Lithops pipeline changes

* I rewrote MolDB segmentation to use a non-recursive algorithm for splitting the work into segments, as I was having trouble making the old version stable when annotating with 80 decoy adducts. Hopefully now it'll be a bit easier to tune the job size.
* Optimized the PNG encoding step - more accurate prediction of CPU time required to encode, and smaller chunks of images
* Added a "store images" step that posts images directly from IBM Cloud Functions to S3. This should significantly speed up the last step and decrease CPU usage on the web server for datasets with >5000 images. 

## `sm.fdr_engineering` directory

This is the code used for selecting & reprocessing datasets, downloading their FDR Diagnostics, and training new models. It's meant to be run interactively - all the `run___.py` files are structured like notebooks, using the [`#%%` "percent format"](https://github.com/mwouts/jupytext/blob/main/docs/formats.md#The-percent-format) to indicate cells. Use PyCharm Scientific Mode, Spyder or Jupytext to make life easier running these files.

# Remaining work

(I'll do these in other PRs)

- [x] Add script & documentation for importing ScoringModels
- [x] Investigate inconsistent results between Spark and Lithops implementations - some metrics differ very slightly between the implementations, usually only by 1-2 bits. I suspect it's either due to differences in the order of pixels or floating point numbers having different precisions in different codepaths
- [x] Add a V3 mode to sci-test with reference results
- [ ] Add an "FDR-normalized" model


docker/docker-compose.yml
```diff
@@ -164,6 +164,7 @@ services:
       - "3010:3010"
       - "5666:5666"
       - "4201:4201"
+      - "9229:9229"
```
LachlanStuart, line 167: Unrelated change - I added this to make graphql node debugging easier

metaspace/engine/.dockerignore
```diff
@@ -0,0 +1,3 @@
+*
+!docker
+!requirements.txt
```
LachlanStuart, line 3: Unrelated change: This speeds up the Docker build by making it so that Docker doesn't need to copy the whole `engine` directory into the build context

metaspace/engine/environment.yml
```diff
@@ -1,5 +1,5 @@
-name: sm
+name: sm38
 dependencies:
-  - python=3.6
+  - python=3.8
```
LachlanStuart, line 3: Unrelated change, and I suspect this file might no longer be needed because it has been invalid for so long...

metaspace/engine/sm/engine/annotation/fdr.py
```diff
@@ -1,94 +1,26 @@
...
+    # fmt: off
+    '+He', '+Li', '+Be', '+B', '+C', '+N', '+O', '+F', '+Ne', '+Mg',
+    '+Al', '+Si', '+P', '+S', '+Cl', '+Ar', '+Ca', '+Sc', '+Ti', '+V',
+    '+Cr', '+Mn', '+Fe', '+Co', '+Ni', '+Cu', '+Zn', '+Ga', '+Ge', '+As',
+    '+Se', '+Br', '+Kr', '+Rb', '+Sr', '+Y', '+Zr', '+Nb', '+Mo', '+Ru',
+    '+Rh', '+Pd', '+Ag', '+Cd', '+In', '+Sn', '+Sb', '+Te', '+I', '+Xe',
+    '+Cs', '+Ba', '+La', '+Ce', '+Pr', '+Nd', '+Sm', '+Eu', '+Gd', '+Tb',
+    '+Dy', '+Ho', '+Ir', '+Th', '+Pt', '+Os', '+Yb', '+Lu', '+Bi', '+Pb',
+    '+Re', '+Tl', '+Tm', '+U', '+W', '+Au', '+Er', '+Hf', '+Hg', '+Ta',
+    # fmt: on
```
LachlanStuart, line 23: Unrelated change, I just got tired of scrolling past this big list every time I opened this file

metaspace/engine/sm/engine/annotation/imzml_reader.py
```diff
@@ -129,6 +129,9 @@ def __init__(self, path: Path):
     def iter_spectra(self, sp_idxs: Sequence[int]):
         for sp_idx in sp_idxs:
             mzs, ints = self._imzml_parser.getspectrum(sp_idx)
+            assert len(mzs) == self._imzml_parser.mzLengths[sp_idx], 'Incomplete .ibd file'
+            assert len(ints) == self._imzml_parser.intensityLengths[sp_idx], 'Incomplete .ibd file'
+            assert len(mzs) == len(ints), f"Spectrum {sp_idx} mz and intensity counts don't match"
```
LachlanStuart, line 134: Unrelated change: this just gives a better error message when an incomplete .ibd file has been uploaded

metaspace/engine/sm/engine/annotation/isocalc_wrapper.py
```diff
@@ -124,13 +121,32 @@ def centroids(self, formula):
...
-            half_width = mzs * self.ppm * 1e-6
-        elif self.instrument == 'Orbitrap':
-            half_width = np.sqrt(mzs / BASE_MZ) * (BASE_MZ * self.ppm * 1e-6)
+        if self.analysis_version == 2:
+            # analysis_version==2 adjusts the width to scale with the same function as the
+            # instrument resolving power. This was not included in analysis_version==3 as we
+            # ran out of time to demonstrate whether it would improve the results.
+            half_width = mass_accuracy_half_width(mzs, self.instrument, self.ppm)
         else:
-            half_width = BASE_MZ * self.ppm * 1e-6
```
LachlanStuart, line 132: These old calculations were wrong for `analysis_version==2`. The FT-ICR codepath used TOF scaling, and the TOF codepath used "constant mass accuracy" scaling, which AFAIK doesn't match any available analyzer.

This was definitely a bug, but fortunately I don't think anyone relied on this behavior it so I just fixed it and added testsin the new version.

metaspace/engine/sm/engine/annotation_lithops/run_fdr.py
```diff
@@ -26,25 +29,27 @@ def _run_fdr_for_db(db_data_cobject: CObj[DbFDRData], *, storage: Storage):
...
+            .merge(modifiers, left_on='modifier', right_index=True, how='outer')
+            # .drop(columns=['modifier'])
         )
 
-        if not db_data['targeted']:
-            results_df = results_df[results_df.fdr <= 1]
+        # if not db_data['targeted']:
+        #     results_df = results_df[results_df.fdr <= 1]
 
         return db_data['id'], results_df
```
LachlanStuart, line 46: The FDR Diagnostics bundle needs the full `results_df` including rejected annotations. I moved the targeted/FDR filtering into `prepare_results`

metaspace/engine/sm/engine/image_storage.py
```diff
@@ -1,24 +1,26 @@
...
 from botocore.exceptions import ClientError
 from scipy.ndimage import zoom
-import PIL.Image
 
 from sm.engine.config import SMConfig
 from sm.engine.storage import get_s3_resource, create_bucket, get_s3_client
 from sm.engine.utils.retry_on_exception import retry_on_exception
 
-try:
+if TYPE_CHECKING:
```
LachlanStuart, line 20: Not sure why, but pylint suddenly started complaining about the `try:` code, so I changed it to this.

metaspace/engine/sm/engine/postprocessing/colocalization.py
```diff
@@ -257,14 +257,16 @@ def analyze_colocalization(ds_id, moldb_id, images, ion_ids, fdrs, h, w, cluster
...
     assert images.ref.shape[0] == ion_ids.shape[0] == fdrs.shape[0], (
         images.ref.shape,
         ion_ids.shape,
         fdrs.shape,
     )
     start = datetime.now()
 
+    if images.ref.shape[1] < 3:
+        logger.warning('Not enough pixels per image to perform colocalization.')
+        return
```
LachlanStuart, line 269: I changed this to a soft fail instead of a hard fail because it's actually possible for 2-pixel datasets to be annotated

metaspace/webapp/src/modules/MetadataEditor/MetadataEditPage.vue
```diff
@@ -210,6 +210,14 @@ export default {
...
             type: 'error',
           })
+        } else if (graphQLError && graphQLError.type === 'dataset_busy') {
+          if (this.currentUser && this.currentUser.role === 'admin' && await this.confirmForce()) {
+            return this.saveDataset(datasetId, payload, { ...options, force: true })
+          }
+          this.$message({
+            message: 'This dataset is busy. Please wait until it has finished processing before making changes.',
+            type: 'error',
+          })
```
LachlanStuart, line 220: This allows admins to edit datasets while they're still processing. I could swear this had already been implemented. It makes life a lot easier when datasets keep stalling.

metaspace/webapp/src/modules/Datasets/list/DatasetTable.spec.ts
```diff
@@ -40,6 +40,7 @@ describe('DatasetTable', () => {
     id: 'REPLACEME',
     status: 'FINISHED',
     metadataJson: mockMetadataJson,
+    description: null,
```
LachlanStuart, line 43: Unrelated change: The missing description caused a very spammy warning in the test logs, as the stub graphql server produced a random string mock value.

metaspace/webapp/src/modules/Annotations/annotation-widgets/default/DiagnosticsMetrics.vue
```diff
@@ -1,59 +1,4 @@
...
-  <el-row id="scores-table">
-    <div class="msm-score-calc">
-      MSM score =
-      <span>{{ annotation.msmScore.toFixed(3) }}</span> =
-      <span>{{ annotation.rhoSpatial.toFixed(3) }}</span>
-      (&rho;<sub>spatial</sub>) &times;
-      <span>{{ annotation.rhoSpectral.toFixed(3) }}</span>
-      (&rho;<sub>spectral</sub>) &times;
-      <span>{{ annotation.rhoChaos.toFixed(3) }}</span>
-      (&rho;<sub>chaos</sub>)
```
LachlanStuart, line 11: I converted this file to TSX because making this into a variable list of metrics would have been a nightmare in the template format.

metaspace/webapp/src/api/annotation.ts
```diff
@@ -162,11 +162,26 @@ gql`query Export($orderBy: AnnotationOrderBy, $sortingOrder: SortingOrder,
...
+
+      msmScore
+      fdrLevel
+
+      metricsJson
+
+      theoreticalPeakMz
+      theoreticalPeakIntensity
+      observedPeakMz
+      observedPeakMzStddev
```
LachlanStuart, line 181: I ran out of time to add these to the UI, but they could be useful in the diagnostics images, perhaps as a popover on the m/zs above each image showing e.g.:

    Theoretical m/z: 623.33587
    Observed m/z: 623.33746 (2.55ppm higher)
    Standard deviation: 0.00031 (0.5ppm)

metaspace/python-client/metaspace/types.py
```diff
@@ -57,7 +57,7 @@ class MetadataMSAnalysis(TypedDict):
     Ionisation_Source: str
     Analyzer: str
     Detector_Resolving_Power: MetadataResolvingPower
-    Pixel_Size: str
+    Pixel_Size: MetadataPixelSize
```
LachlanStuart, line 60: Unrelated fix: This type annotation was wrong due to some bad copy+pasting

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -1347,13 +1348,25 @@ def _get_diagnostic_image(self, image):
...
+                elif format == 'NPY':
+                    image_content = np.load(raw, allow_pickle=False)  # type: ignore
+                elif format == 'JSON':
+                    image_content = json.load(raw)
+                elif format == 'PARQUET':
+                    image_content = pd.read_parquet(raw)
+                else:
+                    print(
+                        f'Warning: Unrecognized image format {format}, returning unparsed content'
+                    )
```
LachlanStuart, line 1365: Unfortunately the old code tries to `np.load` everything that isn't `'PNG'`, which causes errors on new datasets' `'PARQUET'` images.

I changed unrecognized formats to soft failures so that this issue doesn't happen again next time a new format is added.

metaspace/graphql/src/modules/annotation/controller/Annotation.ts
```diff
@@ -58,43 +58,15 @@ const Annotation: FieldResolversFor<Annotation, ESAnnotation | ESAnnotationWithC
...
 
-  adduct: (hit) => hit._source.adduct,
-
-  neutralLoss: (hit) => hit._source.neutral_loss || '',
-
-  chemMod: (hit) => hit._source.chem_mod || '',
-
-  ion: (hit) => hit._source.ion,
-
-  ionFormula: (hit) => hit._source.ion_formula || '', // TODO: Remove ' || ''' after prod has been migrated
```
LachlanStuart, line 69: I've moved these fields into `Query.ts`. There was no reason for them to have their own resolvers - the code is simpler if all these fields are just extracted in the root Query.

docker/sm-webapp/docker-entrypoint.sh
```diff
@@ -19,6 +19,4 @@ yarn install
 npm rebuild node-sass # Ensure the musl version is installed
 yarn run deref-schema
 
-wait_for "nc -z redis 6379" "Redis"
-
```
LachlanStuart, line 23: Unrelated change: Webapp hasn't needed Redis for several years. It used to directly handle auth, but that functionality was all moved into Graphql

LachlanStuart:
I've added the changes to fix the Spark/Lithops discrepancy. The issue was that the `coo_matrix` sparse forms of images had their peaks stored in different orders between the two implementations. Spark used quicksort (an unstable sort) to sort peaks by m/z, whereas Lithops used mergesort (a stable sort). Both using 32-bit m/z. With a stable sort, peaks that have the same m/z are kept in their original order (spectrum index ascending), but with unstable sort they're unpredictably ordered. I fixed this by making Spark use mergesort.

This was enough to make them consistent, but the amount of variance caused by the order-dependent operations indicated that it was also a potential issue with accuracy, so I kept digging.

Many floating point operations have order-dependent results. The weighted `np.average` operation used in `weighted_stddev` turns out to be exceptionally sensitive, with almost every value being different depending on order. To stabilize the outputs, I made this function use 80-bit `longdouble` math. Even with this change it's not 100% order-independent, but the order-dependent error is much smaller.

This wasn't previously detected because the original spectral/spatial/chaos metrics use the non-sparse forms of images, where pixels are in grid order. `total_iso_ints` (calculated from the sparse images) occasionally differed between the implementations because `np.sum` is also order-dependent, but `np.sum` has much less order-dependent error than `np.average` with weights, so we never noticed.

LachlanStuart:
@sergii-mamedov I just pushed the fixes for the two bugs we discussed

metaspace/webapp/src/modules/MetadataEditor/MetadataEditor.vue
```diff
@@ -275,6 +277,7 @@ export default {
         decoySampleSize: isNew ? null : get(config, 'fdr.decoy_sample_size') || null,
         ppm: isNew ? null : get(config, 'image_generation.ppm') || null,
         analysisVersion: isNew ? 1 : get(config, 'analysis_version') || 1,
+        scoringModel: isNew ? null : get(config, 'scoring_model') || null,
```
lmacielvieira, line 280: The scoring model attribute is inside fdr. It should be changed to:
        scoringModel: isNew ? null : get(config, 'fdr.scoring_model') || null,

LachlanStuart:
There's a merge conflict now, but I'll try to wait until after approval before merging in master to minimize disruption.

With `db_mutex.py`, the change in master is better than the change in this branch, so this branch's change will be undone.

lmacielvieira (approved):

sergii-mamedov (approved):

metaspace/engine/docker/lithops_ibm_cf/Dockerfile
```diff
@@ -15,7 +15,7 @@ RUN pip install --no-cache-dir --upgrade pip setuptools wheel \
         tblib==1.7.0
 
 COPY requirements.txt /tmp/requirements.txt
-RUN grep -E "^(c?pyMSpec|c?pyImagingMSpec|numpy|scipy|pandas|msgpack|pyimzML|pyarrow|scikit-learn|Pillow|boto3)" /tmp/requirements.txt > /tmp/requirements-lithops.txt
+RUN grep -E "^(c?pyMSpec|c?pyImagingMSpec|numpy|scipy|pandas|msgpack|pyimzML|pyarrow|scikit-learn|Pillow|boto3|catboost)" /tmp/requirements.txt > /tmp/requirements-lithops.txt
```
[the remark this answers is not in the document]
LachlanStuart: It will only be needed if code is changed to call the `v2_chaos` or `v2_chaos_orig` functions in `metrics.py`. There is no codepath that calls these functions, so `numba` is never imported.

These functions are prototypes that I've left because I ran out of time to prove whether they're better than the original `chaos` metric, but at least according to my intuition they should be better because they handle several failure cases. They'll probably be useful for future development of these metrics.

metaspace/engine/sm/engine/annotation/diagnostics.py
```diff
@@ -176,12 +244,52 @@ def load_npy_image(ds_id: str, image_id: str):
...
+    for attempt in range(3):
+        try:
+            result = fn()
+        except Exception:
+            logger.exception(
+                f'Exception generating {type_} diagnostic after {attempt + 1} attempts',
+                exc_info=True,
+            )
+            result = {'error': format_exc()}
+    return {'ds_id': ds_id, 'type': type_, **result}
```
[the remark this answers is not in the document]
LachlanStuart: This is fixed - there is now a `break` in the `try` section.

PR #1024, opened 2022-01-18
Update sci-test

LachlanStuart:
This change modernizes sci-test and adds new features:
* Update reference results to include FDR and new metrics
* Check FDRs against reference values (with a significant tolerance for error due to FDRs being semi-random)
* Add analysis_version==3 support & install the scoring-model if needed
* Allow CoreMetabolome-v3 to be used instead of HMDB, so that the test is fast enough for CI
* Re-enable sci-test on the master branch
* Fix the image store mocking in Lithops mode (previously was just not implemented)
* If the test didn't pass (or `--save` was passed), save a report comparing new annotations to reference annotations
* Change command line options to have smart defaults:
    * Default to the `config/scitest_config.json` config file
    * Remove `--run` option - always try to run
    * Rename `--save` option to `--save-ref`. Unlike before, it will re-run the test before saving reference results, instead of having to run it twice
    * Add new `--save` option to force it to save the new comparison report even on success
    * Replace `--mock-image-storage` with `--store-images`, swapping the default value. Saving images is slow and they're not checked by the test, so it's better to have it off by default

## Additional changes

* Added the "Untreated tiny" dataset to the repo, just because it's useful for developers and previously I had been sending it via Slack.
* Slightly improved the docker image size for the Lithops CF image
* Tweaked the `setup-dev-env.sh` script

## Validating the new reference results

I ran the following script to verify that there were no significant changes between the old and re-created reference results:
```python
import numpy as np
import pandas as pd

# Need to add "\tmsm" to the first line to make it a valid TSV file before running this
old = pd.read_csv('tests/reports/spheroid_untreated_search_res.csv', sep='\t')
new = pd.read_csv('tests/reports/spheroid-results-v1.csv')
df = old.merge(
    new, on=['formula', 'adduct'], how='outer', suffixes=('_old', '_new'), indicator=True
)
df = df[['formula', 'adduct', *sorted(set(df.columns) - {'formula', 'adduct'})]]
print(df._merge.value_counts())
print('max chaos err:', (df.chaos_new - df.chaos_old).abs().max())
print('max spatial err:', (df.spatial_new - df.spatial_old).abs().max())
print('max spectral err:', (df.spectral_new - df.spectral_old).abs().max())
```

This was the output, indicating no new/missing annotations, and tiny differences to spatial/spectral that are likely explained by getting these values from the `stats` JSON (stored as 64-bit floats) instead of the metric columns (stored as 32-bit floats):
```
both          840
left_only       0
right_only      0
Name: _merge, dtype: int64

max chaos err: 0.0
max spatial err: 1.0205221556702782e-08
max spectral err: 2.683006561365886e-08
```

## CI Performance

To check which `--analysis-version` and `--lithops` option has the best performance for CI, I [ran all 4 combinations](https://app.circleci.com/pipelines/github/metaspace2020/metaspace/2129/workflows/f57cee2b-d124-44fb-8322-8dcc1c5122b9/jobs/18764) and checked the timing & max memory of the "Run Scientific Test" step (all using CoreMetabolome, "large" 4CPU/8GB executor):
* analysis_version==1 Spark: 83s / 3GB
* analysis_version==3 Spark: 37s / 2.9GB
* analysis_version==1 Lithops: 206s / 4.6GB
* analysis_version==3 Lithops: 97s / 5.1GB

"analysis_version==3 Spark" is the fastest, so I've set that as the default. Technically the "large" executor isn't needed with Spark, but the "medium" 2CPU/4GB takes 30s longer and we get plenty of free CircleCI credits.


docker/sm-engine/Dockerfile
```diff
@@ -39,30 +38,23 @@ RUN APT_INSTALL="apt-get install -y --no-install-recommends" && \
...
     rm -rf /var/lib/apt/lists/* /tmp/* ~/.cache/*
 
 # --- Engine dependencies ---
 COPY metaspace/engine/requirements.txt /requirements.txt
 COPY metaspace/engine/requirements-dev.txt /requirements-dev.txt
-RUN APT_INSTALL="apt-get install -y --no-install-recommends" && \
-    PIP_INSTALL="python -m pip install --upgrade" && \
-    apt-get update && \
-    DEBIAN_FRONTEND=noninteractive $APT_INSTALL \
-        git curl postgresql-client libpq-dev && \
```
LachlanStuart, line 60: I moved this apt stuff to the previous step because doing multiple `apt update`s/`apt clean`s in the same build was wasting a lot of time and I had to rebuild this container several times to check what was making the images much larger.

* Numba increased the uncompressed image size by ~100MB (25MB from numba, 75MB from llvmlite), so I just removed it from `requirements.txt` as numba isn't used in production code
* CatBoost increased the uncompressed image size quite substantially because it directly depends on Plotly(128MB) even though it is only used if you call the plot functions. I decided to only mitigate this in the Lithops image, because the code to install CatBoost without dependencies is a bit messy and there are many places where `requirements.txt` is re-installed after image creation (Docker startup script, all of the CircleCI test scripts, etc.)

docker/setup-dev-env.sh
```diff
@@ -7,15 +7,28 @@ cp -n ../metaspace/engine/conf/config.docker.json ../metaspace/engine/conf/confi
...
+echo "Waiting 10 seconds for everything to run init scripts..."
+sleep 10
+
+# set up ElasticSearch
+echo "Creating ElasticSearch index"
+docker-compose run --rm api /sm-engine/create-es-index.sh
+
+# re-run postgres init script in case it didn't run correctly the first time
+echo "Initializing postgres schema (\"already exists\" errors can be safely ignored)"
+docker-compose exec postgres sh /docker-entrypoint-initdb.d/create-sm.sh
```
LachlanStuart, line 22: These changes fix issues that were found while Lucas was setting up on different computers:

* The instructions don't include running `create-es-index.sh`. If the index is autocreated without the creation script, the fields(i.e. columns) don't have their properties (i.e. schema) set up and ElasticSearch just guesses. This can cause hard-to-diagnose differences in behavior compared to correctly set up ElasticSearch DBs.
* If the postgres container fails to run the setup script, it doesn't retry unless you delete the whole PGDATA directory. This forces it to retry.

metaspace/engine/docker/lithops_ibm_cf/Dockerfile
```diff
@@ -15,8 +15,10 @@ RUN pip install --no-cache-dir --upgrade pip setuptools wheel \
         tblib==1.7.0
 
 COPY requirements.txt /tmp/requirements.txt
-RUN grep -E "^(c?pyMSpec|c?pyImagingMSpec|numpy|scipy|pandas|msgpack|pyimzML|pyarrow|scikit-learn|Pillow|boto3|catboost)" /tmp/requirements.txt > /tmp/requirements-lithops.txt
-RUN pip install --no-cache-dir -r /tmp/requirements-lithops.txt
+RUN grep -E "^(c?pyMSpec|c?pyImagingMSpec|numpy|scipy|pandas|msgpack|pyimzML|pyarrow|scikit-learn|Pillow|boto3)" /tmp/requirements.txt > /tmp/requirements-lithops.txt
+# CatBoost installs large optional dependencies (plotly, matplotlib), so install it with --no-deps
+RUN grep -E "^(catboost)" /tmp/requirements.txt > /tmp/requirements-catboost.txt
+RUN pip install --no-cache-dir -r /tmp/requirements-lithops.txt && pip install --no-cache-dir -r /tmp/requirements-catboost.txt --no-deps
```
LachlanStuart, line 21: BTW, the current DockerHub image was generated with this new code so that plotly and matplotlib are excluded. I just didn't want to include it in the ml-scoring branch to minimize updates while it was in review.

metaspace/engine/requirements.txt
```diff
@@ -27,5 +27,5 @@ pyarrow==1.0.1
 pysparkling==0.6.0
 s3fs
 lithops==2.5.8
-numba==0.54.0
-catboost==1.0.3
\ No newline at end of file
+# numba==0.54.0  # Only used by disabled v2 metrics
```
LachlanStuart, line 30: I excluded numba completely just to keep the `sm-engine` image size down. It's still huge(~2GB), but it's ~175MB smaller (uncompressed) without numba.

metaspace/engine/scripts/run_sm_daemon.py
```diff
@@ -43,6 +44,8 @@ def main(daemon_name, exit_after):
     logger.info(f'Starting {daemon_name}-daemon')
 
     conn_pool = ConnectionPool(sm_config['db'])
+    # Ensure the S3 bucket for image storage exists and is configured correctly.
+    image_storage.configure_bucket(sm_config)
```
LachlanStuart, line 48: Before this change, test processes would also try to initialize the image storage bucket, even if it was mocked or not used at all. This was a problem for sci-test, which originally didn't include MinIO. 

I eventually had to include MinIO anyway for diagnostics to be saved, but decided to keep this change because any reduction to the number of calls to `configure_bucket` is good. `configure_bucket` only needs to be called once, and the extra calls have caused issues in the past - conflicts when S3 complains about overlapping calls, etc.

metaspace/engine/sm/engine/annotation_lithops/annotation_job.py
```diff
@@ -307,9 +309,21 @@ def run(self, **kwargs):
...
-            self.db_formula_image_ids = self.pipe.store_images_to_s3(self.ds.id)
+
+            # Save images (if enabled)
+            if self.store_images:
+                self.db_formula_image_ids = self.pipe.store_images_to_s3(self.ds.id)
+            else:
+                self.db_formula_image_ids = {
+                    moldb_id: {formula_i: [None] * n_peaks for formula_i in result_df.index}
+                    for moldb_id, result_df in self.results_dfs.items()
+                }
```
LachlanStuart, line 325: `SearchResults` logs warnings if formula IDs are missing, which is hugely spammy if `self.db_formula_image_ids = {}` is used here. I think these warnings are valuable, so I decided to make empty lists here instead of removing the warnings.

sergii-mamedov (approved):
LGTM

PR #1038, opened 2022-02-02
[DatasetComparisonDialog] UX improvements

lmacielvieira:
### Description

When there are many options (i.e datasets) listed in the select inputs, the popper window position is truncated and hiding the input #1036

#### Solution
A maximum height was set to the select popper. For overflow cases, a scroll behaviour was also added. 
Unrelated: A scroll listener was added so the popper position is updated correctly when the dialog is scrolled in smaller screens.

#### How to test
Access the dataset overview (.../dataset/<ds_id>) page and open the comparison dialog.

###### DatasetComparisonDialog
- [x] Added scroll and max height to select list options
- [x] Added scroll listener to update popper location on small screens
- [x] Appending select poppers to dialog instead to the body

##### Front end
 
- [x] Unit Test
- [ ] e2e Test
 
- [x] Chrome
- [x] Safari
- [x] Firefox
- [x] md, lg, lg
- [x]  sm, xl

Before:
image

After:
image

### DIFF
--- a/metaspace/webapp/src/modules/Datasets/comparison/DatasetComparisonDialog.scss
+++ b/metaspace/webapp/src/modules/Datasets/comparison/DatasetComparisonDialog.scss
@@ -38,3 +38,18 @@
 }
 
 
+.ds-options-popper {
+  max-width: 30%;
+
+  .el-select-dropdown__wrap {
+    max-height: 186px !important;
+  }
+}
+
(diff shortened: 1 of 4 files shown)


sergii-mamedov (approved):
It seems to me that the best option would be the input field on the next line after the selected datasets.

PR #1056, opened 2022-02-16
Removing unused metrics

sergii-mamedov:


LachlanStuart (approved):
LGTM. The sci-test in CI will fail if this is merged before #1057, but will pass again after both are merged.

Issue #1070, opened 2022-03-07
Allow users to see/reprocess failed datasets


theodev:
Now the user does not see his datasets in the FAILED status on the website. I checked that these datasets are visible through the python-client and that there is full access to them. This front-end problem.

PR #1076, opened 2022-03-18
Enable reprocess for common user

lmacielvieira:
### Description

Currently when a dataset fails to process, the user can't see the dataset anymore. In order to make if more friendly and let the user know the dataset failed processing, now the user should be able to reprocess and see the failed dataset #1070 

In the dataset page listing if the user selects the categories, the pagination does not reflect only the datasets from the selected category. So in order to make it easier, the category is also considered as a dataset listing filter.

#### Solution
The FAILED category section is been displayed for common users also, if they have at least one failed dataset. The rules applied for the failed dataset reprocessing are the same as the edition one: user submitted the dataset or it in the group in which the dataset belongs. To allow the listing of only the failed datasets that the user can also edit, a change in the elasticsearch common dataset query was made, where it can only see the public datasets that do not failed processing plus the datasets the user is allowed to see (included the failed ones).

#### How to test
Upload a dataset that will fail, access the dataset page and test it as: not logged, admin, user, user in the group of the failed dataset and user without failed datasets

###### datasets api
- [x] Getting limit parameter from variable

###### DatasetItemActions
- [x] Showing reprocess if user can edit dataset and dataset has failed processing

###### DatasetTable
- [x] Counting pagination total on filter change
- [x] Allowing common user to see failed tab if it has one failed dataset
- [x] Added selected categories as dataset filter

###### datasetFilters
- [x] Changed status filter to accept multiple values

###### esConnector
- [x] Changed list datasetQuery to return only non failed public datasets

##### Front end
 
- [ ] Unit Test
- [ ] e2e Test
 
- [x] Chrome
- [x] Safari
- [x] Firefox
- [x] md, lg, lg
- [x]  sm, xl

### DIFF
--- a/metaspace/graphql/datasetFilters.js
+++ b/metaspace/graphql/datasetFilters.js
@@ -86,6 +86,27 @@ class DatasetIdFilter extends AbstractDatasetFilter {
   }
 }
 
+class DatasetStatusFilter extends AbstractDatasetFilter {
+  constructor() {
+    super('', {})
+  }
+
+  esFilter(status) {
+    status = status.split('|')
+
+    if (status.length > 0) {
+      return { terms: { ds_status: status } }
+    } else {
+      return {}
+    }
+  }
+
+  pgFilter(q, status) {
+    status = status.split('|')
+    return q.whereIn('status', status)
+  }
+}
+
 class NotNullFilter extends AbstractDatasetFilter {
   constructor(schemaPath, options = {}) {
     super(schemaPath, options)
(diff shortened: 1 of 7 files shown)


sergii-mamedov (approved):
Did I understand correctly that only his and his FAILED datasets will be displayed for the user? Or will public FAILED datasets also be visible?

lmacielvieira:
> Did I understand correctly that only his and his FAILED datasets will be displayed for the user? Or will public FAILED datasets also be visible?

Yes, you understand correctly. Only his, his FAILED datasets or his group dataset can be reprocessed.

PR #1097, opened 2022-04-13
Showing  only valid mails on submitter area

lmacielvieira:
### Description

It seems that some old datasets can have submitters with invalid mail fields and submitter data not structured to be displayed in the metadata viewer. So, these cases were treated so It does not look as a bug in the dataset overview page

image

### DIFF
--- a/metaspace/webapp/src/modules/Datasets/overview/DatasetOverviewPage.tsx
+++ b/metaspace/webapp/src/modules/Datasets/overview/DatasetOverviewPage.tsx
@@ -58,13 +58,19 @@ export default defineComponent<Props>({
       result: currentUserResult,
       loading: userLoading,
     } = useQuery<CurrentUserRoleResult|any>(currentUserRoleQuery)
-    const currentUser = computed(() => currentUserResult.value != null ? currentUserResult.value.currentUser : null)
+    const currentUser = computed(() => currentUserResult.value != null ? currentUserResult.value.currentUser
+      : null)
 
     const projectLink = (projectIdOrSlug: string) => {
       return ({
         name: 'project',
         params: { projectIdOrSlug },
       })
+    }
+
+    const isValidMail = (email:string) => {
+      const re = /^\w+@[a-zA-Z_]+?\.[a-zA-Z]{2,3}$/
+      return re.test(String(email).toLowerCase())
     }
 
     return () => {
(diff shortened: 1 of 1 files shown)


metaspace/webapp/src/modules/Datasets/overview/DatasetOverviewPage.tsx
```diff
@@ -67,6 +68,11 @@ export default defineComponent<Props>({
       })
     }
 
+    const isValidMail = (email:string) => {
+      const re = /^\w+@[a-zA-Z_]+?\.[a-zA-Z]{2,3}$/
```
sergii-mamedov, line 72: I would not recommend parsing e-mail with regular expressions. Is to open the gates to hell
For example, you cannot catch such an e-mail with this regex: test@mail.google.com

metaspace/webapp/src/modules/Datasets/overview/DatasetOverviewPage.tsx
```diff
@@ -131,7 +139,7 @@ export default defineComponent<Props>({
             <div class='dataset-overview-holder'>
               <p class='truncate'>{submitter?.name}
                 {group && <a class='ml-1' href={groupLink}>({group?.shortName})</a>}
-                {!group && submitter?.email
+                {!group && submitter?.email && isValidMail(submitter?.email)
```
sergii-mamedov, line 142: As far as I understand, you check if the user has a group. If it's not there, then do you show the e-mail? If so, then I don't understand why. I would simply not show anything for users without a group.

sergii-mamedov (changes requested):

metaspace/webapp/src/modules/Datasets/overview/DatasetOverviewPage.tsx
lmacielvieira: I think the email should be displayed, as maybe some user want to contact the submitter regarding the dataset. As if the submitter is not in a group, there would be no way to get it, if not displayed.

metaspace/webapp/src/modules/Datasets/overview/DatasetOverviewPage.tsx
lmacielvieira: There is a version to check if with all the rules, it is a little bit big though. I just added this to simple ones, as it is already checked the correct way when registering a dataset. This is for the edge cases that we saw on staging where the id is saved on the mail field (probably prior to checking if email is valid on dataset upload)

lmacielvieira: also, by the rules, this is a valid mail. It can have up to three fields after @. I.e brazilian mail lucas@uol.com.br (https://help.returnpath.com/hc/en-us/articles/220560587-What-are-the-rules-for-email-address-syntax-)

sergii-mamedov: this is valid mail, but this regex `^\w+@[a-zA-Z_]+?\.[a-zA-Z]{2,3}$` won't catch him

metaspace/webapp/src/modules/Datasets/overview/DatasetOverviewPage.tsx
lmacielvieira: Email is not being displayed anymore for now. It also solves the last comment

sergii-mamedov (approved):

PR #1109, opened 2022-05-08
imzML browser backend

sergii-mamedov:
### Create S3 buckets sm-imzml-browser-ENV (staging and prod only)
1. Set bucket policy:
   ```
    {
        "Version": "2012-10-17",
        "Statement": [
            {
                "Effect": "Allow",
                "Principal": {
                    "AWS": "arn:aws:iam::Account_ID:user/sm-engine"
                },
                "Action": "s3:*",
                "Resource": "arn:aws:s3:::sm-imzml-browser-ENG/*"
            }
        ]
    }
   ```
2. Set lifecycle rules:
   * Move current versions of objects between storage classes
   * Delete expired object delete markers or incomplete multipart uploads 

### Add to the `load_ds` step saving mz sorted data on S3
Due to the fact that `s3_client` is not serialized inside the Lithops executor, saving the files needed for the imzml browser had to be split into two steps:
1. Save files on COS. Files are automatically deleted within a day. (`_upload_imzml_browser_files_to_cos`)
2. Move files from COS to S3 (`_upload_imzml_browser_files`)

## How to test
1. Run METASPACE.
2. Run a request and save a PNG file:
   ```
   curl -X POST http://localhost:5123/v1/browser/intensity_by_mz -H "Content-Type: application/json" -d '{"ds_id": "ds_id", "mz_low": 100, "mz_high": 500}' > z.png
   ```
3. Run a request to get the spectrum:
    ```
    curl -X POST http://localhost:5123/v1/browser/peaks_from_pixel -H "Content-Type: application/json" -d '{"ds_id": "ds_id", "x": 58, "y": 0}'
    ```

### Necessary improvements:
1. Make a check for the presence of files to prevent re-moving between clouds.


lmacielvieira (approved):
LGTM.
The code itself looks good. If possible just add the performance checks on the reading data routes, as we discussed on slack.

lmacielvieira (approved):
LGTM. The processing and file generation on load_ds looks is faster. The problems I was having with performance before did not happen. In the future it would be nice to profile the image generation performance to make sure it is optimized.

lmacielvieira (approved):
LGTM

PR #1127, opened 2022-06-22
Dataset Comparison ShareLink with filter and multichannel bug

lmacielvieira:
### Description

After creating a link on Dataset Comparison page with multichannel selected ions, which are not anymore displayed a bug where the ion images are not displayed happens. In order to fix it, if the ion channel is not saved on the snapshot, it is not displayed anymore if accessed from a shared link #1120.

#### How to test
1 - Access the dataset comparison page 
2 - Activate multichannel
3 - Select a annotation, Add another annotation
4 - Filter by annotation name (i.e glucose, or any annotation name) where at least one of the selected annotations disappear
5 - Create a permalink
6 - Access ds comparsion from generated permalink

###### DatasetComparisonGrid
- [x] Check if channel info exist before adding as empty
- [x] Added verifications to correct render even with empty data

###### DatasetComparisonPage
- [x] Added verifications to correct render even with empty data

###### DatasetShareLink
- [x] Added verifications to correct render even with empty data

##### Front end
 
- [x] Unit Test
- [ ] e2e Test
 
- [x] Chrome
- [x] Safari
- [x] Firefox
- [x] md, lg, lg
- [x]  sm, xl

Example before fix:
image

### DIFF
--- a/metaspace/webapp/src/modules/Datasets/comparison/DatasetComparisonGrid.tsx
+++ b/metaspace/webapp/src/modules/Datasets/comparison/DatasetComparisonGrid.tsx
@@ -205,7 +205,7 @@ export const DatasetComparisonGrid = defineComponent<DatasetComparisonGridProps>
 
         if (idx !== -1) {
           annotations.push(channel.annotations.annotations[idx])
-        } else if (channel.id) {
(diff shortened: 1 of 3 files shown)


sergii-mamedov (approved):
LGTM

Issue #1136, opened 2022-07-07
[Upload] Unflag ppm field


lmacielvieira:
- [ ] Unflag 'only' the ppm field on upload page
- [ ] Add a maximun of 50 ppm to ppm field
- [ ] Recommend the usage of 10 ppm if 'resolving power' is bellow 7000
- [ ] Add reminder about isobar ambiguity to user

PR #1150, opened 2022-07-18
Enabled ppm field

lmacielvieira:
### Description

Enabled ppm input for dataset upload and edition #1136    

##### How to test

1 - Access the upload page: <metaspace_url>/upload
2 - Access dataset edition page: <metaspace_url>/datasets/edit/<ds_id>

###### MetadataEditor
- [x] Removing null and undefined to merge with defaultMetaspaceOptions
- [x] Changed ppm to required field
- [x] Set default ppm to 3
- [x] Added ppm help tooltip
- [x] Set ppm = 10 if resolving power less than 7 thousand

###### PpmHelp
- [x] Created ppm tooltip component

###### MetaspaceOptionsSection
- [x] Enabled ppm field without feature flag

##### Front end
 
- [x] Unit Test
- [ ] e2e Test
 
- [x] Chrome
- [x] Safari
- [ ]  Firefox
- [x] md, lg, lg
- [x]  sm, xl

image

### DIFF
--- a/metaspace/webapp/src/modules/MetadataEditor/MetadataEditor.vue
+++ b/metaspace/webapp/src/modules/MetadataEditor/MetadataEditor.vue
@@ -124,10 +124,10 @@
 import { defaultMetadataType, metadataSchemas } from '../../lib/metadataRegistry'
 import { deriveFullSchema } from './formStructure'
 import {
-  get, set, cloneDeep, defaults,
+  get, set, cloneDeep, forOwn, defaultTo, defaults,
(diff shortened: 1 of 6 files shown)


sergii-mamedov (approved):
LGTM, but:

1. When I set the resolving power to less than 7000, the ppm value is set to 10. But if I increase the resolving power after that, the ppm value does not change.
2. It seems to me that when discussing the problem, we decided that ppm values can only be integers. Or I'm wrong?

sergii-mamedov (approved):

Issue #1186, opened 2022-08-31
Stop processing in case of a problem with imzML file


sergii-mamedov:
In the event of an error parsing an imzML file, there is no point in trying again using Spark. Need to add logic that will track `ImzMLError` and terminate processing in advance.

Issue #769, opened 2021-02-25
Lithops: Fix custom error messages for ImzML parse errors


LachlanStuart:
The Spark implementation rethrew ImzMLParser exceptions as `ImzMLError` exceptions, which resulted in a more helpful message in the annotation failure email. This logic should be ported to the Lithops code

PR #1187, opened 2022-09-01
Early stop processing if imzML file is brocken & Fix error message for lithops

sergii-mamedov:
This PR fixed two issue:
* #769 
* #1186


lmacielvieira (commented):
LGTM. I was able to process the dataset even by adding a random string in the imzml file. Maybe it is because it was not on closing tags

Issue #1169, opened 2022-08-08
[ROI] Store ROI information


lmacielvieira:
Save ROI information per dataset (like optical images)

PR #1199, opened 2022-09-28
Saving ROIs information

lmacielvieira:
### Description

Save ROI information per dataset (like optical images) #1169 

### How to test
 1 - Open annotation page
 2 - Create ROIs
 3 - Press the 'save ROI' button
 4 - Reload the page
 5 - Show ROIs

image

### DIFF
--- a/metaspace/webapp/src/api/annotation.ts
+++ b/metaspace/webapp/src/api/annotation.ts
@@ -29,6 +29,7 @@ gql`fragment AnnotationDetailItem on Annotation {
     metadataJson
     configJson
     isPublic
+    canEdit
     opticalImages(type: $type) {
       id
       url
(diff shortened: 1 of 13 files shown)


sergii-mamedov:
@lmacielvieira please resolve the conflict

metaspace/webapp/src/api/annotation.ts
```diff
@@ -29,6 +29,7 @@ gql`fragment AnnotationDetailItem on Annotation {
     metadataJson
     configJson
     isPublic
+    canEdit
```
sergii-mamedov, line 32: what is it?

lmacielvieira: I added the dataset 'canEdit' attribute to the annotation query return on graphql, so we can know if the user can edit the dataset. If the user can edit the dataset, we allow it to store the ROI information, if not, we do not. This is needed before pressing the 'save ROI', in order to display the save icon as enabled/disabled before pressing it.

sergii-mamedov (approved):

PR #1204, opened 2022-10-11
Switch to another executor if MemoryError happens

sergii-mamedov:
1. Changed the logic of switching between lithops executors (see description in the issue).
2. Add more info in logging.

### DIFF
--- a/metaspace/engine/sm/engine/annotation_lithops/executor.py
+++ b/metaspace/engine/sm/engine/annotation_lithops/executor.py
@@ -269,7 +269,7 @@ class Executor:
                 and (max_memory is None or runtime_memory < max_memory)
             ):
                 old_memory = runtime_memory
-                runtime_memory *= 2
+                runtime_memory *= 2 if old_memory < 32 * 1024 else 128 * 1024
                 attempt += 1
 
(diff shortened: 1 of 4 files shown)


metaspace/engine/sm/engine/annotation_lithops/executor.py
```diff
@@ -265,11 +265,11 @@ def map(
 
             if (
                 isinstance(exc, (MemoryError, TimeoutError, OSError))
-                and runtime_memory < 32 * 1024
+                and runtime_memory <= 32 * 1024
                 and (max_memory is None or runtime_memory < max_memory)
             ):
                 old_memory = runtime_memory
-                runtime_memory *= 2
+                runtime_memory *= 2 if old_memory < 32 * 1024 else 128 * 1024
```
lmacielvieira, line 272: Do you really want to multiply the runtime_memory times 128 * 1024 if you want to run on vpv? It should not be
`if old_memory < 32 * 1024:
   runtime_memory *= 2
  else:
   runtime_memory = 128 * 1024
`?
As the vpc size is 128 * 1024?

lmacielvieira (approved):

PR #1208, opened 2022-10-12
Added seriate sorting to spotting axis

lmacielvieira:
### Description

Added seriate sorting to spotting axis, updated interlab file and moved spotting to dockerized lambda #1200


metaspace/engine/lambdas/spotting/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM  --platform=linux/amd64 public.ecr.aws/lambda/python:3.8
```
sergii-mamedov, line 1: I would use the official python image that we use everywhere

metaspace/engine/lambdas/spotting/app.py
```diff
@@ -1,10 +1,38 @@
...
+    """
+    matrix = np.zeros((len(data[y_axis].unique()), len(data[x_axis].unique())))
+    for row_idx, row in enumerate(data[y_axis].unique()):
+        for col_idx, col in enumerate(data[x_axis].unique()):
+            try:
+                fraction = data[(data[y_axis] == row) & (data[x_axis] == col)][
+                    'fraction_detected'
+                ].values[0]
+            except:
+                fraction = 1
```
sergii-mamedov, line 28: I will be change on this:
```
df = data[(data[y_axis] == row) & (data[x_axis] == col)]['fraction_detected']
fraction = df.values[0] if df.size else 1  # df.size return a number

```

sergii-mamedov (changes requested):
1. Please add a bash script for creating and uploading docker image on AWS. Example is [here](https://github.com/metaspace2020/metaspace/blob/master/metaspace/engine/docker/lithops_ibm_ce/build_and_push.sh).
2. Please remove or uncomment all commented line.
3. I think it would be good not to use a separate requirements.txt but to do as [here](https://github.com/metaspace2020/metaspace/blob/master/metaspace/engine/docker/lithops_ibm_ce/Dockerfile#L34-L35). But it's up to you.

metaspace/engine/lambdas/spotting/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+# Copy function code
+COPY app.py ${LAMBDA_TASK_ROOT}
+
+# Install the function's dependencies using file requirements.txt
+# from your project folder.
+
+ENV PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python
```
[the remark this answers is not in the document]
lmacielvieira: The seriate sorting use a dependence that needs to downgrade protobuf and only works with this env variable value

sergii-mamedov (approved):

Issue #1196, opened 2022-09-23
[DatabaseUpload] Add parser importing the custom molecular db


lmacielvieira:
Add parser when importing custom database so molecular names/id/inchi can not be bigger than the allowed size of the field

PR #1212, opened 2022-10-13
Checking value length when importing custom db

lmacielvieira:
### Description

Add parser when importing custom database so molecular names/id/inchi can not be greater than the allowed size (2500 characters, which was based on the longer value we have that is of 2100 characters) of the field #1196

### DIFF
--- a/metaspace/engine/sm/engine/molecular_db.py
+++ b/metaspace/engine/sm/engine/molecular_db.py
@@ -63,11 +63,20 @@ class MolecularDB:
 
 def _validate_moldb_df(df):
     errors = []
+    max_value_length = 500
     for idx, row in df.iterrows():
         line_n = idx + 2
         for col in df.columns:
             if not row[col] or row[col].isspace():
                 errors.append({'line': line_n, 'row': row.values.tolist(), 'error': 'Empty value'})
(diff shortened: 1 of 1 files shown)


metaspace/engine/sm/engine/molecular_db.py
```diff
@@ -63,11 +63,20 @@ def to_dict(self):
...
     for idx, row in df.iterrows():
         line_n = idx + 2
         for col in df.columns:
             if not row[col] or row[col].isspace():
                 errors.append({'line': line_n, 'row': row.values.tolist(), 'error': 'Empty value'})
+            elif len(row[col]) > max_value_length:
+                errors.append(
+                    {
+                        'line': line_n,
+                        'row': [row[col]],
```
sergii-mamedov, line 76: Maybe put not cell by row? As above: `row.values.tolist()`

sergii-mamedov (approved):

metaspace/engine/sm/engine/molecular_db.py
lmacielvieira: because it will repeat the line erros as above, because is interating each col, but adding as a line error.

PR #1234, opened 2022-11-03
Add size and hash of imzML/ibd files to PostgreSQL

sergii-mamedov:
I am not very satisfied with the architectural solution, but I could not think of a better one.

In `load_ds.py` I added md5 hash calculation for imzML/ibd files and getting their sizes. I return data in the form of a dict to `pipeline.py`, where there is a function for saving in PostgreSQL.


metaspace/engine/sm/engine/annotation_lithops/load_ds.py
```diff
@@ -220,7 +269,7 @@ def load_ds(
     # most memory-intense part (sorting the m/z array). Also for uploading imzml browser files
     # need plus 1x the ibd file size RAM.
     message = f'Found {ibd_size_mb}MB .ibd and {imzml_size_mb}MB .imzML files.'
-    if ibd_size_mb * 3 + 512 < 32 * 1024:
+    if ibd_size_mb * 4 + 512 < 32 * 1024:
```
sergii-mamedov, line 272: Forgot change in this PR: #1204

lmacielvieira (approved):

PR #1272, opened 2022-12-09
Add/remove datasets to project pagination bug fix

lmacielvieira:
### Description

Add/remove datasets to project pagination bug fix. Showing number of removed and added datasets in the dialog, so the user do not remove them by mistake. Added log of which datasets were removed and added to the project  #1271 

### Problem
When a project has more than the page number from the add dataset to project component, some times a bug removes the datasets not being displayed in the table from the project, if the user remove a project from the first page, what removes all the datasets not displayed in the table.


metaspace/graphql/src/modules/project/controller/Mutation.ts
```diff
@@ -243,6 +244,13 @@ const MutationResolvers: FieldResolversFor<Mutation, void> = {
         }
       }))
 
+      if (datasetIds && datasetIds.length > 0) {
+        logger.info(`Datasets ${datasetIds} added to ${projectId}`)
```
sergii-mamedov, line 248: I would also add information about who exactly (user_id) added or removed datasets from the project.

sergii-mamedov (approved):
LGTM

Issue #1240, opened 2022-11-07
Problems uploading files via Uppy


sergii-mamedov:
## Description
We have an agreement that there can be only two files in each bucket directory: imzML and ibd. Unfortunately, sometimes this is not the case and this contributes to problems with file processing. For this year's files, we have:
```
num. of files | num. of directories
-----------------------------------
           1  |                196
           3  |                  2
           4  |                  9
           6  |                  1
```
We also have one directory each with 8 and 20 files of 2020 year.

## Reproducing the problem
I was able to reproduce the problem of having 3 or more files:
- Upload imzML and ibd file through the web interface.
- Delete any of them, for example the ibd file.
- Upload the ibd file with a different name.
- There will be three files in the directory.

Similarly, you can select two ibd files, both will be uploaded to S3, but only one will be active in the web interface.

One file in the directory appears in the case when two files started to upload at the same time. The smaller file was uploaded successfully, but there was a problem while uploading the larger one.

## Steps to resolve this issue 
 - [ ] Add functionality that makes it impossible to upload two or more files with the same extension
 - [ ] Add functionality that does not upload files until the `Submit` button is pressed.
 - [ ] Deal with Uppy logging and remove `companion.client.version uppy client version 1.0.0` lines from sm-graphql.log
 - [ ] Add to the log information about who downloaded the file: `User <user_id> uploaded <file_name_or_path>`
 - [x] Delete all directories where one file is located
 - [x] Deal with directories where the number of files is more than three.
 
 @lmacielvieira The first four tasks are yours, the last two are mine.

PR #1286, opened 2023-01-16
Feat/uppy behaviour improvement

lmacielvieira:
### Description

Updated uppy version to remove extra logging and added user adding files log #1240


sergii-mamedov (changes requested):
I tested on staging and everything seems to work. But in the logs I see the following:
```
companion: 2023-03-09T10:16:49.996Z [warn] startup.uploadUrls Running without uploadUrls specified is a security risk if running in production
companion: 2023-03-09T10:16:49.997Z [info] jobs.cleanup.start starting clean up job

(node:2066641) [DEP0097] DeprecationWarning: Using a domain property in MakeCallback is deprecated. Use the async_context variant of MakeCallback or the AsyncResource class instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
```
You also need to update the python-client version in [this file](https://github.com/metaspace2020/metaspace/blob/master/metaspace/python-client/metaspace/__init__.py).

sergii-mamedov (approved):

Issue #1297, opened 2023-02-06
[Spotting] Rename app to “standards”


lmacielvieira:
change spotting app to http://metaspace2020.eu/standards?

- [x] Prepare to realease
- [x] Add 'Download raw files' feature to the application
- [x] Changes 12/12

PR #1304, opened 2023-02-10
Spotting improvements pre-release

lmacielvieira:
### Description

Changed 'Spotting' app to 'Standards' and applied changes requested in attached document #1297  

#### how to test it:
access <metaspace_url>/standards


metaspace/webapp/src/modules/SpottingProject/DashboardPage.tsx
```diff
@@ -799,8 +814,9 @@ export default defineComponent({
 
     const handleItemClick = (item: any) => { // get info from clicked chart item and open on a metaspace url
       const baseUrl = 'https://metaspace2020.eu/annotations?db_id=304'
```
sergii-mamedov, line 816: Why this mol_db is default? Will it work on staging?

sergii-mamedov (approved):
LGTM

metaspace/webapp/src/modules/SpottingProject/DashboardPage.tsx
lmacielvieira: No, it will work only on METASPACE. This is the spotting database, so the results are just for this database, and will work once it is made public after the publication.

Issue #1307, opened 2023-02-13
Add METASPACE stats


lmacielvieira:
- [ ] Propose designs
- [ ] User serpapi to show publications
- [ ] Add these two funding links to METASPACE landing page bottom, continuing the European Union links?
[101092644](https://cordis.europa.eu/project/id/101092644)
[101092646](https://cordis.europa.eu/project/id/101092646)

- [ ] Replace “European Union Horizon 2020 Programme” with “European Union Horizon2020 and Horizon Europe Programs”

Add to METASPACE stats on the usage:
## of groups
TBs of data

PR #1333, opened 2023-04-14
Populate filesize of imzML/ibd files

sergii-mamedov:
This PR is a part of #1307 

This script gets the size of the imzML/ibd files for the specified datasets and writes the result to the `dataset` table.

### DIFF
--- a/metaspace/engine/scripts/populate_filesize.py
+++ b/metaspace/engine/scripts/populate_filesize.py
@@ -1,0 +1,112 @@
+import argparse
+import json
+import logging
+
+import boto3
+import pandas as pd
+
+from sm.engine.db import DB
+from sm.engine.util import GlobalInit
+
+
+def update_size_hash(df):
+
+    db = DB()
+    for i in df.iterrows():
+        ds_id, _, imzml_size, ibd_size = i[1]
+        try:
+            size = {'imml_size': imzml_size, 'ibd_size': ibd_size}
+            db.alter(
+                'UPDATE public.dataset SET size_hash=%s WHERE id=%s',
+                params=(
+                    json.dumps(size),
+                    ds_id,
+                ),
+            )
+            logger.info(f'Updated size and hash for {ds_id}')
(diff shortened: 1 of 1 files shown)


metaspace/engine/scripts/populate_filesize.py
```diff
@@ -0,0 +1,112 @@
...
+        logger.info(f'Got {len(datasets)} datasets')
+        return data
+
+    except Exception:
+        logger.error('Failed to get datasets', exc_info=True)
+        return []
+
+
+def get_all_files(bucket):
+    s3 = boto3.client('s3')
```
lmacielvieira, line 54: In case you need to run it on docker or local env, I recommend changing it to 
`s3 = get_s3_client(sm_config)`

metaspace/engine/scripts/populate_filesize.py
```diff
@@ -0,0 +1,112 @@
...
+        'e.g. "status = \'FINISHED\' and ion_thumbnail is null"',
+    )
+    args = parser.parse_args()
+    logger = logging.getLogger('engine')
+
+    if args.sql_where:
+        with GlobalInit(config_path='conf/config.json') as sm_config:
+            datasets = pd.DataFrame(get_datasets(sql_where=args.sql_where))
+
+        sizes = pd.DataFrame(get_all_files(bucket='sm-engine-upload'))
```
lmacielvieira, line 101: To run in local env or docker, you can get the bucket from `sm_config`

lmacielvieira (approved):

metaspace/engine/scripts/populate_filesize.py
sergii-mamedov: got it

metaspace/engine/scripts/populate_filesize.py
sergii-mamedov: we do not have this info in `sm-config`

Issue #1330, opened 2023-04-04
[SummaryPage] Update Analyzer charts


lmacielvieira:
Update analyzer/ion source/matrix chart to an iteractive one, like the one used on spotting paper

PR #1347, opened 2023-04-27
Changed analyzer/ion source/matrix chart

lmacielvieira:
### Description

Created a new component for the 'Number of datasets per analyzer/ion source/matrix' chart #1330 

#### How to test it

Access https://metaspace2020.eu/datasets/summary

image
image
image


sergii-mamedov (approved):
Wow!! It looks really cool and works much faster.
The only thing that surprised me were the labels above the wheels. I would rotate them 45 degrees from the horizontal. Or at least 180 that could be read from the bottom to the mountain.

Issue #1335, opened 2023-04-17
[Metadata] Show image size information on Metadata Details


lmacielvieira:
- [ ] Show n of pixels
- [ ] Show size of ibd
~- [ ] Show size of images~

sergii-mamedov:
Now the information is available in the `dataset` table. Maybe it's worth pushing this information to ES as well?

PR #1351, opened 2023-05-04
Show file size and pixel count on dataset overview

lmacielvieira:
### Description

Show file size and pixel count on dataset overview page #1335  

#### How to test it

Access https://metaspace2020.eu/dataset/<ds_id>

image


metaspace/engine/sm/engine/es_export.py
```diff
@@ -289,6 +293,8 @@ def sync_dataset(self, ds_id):
         """Warning: This will wait till ES index/update is completed"""
         with self._ds_locker.lock(ds_id):
             ds = self._select_ds_by_id(ds_id)
+            print(f'Exporting dataset {ds_id} to ES')
```
sergii-mamedov, line 296: I think it is not needed more.

Issue #1334, opened 2023-04-17
[Hihgh Masses] Calculate and show monoisotopic m/z on annotations


lmacielvieira:
- [ ] Update m/z value to old dataasets
- [x] Show monoisotopic m/z on annotation table
- [x] Show monoisotopic m/z on summary
- [x] Indicate on annotation image if non mono isotopic m/z is being displayed

*Check with Mans the precision needed to check if monoisotopic != non monoisotopic

PR #1388, opened 2023-07-05
Feat/mono mass

lmacielvieira:
### Description

Using mono-isotopic mz as the primary value for mz and showing a disclaimer to ion images that do not display mono-isotopic mz #1334 
image


sergii-mamedov (approved):
It would be good for the future if you would describe in detail the problem we faced and how exactly you solved it.

PR #1411, opened 2023-08-21
Feat/opt image s3

lmacielvieira:
### Description

Changed optical image storing to s3 #803

### Instructions

To copy the old optical images from local storage to s3, run:
1 - cd /opt/dev/metaspace/metaspace/engine
2 - python scripts/update_raw_optical_images.py

### DIFF
--- a/metaspace/graphql/src/modules/webServer/opticalImageUploadMiddleware.ts
+++ b/metaspace/graphql/src/modules/webServer/opticalImageUploadMiddleware.ts
@@ -1,0 +1,25 @@
+import * as http from 'http'
+import * as express from 'express'
+import * as companion from '@uppy/companion'
+import * as uuid from 'uuid'
+import * as bodyParser from 'body-parser'
+
+import getCompanionOptions from './getCompanionOptions'
+
+export default function(httpServer: http.Server) {
+  const options = getCompanionOptions(
+    '/raw_opt_upload',
+    (req: express.Request) => {
+      const datasetId = req.body?.metadata?.datasetId
+      const file_id = req.body?.metadata?.uuid || uuid.v4()
+      return `raw_optical/${datasetId}/${file_id}`
+    },
+    true
+  )
+
+  const router = express.Router()
+  router.use(bodyParser.json({ limit: '1MB' }))
+  router.use(companion.app(options))
+  companion.socket(httpServer, options)
+  return router
+}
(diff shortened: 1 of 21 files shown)


metaspace/graphql/src/modules/webServer/opticalImageUploadMiddleware.ts
```diff
@@ -0,0 +1,25 @@
...
+    (req: express.Request) => {
+      const datasetId = req.body?.metadata?.datasetId
+      const file_id = req.body?.metadata?.uuid || uuid.v4()
+      return `raw_optical/${datasetId}/${file_id}`
+    },
+    true
+  )
+
+  const router = express.Router()
+  router.use(bodyParser.json({ limit: '1MB' }))
```
sergii-mamedov, line 21: Check this 1 MB restriction.

metaspace/webapp/src/modules/ImageAlignment/ImageAlignmentPage.vue
```diff
@@ -350,9 +362,22 @@ export default {
...
+    },
   },
 
   computed: {
+    uuid() {
+      return this.storageKey.uuid
+    },
+
+    uploadEndpoint() {
+      return `${window.location.origin}/dataset_upload`
```
sergii-mamedov, line 378: why it is `dataset_upload`?

metaspace/webapp/src/modules/ImageAlignment/ImageAlignmentPage.vue
```diff
@@ -397,6 +422,55 @@ export default {
...
 
+  mounted() {
+    this.fetchStorageKey()
+    const uppy = Uppy({
+      debug: true,
+      autoProceed: true,
+      allowMultipleUploads: false,
+      restrictions: {
+        maxFileSize: this.limitMB * 1024 * 1024,
+        maxNumberOfFiles: 2,
```
sergii-mamedov, line 433: why do we allow 2 files to be uploaded?

sergii-mamedov (approved):

metaspace/graphql/src/modules/webServer/opticalImageUploadMiddleware.ts
lmacielvieira: This does not limit the size of incoming images or any other types of data. The setting specifically applies to requests that have the Content-Type header set to application/json.

metaspace/webapp/src/modules/ImageAlignment/ImageAlignmentPage.vue
lmacielvieira: Because I am reusing the `${window.location.origin}/dataset_upload/s3/uuid` to get an unique id

PR #1415, opened 2023-08-31
Handling deleted dbs on dataset reprocessing

lmacielvieira:
### Description

Checking if db exists before starting annotation process, as deleted dbs were triggering errors if not checked #1406

## How to test it
1 - Annotate dataset with a custom db
2 - Delete custom db
3 - Reprocess dataset

### DIFF
--- a/metaspace/engine/sm/engine/annotation_lithops/annotation_job.py
+++ b/metaspace/engine/sm/engine/annotation_lithops/annotation_job.py
@@ -321,12 +321,22 @@ class ServerAnnotationJob:
         # TODO: Only run missing moldbs
         del_jobs(self.ds)
         moldb_to_job_map = {}
+        moldb_to_be_removed = []
+
         for moldb_id in self.ds.config['database_ids']:
             try:
                 molecular_db.find_by_id(moldb_id)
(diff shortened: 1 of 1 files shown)


metaspace/engine/sm/engine/annotation_lithops/annotation_job.py
```diff
@@ -316,7 +322,11 @@ def run(self, **kwargs):
         del_jobs(self.ds)
         moldb_to_job_map = {}
         for moldb_id in self.ds.config['database_ids']:
-            moldb_to_job_map[moldb_id] = insert_running_job(self.ds.id, moldb_id)
+            try:
+                molecular_db.find_by_id(moldb_id)
+                moldb_to_job_map[moldb_id] = insert_running_job(self.ds.id, moldb_id)
+            except Exception:  # db does not exist, continue to next
+                continue
```
sergii-mamedov, line 329: Add logging information about db_id with `warning` level.
This will allow us to analyze possible problems after the fact.

sergii-mamedov (approved):

sergii-mamedov:
Maybe it makes sense to delete the previously deleted mol_db from self.ds.config['database_ids']? 
Because otherwise we will accumulate garbage.

Issue #1410, opened 2023-08-16
[OpticalImage] Add function to python API on how to add optical image and check transform info


lmacielvieira:
- Add to documentation how to get transform info
- Add to documentation how to use the copy optical image

PR #1416, opened 2023-09-01
Added copy optical image to python client

lmacielvieira:
### Description

Added feature to python client, in which the user is able to copy an optical image and its transforms from one dataset to another #1410

## How to test it
1 - Update client
2 - Call the  copy_optical_image function
        >>> sm.copy_optical_image(
        >>>     origin_dataset_id='2018-11-07_14h15m28s',
        >>>     destiny_dataset_id='2018-11-07_14h15m30s',
        >>> )

### DIFF
--- a/metaspace/graphql/schemas/dataset.graphql
+++ b/metaspace/graphql/schemas/dataset.graphql
@@ -434,6 +434,8 @@ type Mutation {
 
   addOpticalImage(input: AddOpticalImageInput!): String!
 
+  copyRawOpticalImage(originDatasetId: String!, destinyDatasetId: String!): String!
+
   addRoi(datasetId: String!, geoJson: GeoJson!): String
 
   deleteOpticalImage(datasetId: String!): String!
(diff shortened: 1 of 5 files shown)


sergii-mamedov:
I was thinking of a slightly different implementation. 
I wouldn't want the images to be spread between multiple datasets. At some point, this will result in the original dataset being deleted and with it the optical images for the remaining datasets.
I was thinking of getting only the transformation matrix to use for previously uploaded images. Perhaps it is worth adding a separate function that would receive images from the original dataset, and upload them for the new dataset.

lmacielvieira:
> I was thinking of a slightly different implementation. I wouldn't want the images to be spread between multiple datasets. At some point, this will result in the original dataset being deleted and with it the optical images for the remaining datasets. I was thinking of getting only the transformation matrix to use for previously uploaded images. Perhaps it is worth adding a separate function that would receive images from the original dataset, and upload them for the new dataset.

What do you think about copying the image to the dataset of destiny instead of using the same one? As the transform is based on the source image, it does not make sense to use the same transform if the source image is different.

sergii-mamedov (approved):

PR #1417, opened 2023-09-01
Added one retry to ES update

lmacielvieira:
### Description

Retrying ElasticSearch update once more, in case a version conflict (elasticsearch.ConflictError) error is triggered. This error implies that the document has been modified since you last fetched or referenced it, but update_by_query  does not have the option  refresh='wait_for'  like the ES5, so a retry approach will be applied. #1412

### DIFF
--- a/metaspace/engine/sm/engine/daemons/dataset_manager.py
+++ b/metaspace/engine/sm/engine/daemons/dataset_manager.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import urllib.parse
+import time
 
 import boto3
 import elasticsearch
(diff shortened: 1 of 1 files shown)


metaspace/engine/sm/engine/daemons/dataset_manager.py
```diff
@@ -147,7 +148,11 @@ def index(self, ds: Dataset):
         ds.set_status(self._db, self._es, DatasetStatus.FINISHED)
 
     def update(self, ds, fields):
-        self._es.update_ds(ds.id, fields)
+        try:
+            self._es.update_ds(ds.id, fields)
+        except elasticsearch.ConflictError:  # tries to write to elasticsearch one more time
+            self.logger.info(f'Problem updating ES for: {ds.id}, trying again...')
+            self._es.update_ds(ds.id, fields)
```
sergii-mamedov, line 155: Maybe it makes sense to add a timeout?

PR #1421, opened 2023-09-20
[PythonClient] Added function to upload optical image from local file

lmacielvieira:
### Description

Added feature to python client, in which the user is able to add an optical image using a local file as source #1420

### DIFF
--- a/metaspace/python-client/metaspace/sm_annotation_utils.py
+++ b/metaspace/python-client/metaspace/sm_annotation_utils.py
@@ -198,7 +198,7 @@ def multipart_upload(
                 'source': 'api',
                 'user': current_user_id if current_user_id else 'not-provided',
                 'datasetId': dataset_id if dataset_id else 'not-provided',
-                'uuid': headers['uuid'] if headers and headers['uuid'] else 'not-provided',
+                'uuid': headers['uuid'] if headers.get('uuid') else 'not-provided',
             },
         }
         resp_data = send_request(url, 'POST', json=data, headers=headers)
@@ -2114,7 +2114,7 @@ class SMInstance(object):
         :return: Returns file s3 key
         """
 
-        image_url = self.upload_raw_opt_image_to_s3(local_path=local_path, dataset_id=dataset_id)
+        image_uuid = self.upload_raw_opt_image_to_s3(local_path=local_path, dataset_id=dataset_id)
         query = """
             mutation addOpticalImage($imageUrl: String!,
                 $datasetId: String!, $transform: [[Float]]!) {
@@ -2125,7 +2125,7 @@ class SMInstance(object):
 
         variables = {
             'datasetId': dataset_id,
-            'imageUrl': image_url,
+            'imageUrl': image_uuid,
             'transform': transformation_matrix.tolist(),
         }
 
--- a/metaspace/python-client/metaspace/sm_annotation_utils.py
+++ b/metaspace/python-client/metaspace/sm_annotation_utils.py
@@ -148,6 +148,7 @@ def get_config(
         'moldb_url': '{}/mol_db/v1'.format(host),
         'signin_url': '{}/api_auth/signin'.format(host),
         'gettoken_url': '{}/api_auth/gettoken'.format(host),
+        'raw_opt_upload_url': f'{host}/raw_opt_upload',
         'database_upload_url': f'{host}/database_upload',
         'dataset_upload_url': f'{host}/dataset_upload',
         'usr_email': email,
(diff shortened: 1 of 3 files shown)


metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -157,7 +158,15 @@ def get_config(
...
 
-def multipart_upload(local_path, companion_url, file_type, headers={}, current_user_id=None):
+def multipart_upload(
+    local_path,
+    companion_url,
+    file_type,
+    headers={},
+    current_user_id=None,
+    dataset_id=None,
+    uuid=None,
```
sergii-mamedov, line 168: Please check how Lachlan implemented works with uuid. 
https://github.com/metaspace2020/metaspace/blob/master/metaspace/python-client/metaspace/sm_annotation_utils.py#L287-L297

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -756,7 +774,15 @@ def create_database(
...
     ) -> dict:
         # TODO: s3 -> s3a in GraphQL
-        bucket, key = multipart_upload(local_path, self._config['database_upload_url'], 'text/csv')
+        result = self.query("""query { currentUser { id } }""")
+        current_user_id = result['currentUser'] and result['currentUser']['id']
+
+        bucket, key = multipart_upload(
+            local_path,
+            self._config['database_upload_url'],
+            'text/csv',
```
sergii-mamedov, line 783: I understand that you did not change this line, but why do we use `text/csv`?

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -186,7 +197,9 @@ def init_multipart_upload(filename, file_type, headers={}, current_user_id=None)
                 'name': filename,
                 'type': file_type,
                 'source': 'api',
-                'user': current_user_id,
+                'user': current_user_id if current_user_id else 'not-provided',
+                'datasetId': dataset_id if dataset_id else 'not-provided',
```
sergii-mamedov, line 201: As far as I understood the existing code, previously `uuid` was passed in `headers`. Now you need to form the Path as follows:
```
f'{bucket_name}/raw_images/{ds_id}/{filename(?)}'
```
Please explain how this happens in this code.

sergii-mamedov (changes requested):

metaspace/python-client/metaspace/sm_annotation_utils.py
lmacielvieira: I did not change it, maybe it was expect to be csv before tsv. There is no problem to keep this MIME type for tsv, if our parsing function can handle the tab delimiter. We could also change to 'text/tab-separated-values' if you think it is better.

sergii-mamedov: Sorry, I didn't see that it was create_database. I withdraw my question. You don't need to change anything.

metaspace/python-client/metaspace/sm_annotation_utils.py
lmacielvieira: These uuid referes to the unique file id that it will be used as filename. The same strategy was adopted to create the dataset folder ids and optical images file names. You can find how the path is generated at: https://github.com/metaspace2020/metaspace/blob/d0edc478c3f450209616f23af6154ed7b9f9f4e5/metaspace/graphql/src/modules/webServer/opticalImageUploadMiddleware.ts#L15

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -186,7 +196,9 @@ def init_multipart_upload(filename, file_type, headers={}, current_user_id=None)
...
+                'uuid': headers['uuid'] if headers and headers['uuid'] else 'not-provided',
```
sergii-mamedov, line 201: in the case when `uuid` does not exist in the dictionary, you will receive an exception
```
`headers['uuid'] -> headers.get('uuid')`
```

metaspace/python-client/metaspace/sm_annotation_utils.py
```diff
@@ -2032,6 +2056,81 @@ def add_dataset_external_link(
...
+        >>>     transformation_matrix=np.array([[1, 0, 0], [0, 1, 0], [0, 0, 1], ]),
+        >>> )
+
+        :param local_path:
+        :param dataset_id:
+        :param transformation_matrix:
+        :return: Returns file s3 key
+        """
+
+        image_url = self.upload_raw_opt_image_to_s3(local_path=local_path, dataset_id=dataset_id)
```
sergii-mamedov, line 2117: `upload_raw_opt_image_to_s3` return back only `uuid`. Is it ok? Because variable name is `image_url`, not `image_uuid`

sergii-mamedov (approved):

metaspace/python-client/metaspace/sm_annotation_utils.py
lmacielvieira: Yes, it is ok. The uuid is the key used to find the image on the related s3 bucket. I will rename the variable.

PR #1455, opened 2023-12-08
[CustomDatabase] added feature to enable public custom database

lmacielvieira:
### Description

Added feature to enable public custom databases so that users can process datasets with it #1454 
Added custom db upload instructions wiki link to 'Upload database' modal #1337 

image
image
image


sergii-mamedov (approved):
Should we add this functionality to python-client? I think so.
You can make it a separate PR if you want.

PR #1466, opened 2024-01-31
Removing CHNOPS adducts from decoy pool

Bisho2122:
Having CHNOPS based adducts will increase the chance for a generated decoy ion to be isomeric or isobaric to an ion of a real molecule. As this affects the FDR calculation, it's important to remove those adducts from the decoy selection pool.


sergii-mamedov:
@theodev Look at this PR from the point of view of science.

I think we should regenerate all `mol_db cache` after deploying this code to production. @lmacielvieira @theodev
The number of annotations will change in the case of reprocessing for existing datasets. This is the problem investigated by @lmacielvieira early.

PR #1515, opened 2024-04-16
Non deterministic behavior of np.random with ThereadPoolExecutor

sergii-mamedov:
### Overview
In the docker image that uses AWS Lambda via remote `/dev/shm`. The problem is explained in more detail [here](https://stackoverflow.com/questions/34005930/multiprocessing-semlock-is-not-implemented-when-running-on-aws-lambda). To work around this situation, when migrating to AWS, I changed `ProcessPoolExecutor` to `ThreadPoolExecutor`.
I didn't consider that in the case of `ThreadPoolExecutor` `numpy.random` would share between all threads running at the same time (see [Good practices with numpy random number generators](https://albertcthomas.github.io/good-practices-random-number-generators/)).

### Solution
Replaced `ThreadPoolExecutor` with the usual `for loop`.

### DIFF
--- a/metaspace/engine/sm/engine/annotation_lithops/build_moldb.py
+++ b/metaspace/engine/sm/engine/annotation_lithops/build_moldb.py
@@ -1,8 +1,6 @@
 from __future__ import annotations
 
 import logging
-from concurrent.futures import ThreadPoolExecutor, ProcessPoolExecutor
-from itertools import repeat
 from typing import List, Tuple, TypedDict, Optional, cast
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


sergii-mamedov:
### Legacy Generator vs Random Generator
Our code uses the `Legacy Generator (RandomState)`, which is no longer being developed. The new `Random Generator` implementation is thread safe and faster. Due to this, backward compatibility in sequence generation [was broken](https://numpy.org/doc/1.23/reference/random/new-or-different.html). This affects us with another subset of decoys for existing molecular databases.
A code snippet that shows the difference:
```
import numpy as np

# old generator
np.random.seed(42)
old = data.copy()
np.random.shuffle(old)

# new generator
rng = np.random.default_rng(seed=42)
new = data.copy()
rng.shuffle(new)

print(old)
print(new)
```

lmacielvieira (approved):

PR #1516, opened 2024-04-17
[Scitest] updated v3 model output

lmacielvieira:
### Description

* Update sci-test files for sanity check. The files for v3 were generated with the model Animal v2.2023-12-14.
* Added no-cache option to UUID fetch request on upload page.


sergii-mamedov (approved):
Note that `sci-test-on-demand` only runs the spark implementation.

PR #1541, opened 2024-06-07
[DatasetComparison] updating ion image on filter change

lmacielvieira:
### Description

Fixed bug for dataset comparison where ion image was not updated on filter change.
Disable loading archived scoring models from old datasets as templates to new ones.


sergii-mamedov (approved):
LGTM

PR #1542, opened 2024-06-11
Upgraded docker files (Ubuntu 20.04)

lmacielvieira:
### Description

Upgrade docker files to meet service upgrades. #1289 
Redis 4.0.9 -> 7.2.5
Elasticsearch 8.7.1 -> 8.14.0
Postgres 9.5 -> 16.3
Nginx 1.19.6 -> 1.25.4
Ubuntu 18.04 -> 20.04
Node 16.20 -> 20.14

## Instructions
Before deploying on the machine. Node version must be present
1 - Install node: `nvm install 20.14.0`
2 - Select installed version: `nvm use 20.14.0`
3 - Set default node: `nvm alias default 20.14.0`

*** [Local run] In case you have a previous installation, please delete the data related folders: data/elasticsearch, data/s3, data/pgdata


metaspace/graphql/.nvmrc
Bisho2122, line 1: Different version of nvm trigger this error https://stackoverflow.com/questions/72921215/getting-glibc-2-28-not-found 

16.20 worked but 20.14 invoked the error. You've probably tested it internally and maybe it works, but just wanted to point it out cause I got this before with 16.20. But, I had Ubuntu 18.04, so maybe upgrading to 20.4 will solve it.

lmacielvieira: I've not tested it only internally. The test was done on docker and also on the staging machine (without docker). The problem you pointed is because node version 20.14 is supported on Ubuntu 18.04, that's why the upgrade (as instructed an as default on upgrading OS) and the reference task (https://github.com/metaspace2020/metaspace/issues/1289). I must have marked you as reviewed by mistake on this task.

sergii-mamedov (approved):
1. We need to specify rabbitmq version.
2. Please update also MINIO.
3. Database files must be deleted (add this in instruction).

Issue #1507, opened 2024-03-29
[Enrichment] Update metabolite enrichment


lmacielvieira:

Bisho2122:
Add the following filters for the users **After processing** :

- [ ] If annotation database is HMDB / CoreMetabolome : 
  - [ ] Consider pathway associated annotations
  - [ ] Remove expected & predicted annotations
- [ ] p-value / q-value cutoff 
- [ ] minimum intersection size (default =3) : Intersection size between input metabolites and each term. 

Add the following filters for the users **Before processing** (align with the upload page) :

- [ ] Consider isomers/isobars or not (Bootstrap or not)
- [ ] Background type (Super class, main class, subclass)
- [ ] Molecule type : Lipid, Metabolites
- [ ] Add upload of custom universe (optional)

PR #1543, opened 2024-06-13
Feat/metabolite enrichment

lmacielvieira:
### Description

Added metabolite enrichment feature #1507 #1287 
Changed enrichment calculation strategy #1369 

### How to test it

1 - Run the script docker/sm-engine/install-enrichment-dbs.sh to add the new ontology databases
2 - Update 'LION' existing record column 'mol_type' to 'lipid' and 'category' to 'class'
3 - Go to upload page using the feature flag (<metaspace_url>/upload?feat=metabo_enrich)
 3.1 - As it is running forked from an older version with a bug that was corrected, it my be needed to access first the datasets page with the feature flag, and then click on the tabgo to the upload page (<metaspace_url>/datasets?feat=metabo_enrich)
4 -Select desired ontology databases and upload the dataset
5 - Go to dataset overview, click actions and then click on ontology enrichment

** Annotate using Original MSM, as this is prior to the scoring Models change

image
image
image


metaspace/graphql/src/modules/dataset/controller/Query.ts
Bisho2122, line 1: Wondering if it would be better if we invoke those SQL queries in python rather than typescript. I understand that it's meant to be on the fly everytime the user refresh the enrichment page, but wondering if the dataset is bigger would it have a longer waiting time or not.

Bisho2122: @sergii-mamedov mentioned that we might need to do some profiling first before taking that decision if need be.

metaspace/webapp/src/modules/Datasets/overview/DatasetActionsDropdown.tsx
```diff
@@ -86,12 +86,12 @@ export const DatasetActionsDropdown = defineComponent({
     const confirmReprocessEnrichment = async () => {
       try {
         await ElMessageBox.confirm(
-          'The changes to the analysis options require the dataset to be reprocessed. ' +
-            'This dataset will be unavailable until reprocessing has completed. Do you wish to continue?',
-          'Reprocessing required',
+          'To perform the ontology enrichment, please select the desired enrichment ontology ' +
```
Bisho2122, line 89: I'd replace ontology with background. Only LION is an ontology here

metaspace/webapp/src/modules/MetadataEditor/sections/MetaspaceOptionsSection.vue
```diff
@@ -141,12 +141,13 @@
             <el-col :span="8">
               <popup-anchor feature-key="v2" placement="top" :show-until="new Date('2022-09-01')" class="block">
                 <form-field
-                  type="switch"
-                  name="LION enrichment"
+                  type="selectMultiTree"
+                  name="Enrichment ontology"
```
Bisho2122, line 145: Same as before, anything on the web interface, we can replace ontology with background.

metaspace/engine/sm/rest/enrichment.py
Bisho2122, line 1: Will probably need to change the calculation to account for false negatives. In the sense that similar isomers of a given formula observed as a true positive should not be counted as false negatives. 

Since in bootstrapping only 1 isomer is selected for observed while in background, there are multiple isomers of the same formula for a given term, It's removed from false negatives if any of the isomers are observed as true positve.

PR #1549, opened 2024-07-02
Fix AttributeError when checking `logged_in`

aeisenbarth:
When METASPACE login fails in the constructor, the error is silently caught and printed, which goes unnoticed if not using an interactive console/notebook. When checking `logged_in()`, an AttributeError is raised because the instance is only fully initialized after successful login.

## Example

```pycon
>>> from metaspace.sm_annotation_utils import SMInstance
>>> sm = SMInstance(email="partyparrot@ateam.com", api_key="invalid-api_key")
Failed to connect to https://metaspace2020.eu: Invalid API key

>>> sm
SMInstance(https://metaspace2020.eu/graphql)
>>> sm.logged_in()
Traceback (most recent call last):
  File "…/metaspace/metaspace/python-client/metaspace/sm_annotation_utils.py", line 1707, in logged_in
    return self._gqclient.logged_in
AttributeError: 'SMInstance' object has no attribute '_gqclient'
```

- Login with API key and password take different code paths, only invalid API keys make GraphQLClient raise `BadRequestException("None: Invalid API Key.")`.
- Ideally, instance attributes are only defined inside `__init__`.
- An instance should either be completely initialized, or an error should be raised instead of returning a partially initialized instance.
- From `logged_in`, one would expect True/False, almost never an exception.

## Solution

I changed the API key login in `GraphQLClient` to be handled the same as password login: no exception, but setting login status and printing a message. This way, `GraphQLClient` can be instantiated despite invalid API key, and `SMInstance` can be completely initialized.

### DIFF
--- a/metaspace/python-client/metaspace/sm_annotation_utils.py
+++ b/metaspace/python-client/metaspace/sm_annotation_utils.py
@@ -348,7 +348,14 @@ class GraphQLClient(object):
         self.logged_in = False
 
         if self._config.get('usr_api_key'):
-            self.logged_in = self.query('query { currentUser { id } }') is not None
+            try:
+                self.logged_in = self.query('query { currentUser { id } }') is not None
+            except BadRequestException as ex:
+                if 'Invalid API key' in ex.message or 'Login failed' in ex.message:
+                    print('Login failed. Only public datasets will be accessible.')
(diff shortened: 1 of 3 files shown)


lmacielvieira (approved):
LGTM

PR #1590, opened 2024-10-21
A new script for archiving files

sergii-mamedov:
1. Removed the script for saving files from IBM Cloud, as it is no longer needed.
2. The script for AWS Cloud has been significantly changed, the possibility of using SQL Query to define datasets has been added.


metaspace/engine/scripts/download_imzml_ibd.py
```diff
@@ -0,0 +1,180 @@
...
+
+    # check if two argument is initialized
+    if bool(args.datasets_file) & bool(args.sql_where):
+        parser.print_usage()
+        print('Error: must specify only one argument either --datasets-file, --sql-where')
+        sys.exit(1)
+
+    sql_where = args.sql_where if args.sql_where else None
+
+    s3_client = boto3.client('s3')
```
lmacielvieira, line 168: Maybe move to inside of GlobalInit and add parameters to work with minio:
`
        s3_client = boto3.client('s3',
                                 aws_access_key_id=sm_config["storage"]["access_key_id"],
                                 aws_secret_access_key=sm_config["storage"]["secret_access_key"],
                                 endpoint_url=sm_config["storage"]["endpoint_url"])
`

lmacielvieira (approved):

metaspace/engine/scripts/download_imzml_ibd.py
sergii-mamedov: We don't need minio support for this script. It should run only on the production instance.

PR #1591, opened 2024-10-24
Lithops 3.5.1

sergii-mamedov:
Upgrade Lithops from 3.2.0 to 3.5.1


metaspace/engine/sm/engine/annotation_lithops/annotation_job.py
```diff
@@ -287,7 +287,7 @@ def __init__(
         self.perf = perf
         self.store_images = store_images
         self.perform_enrichment = perform_enrichment
-        self.ont_db_ids = self.ds.config['ontology_db_ids'] or []
+        self.ont_db_ids = self.ds.config.get('ontology_db_ids', [])
```
sergii-mamedov, line 290: It does not apply to this task, but this is a bug that is definitely present in the staging. 
Most likely, the staging config does not have this field.

lmacielvieira (approved):

PR #1600, opened 2024-11-25
Exclude chem_mods and neutral losses from decoy selection

Bisho2122:
Changed decoy sampling so that chemical modifications and neutral losses won't be sampled from the list of possible decoy adducts.


lmacielvieira (approved):
LGTM

sergii-mamedov:
@Bisho2122 I asked this not for myself, but for someone who will be involved in this project after us ;)

PR #1658, opened 2025-12-04
Feat/add_mass_accuracy

Bisho2122:
Adds mass accuracy (mass error in ppm) as a new column to the Annotation Table with full sorting and CSV export support.


metaspace/webapp/src/modules/Annotations/AnnotationTable.vue
```diff
@@ -661,6 +681,23 @@ export default defineComponent({
       }
     }
 
+    const formatMassError = (row) => {
```
lmacielvieira: This function is called for every table row on every render, causing performance issues with large datasets. Please move this calculation to the GraphQL side (src/modules/annotation.controller/Annotation.ts) and make the necessary adjustments.
I.e
1 - Add massErrorPpm: Float field to GraphQL schema
2 - Implement resolver that pre-computes the ppm value
3 - Simplify this function to: return row.massErrorPpm?.toFixed(2) ?? '-'

lmacielvieira (commented):
I've added suggestions straight in the code snippets. In summary, we have some issues:
1 - JSON parsing on every table render (performance bottleneck)
2 - Duplicate calculation logic in multiple places
3 -  Try/catch overhead for every row
4 -  Inconsistent error handling

Proposed Solution:
1 - Pre-compute mass error in GraphQL resolver
2 - Single source of truth for calculation
3 -  Eliminate frontend JSON parsing for this part

Additional to that:
Please use a meaningful name for the PR, for example: “Add mass accuracy to annotation table.” The feat/branch_name notation is only for maintaining GitFlow-style naming conventions (even though we use a trunk-based approach), and should not be used as the PR title.

PR #1703, opened 2026-03-19
Enable adding optical image from dataset overview actions

Bisho2122:
Add optical image route to `/dataset` prefix and expose via Actions dropdown                                                                                        
                                                                                                                                                                    
  Summary                                                                                                                                                           
                                                                                                                                                                    
  - Moved the add-optical-image route from `/datasets/:dataset_id/add-optical-image` to `/dataset/:dataset_id/add-optical-image` to align with the existing `/dataset/:dataset_id/...` route convention used across the webapp
  - Added an "Add optical image" item to the `DatasetActionsDropdown` on the dataset overview page.


lmacielvieira (approved):
LGTM.

I've made a few changes tho:
1 - Fixed tests
2 - Added a redirect from the old route /datasets, in case users have it saved
3 - Sort the 'Dataset options' alphabetically

PR #1719, opened 2026-04-17
Eliminate toarray() for spectral/spatial metrics

Bisho2122:
## Problem

Follow-up to #1718 . The annotation pipeline sometimes crashes with OOM on datasets where spectra form isolated blobs inside a large bounding box. After #1718 fixed `chaos_metric`, the remaining crash point is `formula_validator.py` which still materialises the full h×w dense array for every peak of every formula when building iso_imgs_flat.
             
## Change

  - `imzml_reader` gets a `pixel_to_flat_idx` lookup table built once at init — maps any pixel coordinate directly to its position in the           
  masked-flat metrics array.
  - `formula_validator.py` uses it to scatter sparse `coo` values directly into a 1D array of size `n_spectra`, removing the `toarray()` call and the intermediate dense `iso_imgs` list entirely.
  - `pipeline.py` logs a warning when pixel density is below 5% so blob datasets are visible in logs.


Bisho2122:
Replaced with #1744

PR #1757, opened 2026-06-24
Add RMS and Median global normalization options

Bisho2122:
Replaces the TIC normalization checkbox in the ion image viewer settings panel with a Normalization dropdown (None / TIC / RMS / Median).

  ## Engine
  - Computes RMS (√mean(i²)) and Median intensity per pixel during spectrum reading in `ImzMLReader`
  - Stores both as new `DiagnosticType.RMS` and `DiagnosticType.MEDIAN` dataset diagnostics (.npy images), alongside the existing TIC diagnostic 
 
  ## Frontend
  - Settings panel gear icon now shows a "Normalization" dropdown instead of a TIC checkbox
  - Fetches the appropriate diagnostic based on the selected type; falls back to `None` with a warning if the diagnostic isn't available (e.g. dataset predates this change)
  - Normalization badge text updates dynamically ("TIC normalized" / "RMS normalized" / "Median normalized")
  - URL param `?norm` now stores the type string (TIC, RMS, MEDIAN); old `?norm=true` links default to TIC

  ```Note: Existing datasets will not have RMS/Median diagnostics until reprocessed.```


lmacielvieira (commented):
Reviewed the branch and pushed fixes for the issues I found.

**Engine**

- RMS/median were broken for any dataset with blank pixels.** Intensities are filtered to `> 0`
before the per-pixel stats, so an all-zero spectrum left `NaN` in `_sp_rms`/`_sp_median`. `rms_image()`/`median_image()`
assert on NaN, so both diagnostics retried 3× and got stored as errors — while TIC worked fine,
because `np.sum([])` is `0.0`. Fixed by writing `0.0` for empty spectra, matching TIC's semantics.

- Integer intensities overflowed in `ints**2`, wrapping negative and making RMS NaN. Square is now done in float64.
- `update_diagnostics.py` skipped reading spectra when metadata had m/z and TIC, but RMS/median only come from spectra. Always reads now (might make it slower).
- Added `DATASET_DIAGNOSTIC_TYPES`, used by both the SQL filter and `extract_dataset_diagnostics`, so the list and the function can't drift.
- Added tests for rms/median images, the empty-spectrum case, integer intensities, and the two new diagnostics in the lithops test.

**Webapp**

- Added normalizations do dataset overview page
- Fixed URL/snapshot restore called `handleNormalizationChange(true)`, losing the type, so the badge always fell back to "TIC normalized".
- CSV exports were headed "TIC normalized" and named `_tic_normalized` in every mode, in the annotations table, ROI export and comparison.
- Pixel-hover tooltip always read "TIC-relative intensity"; now reads the type from the normalization data that was actually applied.
- Types, labels, `min_*`/`max_*` prefixes, URL parsing and diagnostic lookup were duplicated in 5 places; now in `lib/normalization.ts`.

**Verification**

- Please re-check the comparison page, annotations page, tooltip and URLs again, to see if it is all sound

